PR #7, opened 2023-10-19
Rename to tenantId for consistency

joll90:
All endpoints and interfaces renamed with tenantId to be consistent with base interfaces.

### DIFF
--- a/00_Base/src/interfaces/api/MessageQuerystring.ts
+++ b/00_Base/src/interfaces/api/MessageQuerystring.ts
@@ -19,5 +19,17 @@
  */
 export interface IMessageQuerystring {
     identifier: string;
-    partyId: string;
+    tenantId: string;
+}
+
+/**
+ * This message querystring schema describes the {@link IMessageQuerystring} interface.
+ */
+export const IMessageQuerystringSchema = {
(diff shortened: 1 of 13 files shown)


00_Base/src/interfaces/api/MessageQuerystring.ts
```diff
@@ -19,5 +19,17 @@
  */
 export interface IMessageQuerystring {
     identifier: string;
-    partyId: string;
+    tenantId: string;
```
elliot-sabitov, line 22: @joll90 Have you considered using constants for the keys? I have found this helpful in the past. For example:

```
export enum SomeObjectProperties {
  id = 'id',
  ...
}
export interface SomeObject extends SomeObjectBase {
  [SomeObjectProperties.id]?: string; // nullable
}
```
this way when you have to refactor, you can just go to the enum and change in one place and everywhere where you have tenantId can just be easily connected and updated

joll90: Thanks @elliot-sabitov for bringing this up. Most of the places this `tentantId` is used is method signatures, only two interfaces really make use of this field name.

PR #8, opened 2023-10-20
Feature/expanding cache methods

thanaParis:
onChange and getAndRemove added to ICache. Impls added to MemoryCache and RedisCache.
Method descriptions:
onChange
```
   * Monitors a key for potential changes to its value.
   * If key-value does not exist this method will wait for it to exist or return null at the end of the wait period.
   * If value is removed, the method will return null.
```
getAndRemove
```
   * Gets and removes a value asynchronously from the underlying cache.
   * This is an atomic operation. The value will not be modified or read by any other process between getting and removing.
```

### DIFF
--- a/00_Base/src/interfaces/cache/cache.ts
+++ b/00_Base/src/interfaces/cache/cache.ts
@@ -32,6 +32,18 @@ export interface ICache {
   remove(key: string, namespace?: string): Promise<boolean>;
 
   /**
+   * Monitors a key for potential changes to its value.
+   * If key-value does not exist this method will wait for it to exist or return null at the end of the wait period.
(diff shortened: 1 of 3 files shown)


joll90 (approved):
What's the intention with the `_lockMap` it seems to be only used in the `getAndRemove`?

thanaParis:
> What's the intention with the _lockMap it seems to be only used in the getAndRemove?

to ensure the atomicity of getAndRemove
> This is an atomic operation. The value will not be modified or read by any other process between getting and removing.
lock map prevents any other threads from performing operations between get and remove. in the future, it could have other uses for locking a cache value from other threads.

PR #9, opened 2023-10-25
Update issue templates

thanaParis:
As suggested https://github.com/citrineos/citrineos-core/issues/5


elliot-sabitov (approved):
Thank you!

Issue #13, opened 2023-10-29
docker-compose setup


Anmirazik:
```
> @citrineos/server@1.0.0 start-unix:docker
> export APP_ENV=docker && npx nodemon

[nodemon] 2.0.22
[nodemon] to restart at any time, enter `rs`
[nodemon] watching path(s): src/**/*
[nodemon] watching extensions: ts,js
[nodemon] starting `npx ts-node ./src/index.ts`
2023-10-29 19:54:41.496 INFO    /node_modules/@citrineos/provisioning/lib/module/module.js:96   CitrineOS Logger:ProvisioningModule     Initializing...
Error: getaddrinfo ENOTFOUND host.docker.internal
    at GetAddrInfoReqWrap.onlookup [as oncomplete] (node:dns:108:26)
    at module.exports.runLoopOnce (/usr/server/node_modules/deasync/index.js:65:10)
    at module.exports (/usr/server/node_modules/deasync-promise/index.js:13:9)
    at new ProvisioningModule (/usr/server/node_modules/@citrineos/provisioning/lib/module/module.js:97:44)
    at new CitrineOSServer (/usr/server/src/index.ts:84:36)
    at Object.<anonymous> (/usr/server/src/index.ts:155:1)
    at Module._compile (node:internal/modules/cjs/loader:1256:14)
    at Module.m._compile (/usr/server/node_modules/ts-node/src/index.ts:1618:23)
    at Module._extensions..js (node:internal/modules/cjs/loader:1310:10)
    at Object.require.extensions.<computed> [as .ts] (/usr/server/node_modules/ts-node/src/index.ts:1621:12) {
  errno: -3008,
  code: 'ENOTFOUND',
  syscall: 'getaddrinfo',
  hostname: 'host.docker.internal'
}
[nodemon] app crashed - waiting for file changes before starting...
```

I face this error when run `docker-compose up -d `in the `50_Server ` file . This errors comes from the citrine docker container which leads to the port 8081 unreachable from the browser

joll90:
Unfortunately `v0.0.1` encounters a problem running locally on docker on Ubuntu operating systems. The fix for this problem is described below:

1. Go to the folder 50_Server/src/config/envs where you can find a file called `docker.ts`
2. In this file replace all occurrences of `host.docker.internal` with `172.18.0.1` 
3. Save the file and go back into `50_Server` and run `docker-compose up -d`
4. CitrineOS now loads without a problem

scarter93:
[...]

PR #14, opened 2023-10-31
Bugfix- for linux docker-compose network #13

ChrisWeissmann:
#13 Describes that on ubuntu they can not run `docker-compose up`, as their docker engine can not resolve `host.docker.internal`. This seems to be specific for linux distribution. 

In this PR we move to use Dockers service discovery instead of `localhost` or `host.docker.internal` and use the name of the other container in the docker-compose file. So we simply adjust the naming in the config class. 

Further I did a slight update to the README.


scarter93 (approved):
Looks good! thanks for including the readme changes 👍

PR #15, opened 2023-10-31
Callback support

thanaParis:
New optional http request parameter for message api: `callbackUrl`. OCPP responses to requests sent via message api will be POSTed to the callbackUrl in the request param, if provided.

### DIFF
--- a/00_Base/src/interfaces/modules/AbstractModule.ts
+++ b/00_Base/src/interfaces/modules/AbstractModule.ts
@@ -166,6 +166,7 @@ export abstract class AbstractModule implements IModule {
                     body: JSON.stringify(message.payload)
                 });
             } catch (error) {
+                // TODO: Ideally the error log is also stored in the database in a failed invocations table to ensure these are visible outside of a log file.
                 this._logger.error("Failed sending call result: ", error);
             }
         }
--- a/00_Base/src/interfaces/modules/AbstractModule.ts
+++ b/00_Base/src/interfaces/modules/AbstractModule.ts
@@ -18,6 +18,7 @@
 
 import 'reflect-metadata';
 import { ILogObj, Logger } from "tslog";
+import { v4 as uuidv4 } from "uuid";
 import { AS_HANDLER_METADATA, IHandlerDefinition, IModule } from ".";
 import { OcppRequest, OcppResponse } from "../..";
 import { SystemConfig } from "../../config/types";
@@ -26,8 +27,11 @@ import { RequestBuilder } from "../../util/request";
 import { CacheNamespace, ICache } from "../cache/cache";
 import { ClientConnection } from "../centralsystem";
 import { EventGroup, HandlerProperties, IMessage, IMessageConfirmation, IMessageHandler, IMessageSender, MessageOrigin, MessageState } from "../messages";
+import axios from 'axios';
 
 export abstract class AbstractModule implements IModule {
+
+    public readonly CALLBACK_URL_CACHE_PREFIX: string = "CALLBACK_URL_";
 
     protected _config: SystemConfig;
     protected readonly _cache: ICache;
@@ -52,6 +56,10 @@ export abstract class AbstractModule implements IModule {
     /**
      * Getters & Setters
      */
+
+    get cache(): ICache {
+        return this._cache;
+    }
 
     get sender(): IMessageSender {
         return this._sender;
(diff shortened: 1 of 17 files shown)


00_Base/src/interfaces/modules/AbstractModule.ts
```diff
@@ -144,6 +155,17 @@ export abstract class AbstractModule implements IModule {
         }
     }
 
+    async handleMessageApiCallback(message: IMessage<OcppResponse>): Promise<void> {
+        const url: string | null = await this._cache.getAndRemove(message.context.correlationId, this.CALLBACK_URL_CACHE_PREFIX + message.context.stationId);
+        if (url) {
+            try {
+                await axios.post(url, message.payload);
```
joll90, line 162: @thanaParis I would recommend to use `NodeJS` fetch instead of `Axios` as `Axios` adds another dependency and is not necessary to make a simple post request.

00_Base/src/interfaces/modules/AbstractModule.ts
```diff
@@ -144,6 +155,17 @@ export abstract class AbstractModule implements IModule {
...
+            } catch (error) {
+                this._logger.error("Failed sending call result: ", error);
```
joll90, line 164: @thanaParis Ideally the error log is also stored in the database in a failed invocations table to ensure these are visible outside of a log file.

01_Provisioning/src/module/api.ts
```diff
@@ -69,71 +72,54 @@ export class ProvisioningModuleApi extends AbstractModuleApi<ProvisioningModule>
...
         while (setVariableData.length > 0) {
-            const setVariableDataSubset = setVariableData.slice(0, itemsPerMessageSetVariables);
-            const messageConfirmation = await this._module.sendCall(identifier, tenantId, CallAction.SetVariables,
-                { setVariableData: setVariableDataSubset } as SetVariablesRequest);
-            if (!messageConfirmation.payload) {
-                finalMessageConfirmation.success = false;
-                finalMessageConfirmation.payload = finalMessageConfirmation.payload ?
-                    (finalMessageConfirmation.payload as SetVariableDataType[]).concat(setVariableDataSubset) : setVariableDataSubset
-            }
+            await this._module.sendCall(identifier, tenantId, CallAction.SetVariables,
```
joll90, line 76: @thanaParis what if this `sendCall` fails? We seem to only return `success` here?

01_Provisioning/src/module/api.ts
```diff
@@ -69,71 +72,54 @@ export class ProvisioningModuleApi extends AbstractModuleApi<ProvisioningModule>
...
         while (getVariableData.length > 0) {
-            const getVariableDataSubset = getVariableData.slice(0, itemsPerMessageGetVariables);
-            const messageConfirmation = await this._module.sendCall(identifier, tenantId, CallAction.GetVariables,
-                { getVariableData: getVariableDataSubset } as GetVariablesRequest);
-            if (!messageConfirmation.payload) {
-                finalMessageConfirmation.success = false;
-                finalMessageConfirmation.payload = finalMessageConfirmation.payload ?
-                    (finalMessageConfirmation.payload as GetVariableDataType[]).concat(getVariableDataSubset) : getVariableDataSubset
-            }
+            await this._module.sendCall(identifier, tenantId, CallAction.GetVariables,
```
joll90, line 98: @thanaParis what if this `sendCall` fails? We seem to only return `success` here?

joll90 (changes requested):
@thanaParis This is a great addition and provides added flexibility of integrating CitrineOS into existing systems. Please take a look at some of my comments and let me know what you think.

00_Base/src/interfaces/modules/AbstractModule.ts
thanaParis: This is currently impossible, given our persistence logic is in 10_Data which relies on 00_Base. However, I've added a comment to reflect this point.

thanaParis:
Thank you for the feedback! I've updated the pr. Axios is no longer used and SetVariables/GetVariables now return a verbose response containing information on all the messages they attempted to send to the station. I've left a comment on your remark about storing failed invocations. This may be a reason in the future to move the '/layers/sequelize/repository/Base.ts' and '/layers/sequelize/util.ts' files to 00_Base.

joll90 (approved):

PR #16, opened 2023-11-06
Feature/multi stage dockerfile

thanaParis:
- Added .gitatrributes file to ensure consistent line endings for builds
- Expanded .dockerignore to trim size of docker image
- Altered Dockerfile to build the project on container rather than use host files
- Altered Dockerfile to use multiple stages with careful layering to ensure only altered parts of the project rebuild, to make development easier

Note: changes related to the getAndRemove and onChange ICache methods are in this pr despite having been reviewed and merged already. A previous merge appears to have undone them accidentally. Those changes (to ICache, MemoryCache, RedisCache) can be ignored.


99_Util/src/cache/memory.ts
```diff
@@ -36,30 +72,99 @@ export class MemoryCache implements ICache {
...
 
-  remove(key: string, namespace?: string): Promise<boolean> {
+  async remove(key: string, namespace?: string): Promise<boolean> {
     namespace = namespace || "default";
-    key = `${namespace}:${key}`;
-    return Promise.resolve(this._cache.delete(key));
+    const namespaceKey = `${namespace}:${key}`;
+    const lock = this._lockMap.get(namespaceKey);
+    if (lock) {
+      await lock;
```
scarter93, line 80: does this promise have a timeout? we should not await forever on a lock

thanaParis:
Hi @scarter93 ! Unfortunately, those changes were reviewed in [this PR](https://github.com/citrineos/citrineos-core/pull/8) and are not up for review as part of this PR. they are only in the diff because of an issue with the commit history caused by rebasing before the last merge to the release candidate. I'm thankful for the feedback, and will consider a PR for some of your suggested features (unless you beat me to it!).

PR #20, opened 2023-11-16
Rc 0.2.0

thanaParis:
- ws server now uses http server wrapper in noServer mode to enable http basic authentication and TLS.
   - new config values (webProtocol, httpsPrivateKeysFilepath, httpsCertificateChainFilepath) added to support TLS
- Request and/or Response handling added for DataTransfer, MeterValues, CustomerInformation, GetInstalledCertificateIds, GetLocalListVersion, GetLogs, RequestStartTransaction, RequestStopTransaction, InstallCertificate, DeleteCertificate, TriggerMessage, UpdateFirmware, GetReport, SendLocalList, UnlockConnector, LogStatusNotification, FirmwareStatusNotification, NotifyCustomerInformation
- Variable status handling expanded
  - VariableStatus entity created and associated many-to-one with VariableAttribute, replacing status fields in order to keep a history of all previous statuses of all previous values
  - VariableStatus entries created for SetVariables, GetVariables, NotifyReport, and on use of device model data PUT endpoint with 'setOnCharger' param equal to true (reflectes values manually set on charger, or otherwise set outside of ocpp)
- AutoAccept option added to system config, default value true. when false, BootNotification handling never updates boot config to Accepted, can only happen when set via data endpoint.
- JSON stringified CallResults (to csms) are now automatically cached under key correlationId namespace stationId for maxCallLengthSeconds, so that onChange can be used to listen for CallResults (new pattern used in boot notification handling to wait for SetVariablesResponses)
- Custom set function added to VariableAttribute for value; when dataType of VariableCharacteristics is passwordString the value is hashed using bcrypt
  - DataEnumType in spec is missing passwordString and identifierString types despite both being referenced in Appendix. PasswordString was added to this enum so that it could be handled differently. in the future, 'get' override could be used to return object types appropriate for dataType value, such as number or array
  - 50_Server's server impl accesses device model repo to fetch passwordString for http basic authentication and uses bcrypt to compare incoming auth headers to hashed value in db
- Bugfixes to DeviceModel repo impl for correct VariableCharacteristics handling and update behavior
- Bugfixes to Authorization repo impl for correct update behavior
- Authorize request handling fixed so that idTokenInfo is extracted without sequelize Model<any, any> fields
- Bcrypt rebuilt in dockerfile because --ignore-scripts option prevents normal behavior
- ajv addFormats used to include date-time string format used in json schemas
- TransactionEvent repo impl's readAllTransactionsByStationIdAndChargingStates updated to readAllTransactionsByStationIdAndEvseAndChargingStates in anticipation of usage by MeterValues handling, which can be specific to an evse


00_Base/src/config/types.ts
```diff
@@ -103,8 +104,11 @@ export const systemConfigSchema = z.object({
         })
     }),
     websocketServer: z.object({
+        webProtocol: z.enum(['http', 'https']).default('http'),
```
joll90, line 107: @thanaParis This should be an `TLS flag` instead of `webProtocol`. If we use something like `webProtocol` it should be named `socketProtocol` and would require values like `WSS` or `WS`.

00_Base/src/config/types.ts
```diff
@@ -103,8 +104,11 @@ export const systemConfigSchema = z.object({
...
+        httpsPrivateKeysFilepath: z.string().optional(),
```
joll90, line 108: Similarly here this would be `tlsKeysFilepath`.

00_Base/src/config/types.ts
```diff
@@ -103,8 +104,11 @@ export const systemConfigSchema = z.object({
...
+        httpsCertificateChainFilepath: z.string().optional(),
```
joll90, line 109: Similarly here this would be `tlsCertificateChainFilepath`

00_Base/src/interfaces/centralsystem.ts
```diff
@@ -30,7 +30,7 @@ export class OcppError extends Error {
     private _errorCode: ErrorCode;
     private _errorDetails?: object;
 
-    constructor(messageId: string, errorCode: ErrorCode, errorDescription: string, errorDetails?: object) {
+    constructor(messageId: string, errorCode: ErrorCode, errorDescription: string, errorDetails: object = {}) {
```
joll90, line 33: @thanaParis Why did this change from an optional? The class field is still optional?

00_Base/src/interfaces/modules/AbstractModule.ts
```diff
@@ -202,7 +205,7 @@ export abstract class AbstractModule implements IModule {
         if (callbackUrl) {
             // TODO: Handle callErrors, failure to send to charger, timeout from charger, with different responses to callback
             this._cache.set(_correlationId, callbackUrl, this.CALLBACK_URL_CACHE_PREFIX + identifier,
-                this.config.websocketServer.maxCallLengthSeconds * 20); // Fudge factor for any network lag
+                this.config.websocketServer.maxCallLengthSeconds * 2); // x2 fudge factor for any network lag
```
joll90, line 208: @thanaParis Should be a separate configuration, not a hardcoded factor.

00_Base/src/ocpp/model/enums/index.ts
```diff
@@ -188,7 +188,8 @@ export const enum DataEnumType {
   boolean = "boolean",
   OptionList = "OptionList",
   SequenceList = "SequenceList",
-  MemberList = "MemberList"
+  MemberList = "MemberList",
+  passwordString = "passwordString"
```
joll90, line 192: @thanaParis This file is generated and might be overwritten when the toolchain is ran again, do we know why this was not in the OCPP 2.0.1 enums as per specification?

04_Transaction/src/module/module.ts
```diff
@@ -182,4 +207,20 @@ export class TransactionModule extends AbstractModule {
...
   }
+
+  @AsHandler(CallAction.MeterValues)
+  protected async _handleMeterValues(
+    message: IMessage<MeterValuesRequest>,
+    props?: HandlerProperties
+  ): Promise<void> {
+    this._logger.debug("MeterValues received:", message, props);
+
+    // TODO: Add meterValues to transactions
```
joll90, line 218: @thanaParis Are we not at least storing the `meterValues` in a separate table?

joll90 (changes requested):
@thanaParis I left a few comments to address, the OCTT test results show the confidence in logic changes. Thanks!

00_Base/src/interfaces/centralsystem.ts
thanaParis: It's still optional, however now it has a default value, which is necessary because it's a protocol violation for it to be empty, it must be at least an empty object.

00_Base/src/ocpp/model/enums/index.ts
thanaParis: As far as I know it's an oversight. They are also missing identifierString, which is fairly important as it enforces a validation (charging station ids CANNOT have ':' in them) that breaks the protocol if not followed.

04_Transaction/src/module/module.ts
thanaParis: They would be inaccessible if we did, there's only data endpoints for fetching Transactions that have meter values not just meter values. MeterValues is deprecated in 2.0.1, only included for ease of upgrading from 1.6 systems. MeterValue data processing is on my fast-follow list after establishing core & advanced security.

joll90 (approved):

Issue #24, opened 2023-12-12
Project Structure Refactor


thanaParis:
The package structure of CitrineOS as it currently exists has issues:

- The numbering is not intuitive--01-05 are 'modules', designed to be separately deployable and independently scaled if needed, 00, 10, and 99 are dependencies, and 50 is an implementation of the abstract server design created to be locally deployable. The numbering doesn't communicate or lend itself to this structure well.
- The underscores cause problems. AWS does not allow underscores in service names.
- The naming doesn't make clear sense. Provisioning, Authorization, Availability, and Transactions are OCPP 2.0.1 functional blocks, but Monitoring is not. Additionally, creating a module per functional block doesn't make sense. It would be too many modules, and some messages that ought to be together, such as Authorize and the LocalAuthorizationList-related messages, are in different functional blocks. Some messages are even used in multiple functional blocks, which the subscription-based design of CitrineOS does not support.

I have a proposed restructuring.
Firstly, remove the number prefix from 01-05 and 50. Change Data and Util to have numbering reflecting their build order; i.e. Base is 00 as it is built first, Data is 01 because it is reliant on Base and built second, Util is 03 because it is reliant on Base and Data and built 3rd.
Second, use the following module definitions (broken down by module then by the messages they would handle):

1. User Control
   - Reservation
     - ReserveNow
     - ReservationStatusUpdate
     - CancelReservation
   - Commands
     - RequestStartTransaction
     - RequestStopTransaction
     - UnlockConnector
   - Authorization
     - Authorize
     - GetLocalListVersion
     - SendLocalList
     - ClearCache
2. Configuration
   - Charger requests configuration
     - BootNotification
     - Heartbeat
   - Charger receives configuration
     - ChangeAvailability
     - SetNetworkProfile
     - Reset
     - Display
       1. GetDisplayMessages
          - NotifyDisplayMessages
       2. SetDisplayMessage
       3. ClearDisplayMessage
     - Firmware
       1. PublishFirmware
          - PublishFirmwareStatusNotification
       2. UnpublishFirmware
       3. UpdateFirmware
          - FirmwareStatusNotification
   - Misc
     - DataTransfer (rejection only, see test case TC_P_02)
     - TriggerMessage
3. Certificates
   - CertificateSigned
   - DeleteCertificate
   - Get15118EVCertificate
   - GetCertificateStatus
   - GetInstalledCertificateIds
   - InstallCertificate
   - SignCertificate
4. Reporting
[...]

PR #37, opened 2024-01-24
Rc 1.0.0

thanaParis:
First full release!
Major changes
- [Project structure refactored](https://github.com/citrineos/citrineos-core/issues/24)
  - Modules renamed and moved under new '03_Modules' folder
  - 50_Server renamed to Server
- Swarm implementation added
  - In folder Swarm/, docker image created for each module as an independent server
  - Utilizes Redis cache
  - README updated
- Multiple websocket support with different SecurityProfiles added
  - Security profile 2 & 3 now supported (certificates required not included, as that would defeat the security, they can be generated with tools such as openssl)
  - SignCertificate handling for ChargingStationCertificate added

Minor changes
- CallErrors now returned when messages are mishandled by central system or module
  - CallError now thrown for all unhandled messages
- New RetryMessageError introduced which, when thrown from a handler, will cause the message to be returned to the message broker by skipping acknowledgement (or via explicit nack depending on queue implementation)
  - RetryMessageError thrown when new central system initiated Call cannot be sent to station because of an ongoing Call, allowing it to be retried
- json-schema-processor.js adjusted:
  - OcppRequest && OcppResponse interfaces now automatically added to generated classes
  - PasswordString added to DataEnumType due to spec erroneously missing it in part 3
  - license header added to generated classes
  - README updated to describe usage of script
- SecurityEventNotifications now handled and stored in db
- Health check added to central system in server/swarm

Bugfixes
- License headers altered for all files to include SPDX license reference & new copyright statement
- SystemConfig altered to hide database password


joll90 (approved):
🚀

ChrisWeissmann (approved):
I have looked at most of this from other PRs. So looks good to me :)

PR #38, opened 2024-01-25
fixing issue where expiration seconds not truly optional

thanaParis:
https://github.com/citrineos/citrineos-core/pull/35#discussion_r1465119322


ChrisWeissmann (approved):
Looks good to me. Thanks for following up :)

PR #46, opened 2024-03-13
feat: adding dependency injection

elliot-sabitov:
feat: adding tsyrnge as a base package dependency and moving config and cache into base and enabling dependency injection of config and cache into other modules
chore: creating helper npm run clean script
chore: renamed AbstractModule to BaseModule because it did not utilize any abstract methods
feat: getting server to run with dependency injection of SystemConfigService and CacheService as well as their needed child dependencies.
feat: moving env var configs back to server as discussed where via SystemConfigInput provider we are able to inject inputConfig depending on process.env.APP_ENV in Server.
feat: adjusted swagger logo to support base64 encoded image which may simplify image retrieval
chore: turning off no-explicit-any


elliot-sabitov:
@thanaParis the current setup with the Server module simply does not work as it is not the final set up and does not allow for modules to be initialized independently of each other without needing to import other modules.

This PR serves as an example demonstrating how dependency injection can be used to simplify the code base. I agree that the configuration files are implementation specific, but I would imagine that these values would be read from environment variables or some input files, but there still need to be defaults. The in memory config itself though, needs to be injectable via DI so it can be accessible throughout the code base, separating where the values came from against access to the store to read/update those values within the rest of the code, the latter of which would make sense to come from base module. 

In general we may want to use dependency injection all throughout the implementation to prevent tedious constructor management.  

Regarding the formatting, I do apologize. I will be more mindful of the macros that I have running that apply formatting and optimizing imports on save. As we have discussed before, we will look to set up an automated process that applies prettier throughout the code base so that we can all stay aligned. Aside from the new files and files that were moved around, for the changes in this PR that do have these formatting changes, if you can please review it keeping in mind that there were no changes in the code, only in the constructors and imports 🙏

thanaParis:
> @thanaParis the current setup with the Server module simply does not work as it is not the final set up and does not allow for modules to be initialized independently of each other without needing to import other modules.

The modules are already not co-dependent. none of them rely on each other. If you mean to ask how CitrineOS would use a single system config with the modules as separate services, that has already been taken care of. there are REST endpoints on each module to push the system config to so that it can be updated and managed from the central system.

> I agree that the configuration files are implementation specific, but I would imagine that these values would be read from environment variables or some input files, but there still need to be defaults.

There are already defaults in the zod schema definition.

System Config should not be injected like this.

00_Base/.eslintrc.js
```diff
@@ -1,7 +1,10 @@
...
-  extends: ['eslint:recommended', 'plugin:@typescript-eslint/recommended'],
-  parser: '@typescript-eslint/parser',
-  plugins: ['@typescript-eslint'],
+  extends: ["eslint:recommended", "plugin:@typescript-eslint/recommended"],
+  parser: "@typescript-eslint/parser",
+  plugins: ["@typescript-eslint"],
   root: true,
-};
\ No newline at end of file
+  rules: {
```
elliot-sabitov, line 7: this change can be ignored as it will be addressed in the prettier/eslint PR but TS was complaining about an explicit any and I figured I would just adjust it here

00_Base/src/config/config/system.config.service.ts
```diff
@@ -0,0 +1,115 @@
...
+
+@singleton()
+export class SystemConfigService {
+  systemConfig: SystemConfig;
+
+  constructor(
+    @inject("SystemConfigInput")
+    private readonly inputConfig?: SystemConfigInput
+  ) {
+    const config = merge<SystemConfigInput>(
```
elliot-sabitov, line 109: trying to introduce defaults for the config so that usages in Server and other places can override only the settings that are desired without having to pass in the entire config

Server/src/assets/logo.ts
```diff
@@ -0,0 +1,2 @@
+export const logo: string =
```
elliot-sabitov, line 1: Easier way to manage importing an image without needing `fs` by just having it as a string which can be useful when setting default images in the codebase

thanaParis:
Outdated.

PR #48, opened 2024-03-14
Made unix-init-install-all.sh executable

folkengine:
This PR makes the file `Server/unix-init-install-all.sh` executable, so that when someone checks the file out step #3 of the instructions (`./unix-init-install-all.sh`) works as expected.


thanaParis (approved):
Good catch! Thanks @folkengine

PR #49, opened 2024-03-14
Feature/module receiver refactor

thanaParis:
Refactor to put MessageHandler's handle() func and get/set for module into AbstractMessageHandler, simplifying code at implementation level.

Corrected accidental use of PubSub instead of RabbitMQ.


lydiazcheng (approved):
The code looks good. Only a minor point from my side.

PR #54, opened 2024-03-19
Disable logging colors for cloud

ChrisWeissmann:
Using a new environment variable `DEPLOYMENT_TARGET` to set not logging with colors to fix ascii encoding issues with cloud watch.
I put this where the logger is initialized, further down in the implementation only gets the sub logger.


thanaParis (approved):
Looks good!

PR #55, opened 2024-03-19
Feat: Check request size for get vars and get custom repost

lydiazcheng:
Add a check for size of the request according to 2.0.1

1. GetVars Screenshot 2024-03-19 at 10 50 33 AM
2. GetCustomReport  Screenshot 2024-03-19 at 10 50 07 AM


thanaParis (approved):
Looks good!

PR #57, opened 2024-03-20
Feature/directus message api flows

thanaParis:
Adds flows to the Directus Locations collection to send OCPP messages to the charger via Message API.

SystemConfig extended to configure Directus access and flow creation.

Directus flows automatically generated without issue using preseeded postgress image. In order to work from scratch, the Locations DB needs to be added by hand. Locations will be added to the code base directly in a future PR.

Swarm mode works, however you must type in the appropriate url for the call. Example: to send a Reset, you will need to set the url to "http://configuration:8084" because 'configuration' is the name of the docker image that spins up the configuration module, which has the Message API's Reset endpoint (port is set in SystemConfig, docker image name is in docker-compose.yml). If possible, adding the docker image names to the list of options in Swarm mode would be a nice QoL enhancement.

Pictures:
image

image


Swarm/src/index.ts
```diff
@@ -79,6 +79,12 @@ class CitrineOSServer {
             initSwagger(this._config, this._server);
         }
 
+        // Add Directus Message API flow creation if enabled
+        if (this._config.util.directus?.generateFlows) {
+            const directusUtil = new DirectusUtil(this._config, this._logger);
+            this._server.addHook("onRoute", directusUtil.addDirectusMessageApiFlowsFastifyRouteHook.bind(directusUtil));
+        }
```
ChrisWeissmann, line 86: From previous sessions, I think we would benefit to know when the directus flows setup finished. I looked at the docs here: https://fastify.dev/docs/latest/Reference/Hooks/#onready and it seems that Fastify has a hook when it's stood up and all routes are registered. I think that's the place to log a success. 

```suggestion
            this._server.addHook('onReady', async () => {
                this._logger.info('Directus actions initialization finished');
            });
        }
```

02_Util/src/util/directus.ts
```diff
@@ -0,0 +1,245 @@
...
+        let client;
+        if (this._config.util.directus?.token) { // Auth with static token
+            client = createDirectus(`http://${this._config.util.directus?.host}:${this._config.util.directus?.port}`)
+                .with(staticToken(this._config.util.directus?.token)).with(rest());
+        } else if (this._config.util.directus?.username && this._config.util.directus?.password) { // Auth with username and password
+            client = createDirectus<Schema>(`http://${this._config.util.directus?.host}:${this._config.util.directus?.port}`)
+                .with(authentication()).with(rest());
+            this._logger.info(`Logging into Directus as ${this._config.util.directus.username}`);
+            client.login(this._config.util.directus.username, this._config.util.directus.password);
+        } else { // No auth
```
ChrisWeissmann, line 33: Can this actually happen? Since we set defaults in the schema above. I don't think there is a way to clear them.

ChrisWeissmann (commented):
Really like this PR! 😄

02_Util/src/util/directus.ts
thanaParis: Right now, these defaults don't apply; there's an issue with Zod regarding how default works with nested optional objects. This is a good point though--there shouldn't be a default username/password. I'll remove it!

02_Util/src/util/directus.ts
```diff
@@ -0,0 +1,245 @@
...
+            client = createDirectus<Schema>(`http://${this._config.util.directus?.host}:${this._config.util.directus?.port}`)
+                .with(rest());
+        }
+        this._client = client;
+    }
+
+    public addDirectusMessageApiFlowsFastifyRouteHook(routeOptions: RouteOptions) {
```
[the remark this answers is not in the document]
thanaParis: added much more explanation in the comments for this method!

Server/src/index.ts
```diff
@@ -79,6 +79,12 @@ class CitrineOSServer {
             initSwagger(this._config, this._server);
         }
 
+        // Add Directus Message API flow creation if enabled
+        if (this._config.util.directus?.generateFlows) {
+            const directusUtil = new DirectusUtil(this._config, this._logger);
+            this._server.addHook("onRoute", directusUtil.addDirectusMessageApiFlowsFastifyRouteHook.bind(directusUtil));
+        }
```
ChrisWeissmann, line 86: Sorry, should have put it in here as well right away. Had the below suggestion on the Swarm index.ts to log when then fastify server is ready and has also set up all the Directus flows

```suggestion
            this._server.addHook('onReady', async () => {
                this._logger.info('Directus actions initialization finished');
            });
        }
```

ChrisWeissmann (approved):
Looks good to me! Can't wait to have those buttons :)

PR #58, opened 2024-03-21
Make node version consistent

ChrisWeissmann:
fix: align engine node version with docs and 01_Data


thanaParis (approved):
Approved! Good catch

PR #59, opened 2024-03-21
Add docker image release action and PR build action

ChrisWeissmann:
Added an action that when a tag that follows vx.x.x is pushed it will create an image and push it to the GitHub container registry.
Added an action that builds the docker image 1x when a PR is opened, as a sanity check that at least the current set up still builds.


thanaParis (approved):
Looks good!!

PR #63, opened 2024-03-27
Feature/workspaces

elliot-sabitov:
feat: Utilizing npm workspaces to simplify multi npm package local dependencies
feat: Adjusting Docker related files to expose debugger port
chore: Created run configuration support files to run Docker and attach debugger


package.json
```diff
@@ -0,0 +1,32 @@
+{
+  "name": "@citrineos/workspace",
```
ChrisWeissmann, line 2: I wonder if we want to instead have this in the `server` and `swarm` directory.

ChrisWeissmann (commented):
The test build in the GitHub workflow seems broken now. We should probably fix this.

01_Data/tsconfig.json
```diff
@@ -1,20 +1,17 @@
...
-    "**/__tests__/*"
+  "compilerOptions": {
+    "outDir": "./dist/",
+    "composite": true,
+    "rootDir": "./src"
+  },
+  "references": [
+    {
+      "path": "../00_Base"
+    }
```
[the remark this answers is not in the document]
elliot-sabitov: I took a look through what they have for `declarationMap` and it is not 100% clear to me if this will be helpful. In my case, I am able to jump to definitions between modules so long as I run the `npm i` first successfully

02_Util/src/server/example/server.ts
```diff
@@ -0,0 +1,702 @@
+// Copyright (c) 2023 S44, LLC
+// Copyright Contributors to the CitrineOS Project
+//
+// SPDX-License-Identifier: Apache 2.0
+/* eslint-disable */
+
```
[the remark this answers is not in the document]
elliot-sabitov: Yeah I think previously this was all in Base before, and I believe I moved it here because to build things properly because of the dependencies not being available where they needed to be, I moved it to Util as it made more sense. At some point Thana made this PR that moved this into Util - `https://github.com/citrineos/citrineos-core/pull/56` and this code is now here - https://github.com/citrineos/citrineos-core/blob/rc-1.1.0/02_Util/src/networkconnection/WebsocketNetworkConnection.ts - I will remove these files from this PR

package.json
elliot-sabitov: In theory, since we can set the paths it should be possible. Most examples I saw have a parent package.json and a `packages` folder that contains all of the modules and in the package.json they can just do ```"workspaces": [ "./packages/*]```. NPM then figures out all of the dependencies and creates a node_modules dir in the root which has all the shared dependencies and each module may, but does not need to, contain its own node_modules directory if they have any specific dependencies that arent used elsewhere. 

If we move this to Server, I would be concerned that we may run into some hick ups with the workspaces setup, but in theory should be possible. I would take the approach of just having `src/index.ts` to init the modules and eliminate the need for `server`. Swarm variation of server and various compositions can all be achieved by modifying env vars and docker files

01_Data/src/index.ts
```diff
@@ -4,4 +4,5 @@
 // SPDX-License-Identifier: Apache 2.0
 
 export * as sequelize from "./layers/sequelize";
-export * from "./interfaces";
\ No newline at end of file
+export * from "./interfaces";
+export {DeviceModelRepository, Boot, Component, Variable, VariableAttribute} from "./layers/sequelize"; // todo export better as these seem to be used in other modules
```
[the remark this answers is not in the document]
elliot-sabitov: I took a look -
```export * as sequelize from "./layers/sequelize";
export * from "./interfaces";
export {DeviceModelRepository, Boot, Component, Variable, VariableAttribute} from "./layers/sequelize";```
the first line `exports * as sequilize` so in other places you can then import sequelize from here, but I found that in other parts of the code were were trying to import these other things and the way we did it was like
`import {DeviceModelRepository} from '@citrineos/data/layers/sequilize`;`
and it somehow worked but in either case this is here so that these other classes that are being used elsewhere are available. I made this change as well as others to get builds to work properly =/

File: 00_Base/package.json [hunk 5]
```diff
@@ -37,5 +36,6 @@
     "tslog": "^4.9.2",
     "uuid": "^9.0.0",
     "zod": "^3.22.2"
-  }
+  },
+  "workspace": "../"
```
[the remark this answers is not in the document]
ChrisWeissmann: Ah got it. So in theory Base could have this omitted. The connection I hadn't made was how the dependancies work with each other. E.g. Data depending on Base and setting the workspace to `../` But since you need it to be part of the workspace, I understand.

.idea/runConfigurations/Attach_Debugger.xml
thanaParis, line 1: We should remove the .idea folder. IDE-specific code doesn't belong in the main repo. Feel free to keep this on a fork if you want!

00_Base/package.json
```diff
@@ -2,17 +2,17 @@
...
-  "types": "lib/index.d.ts",
+  "main": "dist/index.js",
+  "types": "dist/index.d.ts",
   "files": [
-    "lib"
+    "dist"
   ],
   "scripts": {
     "prepublish": "npx eslint ./src",
-    "prepare": "npm run build",
```
thanaParis, line 12: why are we removing the 'prepare' script? `prepare` always runs when you execute `npm install`. needing to execute `npm run compile` after `npm install` is a definite downgrade.

00_Base/src/interfaces/messages/index.ts
```diff
@@ -33,6 +34,14 @@ export enum EventGroup {
   Transactions = 'transactions',
 }
 
+export const eventGroupFromString = (source: string): EventGroup => {
```
thanaParis, line 37: Why is this in this PR? There are enough changes need just to move to using an npm workspace without adding additional features. Please reduce the changes per PR where possible.

01_Data/src/index.ts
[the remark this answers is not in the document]
thanaParis: this change should be in its own pr. Exporting only those specific classes circumvents the problem rather than solving it. At this point I think it is clear we should move away from the 'sequelize' export and individually export the classes in 02_Data/layers/sequelize.

03_Modules/Configuration/src/module/api.ts
thanaParis, line 1: is this just a formatting change?

03_Modules/Monitoring/src/module/api.ts
thanaParis, line 1: is this just a formatting change?

README.md
```diff
@@ -41,45 +46,74 @@ Before you begin, make sure you have the following installed on your system:
...
+1. Build project from root dir:
 
    ```shell
-   ./unix-init-install-all.sh
+   npm run build
    ```
 
-1. Start the server and its supporting infrastructure with:
+1. The docker container should be initialized by running `docker-compose -f ./docker-compose.yml up -d` or
+   by using the IntelliJ `Server` Run Configuration which was created for this purpose.
```
thanaParis, line 62: The instructions should `cd` into `Server/` and just run `docker compose up`. Also, we should not not be prescriptive about IDEs. The references to IntelliJ need to be removed. It would be easier to understand the changes in this file if they weren't combined with changes to the formatting.

Server/.dockerignore
thanaParis, line 1: Moving the docker files out of Server/docker/ should be in its own PR.

Swarm/.eslintrc.json
thanaParis, line 1: Please put all things related to removing Swarm in their own PR.

Server/docker-compose.yml
```diff
@@ -26,24 +26,57 @@ services:
...
       retries: 3
   citrine:
     build:
-      context: ../
-      dockerfile: ./Server/docker/Dockerfile
+      context: ..
+      dockerfile: ./Server/local.Dockerfile
+    volumes:
+      - ../:/usr/local/apps/citrineos
+      - /usr/local/apps/citrineos/node_modules
```
thanaParis, line 49: why are these volumes being added?

thanaParis (changes requested):
This PR is doing too much at the moment. I'd appreciate if we could focus on what's needed to move towards an npm workspace, and remove references to IntelliJ as we previously discussed. The tsconfig inheritance changes can stay--I think they're related enough.

00_Base/package.json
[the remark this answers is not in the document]
elliot-sabitov: I would omit it the other way, if you call npm commands from the root then you can omit it in the children. Including it in both I would say allows you to run npm commands from both the parent and the child module folders

.idea/runConfigurations/Attach_Debugger.xml
elliot-sabitov: @thanaParis can we talk about this. I think a better approach would be to commit the vscode/idea/etc IDE specific stuff because it can really speed up dev time!

elliot-sabitov: Just to note - these are run configurations that allow you to run stuff with a simple click - this isnt like code style or any other IDE specific configs, these are to run the various commands

00_Base/package.json
elliot-sabitov: we dont need to run compile, the `npm run build` in root does everything

00_Base/src/interfaces/messages/index.ts
elliot-sabitov: @thanaParis I am trying to make a lot happen in a small period of time here, if needed I can walk you through the PR in detail and explain these, but if you feel that breaking these out into separate PRs, as it would require additional effort, then I am open to a discussion. The changes I am making tend to have a good reason and all of the changes are open for discussion. I can break these out into multiple PRs but in this case, this was needed as EventGroup is being parsed from the process.env which was the setup prior to this PR

03_Modules/Configuration/src/module/api.ts
elliot-sabitov: yes I ran the app after the changes so if you see any imports being adjusted it is most likely my automated macro to optimize imports, but I know there is still formatting stuff here which will be solved with prettier/lint, but if you see anything that is just imports then yes please assume that it is working properly. In many cases I came across improts like `import { VariableAttribute } from "@citrineos/data/lib/layers/sequelize";` which needed to be fixed

README.md
elliot-sabitov: We should talk about having instructions for most common IDEs

Server/.dockerignore
elliot-sabitov: ? wait I moved the docker files back into /Server...

Swarm/.eslintrc.json
elliot-sabitov: why is this needed? If we are okay with moving away from Swarm, then why would we spend time on that?

03_Modules/Monitoring/src/module/api.ts
elliot-sabitov: yes and please note:

```
import { Variable, Component } from '@citrineos/data/lib/layers/sequelize';
import {getBatches, getSizeOfRequest} from "@citrineos/util/lib/util/parser";
```
being removed

00_Base/package.json
thanaParis: understood!

README.md
elliot-sabitov: adjusted readme to mention `Server/` dir

Server/docker-compose.yml
thanaParis: Makes sense! Please add this improvement to Swarm as well.

elliot-sabitov: Yeah I was thinking about doing this but left it as though the Swarm docker setup would be used as a "deployed" environment and not something that you would really run locally. I can agree that it would also make sense to have a Swarm version that has the same volumes added, but then don't we also need another file added that does not mount the volumes for cases where you just want to run the setup in a deployed environment?

README.md
thanaParis: Noted! [I've opened a Github discussion.](https://github.com/orgs/citrineos/discussions/18)

Server/docker-compose.yml
elliot-sabitov: I am going to see how we can do this easily because I really don't wish to copy that long list everywhere, it is really meant more for ease of development purposes, but if we do want to run swarm locally in dev as we may really want to do, then we need this in swarm for sure as well

thanaParis: There are other considerations for Swarm, for sure. I can see leaving this out of it for now. Please add back the Swarm folder. We can discuss how best to carry forward the intent behind Swarm in future PRs. [I have created a discussion topic for it.](https://github.com/orgs/citrineos/discussions/19)

Swarm/src/config/envs/docker.ts
```diff
@@ -51,7 +51,8 @@ export function createDockerConfig() {
             transactions: {
                 endpointPrefix: "transactions",
                 host: "0.0.0.0",
-                port: 8089
+                port: 8089,
+                costUpdatedInterval: 60
```
elliot-sabitov, line 55: is this a bug, my diff doesnt show any changes in Swarm

thanaParis: it does appear to be a bug! I'm not seeing a diff here either

thanaParis (approved):

PR #64, opened 2024-03-27
Feature/station connection status

thanaParis:
Work In Progress, opening to review early

Changes:

1. `DirectusExtensions/charging-stations-bundle` added
  a) Custom display "Present or True Count" added, which displays total number of rows with a value that is present, if non-boolean column, or true, if boolean column
  b) Custom endpoint "POST /charging-stations/update-station-status" which updates isOnline in response to events POSTed from CitrineOS's OcppRouter Subscription API
  c) Custom hook "On Create Charging Station" which subscribes new stations to connect and close events via CitrineOS's OcppRouter Subscription API

Misc:

Missing @Column annotation added in ChargingStations data model
Unused images in Server/data/directus/uploads removed
A config.js file was added to Directus via CONFIG_PATH env var to provide necessary context for custom extensions, such as the OcppRouter's url, subscription api path, Directus's url, and the custom update-station-status api path


directus-extension-charging-stations-bundle/src/hook-on-create-charging-station/index.ts
```diff
@@ -0,0 +1,33 @@
...
+// SPDX-License-Identifier: Apache 2.0
+
+import { defineHook } from '@directus/extensions-sdk';
+
+export default defineHook(({ action }, { env }) => {
+
+    action('ChargingStations.items.create', (input) => {
+        console.log("Subscribing " + input.key + " to connect and close events");
+        const stationId = input.key;
+        const subscriptionUrl = `http://citrine:8080/data/ocpprouter/subscription`;
```
ChrisWeissmann, line 12: Is this something the we should externalize and make configurable somehow?

directus-extension-charging-stations-bundle/tsconfig.json
```diff
@@ -0,0 +1,28 @@
+{
+	"compilerOptions": {
```
ChrisWeissmann, line 2: I think this is very nitpicky: Should our tsconfig be consistent of what is in the other packages? Or does directus need some special things?

ChrisWeissmann (commented):
I want to be sure I understand how this works: 
The directus extensions offers an API that can get hit when a charger opens a web socket connection. 
To get this update, on commissioning/adding the charger to the location we make a http request to citrine on the endpoint provided by `@AsDataEndpoint(Namespace.Subscription, HttpMethod.Put)` 
This keeps track of the subscriptions in an array with the associated callback actions to take. 
What I don't quite grasp is if we are going to run into a problem scaling horizontally, as the callbacks seem to be held in an array in memory. 

Is this correct?

directus-extension-charging-stations-bundle/src/hook-on-create-charging-station/index.ts
thanaParis: Already the plan! The config.js I mention in the 'WIP' section is for this.

directus-extension-charging-stations-bundle/tsconfig.json
thanaParis: this is automatically generated when you use the command `npx create-directus-extension@latest` to create an empty directus extension package of your choice of extension type. I doubt it *needs* any of this though. I'll make it match our other tsconfigs

thanaParis:
> What I don't quite grasp is if we are going to run into a problem scaling horizontally, as the callbacks seem to be held in an array in memory.
You do understand how it works, and this is a good point. I'll need to make the subscription api persistent.

Server/data/directus/uploads/2937fa22-293e-4aae-805e-ad83c7843a6e.png
lydiazcheng, line 1: Images look the same. What is the use case of these images?

thanaParis: these images are used for the Directus UI to have CitrineOS branding. Two of the images are the citrine crystal, of different sizes, then the others are the orange pixel 'glow' that goes behind the citrine crystal.

thanaParis:
@ChrisWeissmann PR opened to make subscriptions persistent: https://github.com/citrineos/citrineos-core/pull/66

directus-extension-charging-stations-bundle/tsconfig.json
thanaParis: Updated!

elliot-sabitov (changes requested):

directus-extension-charging-stations-bundle/package.json
```diff
@@ -0,0 +1,50 @@
...
+			"source": "src/endpoints-charging-stations/index.ts"
+		}, {
+			"type": "hook",
+			"name": "On Create Charging Station",
+			"source": "src/hook-on-create-charging-station/index.ts"
+		}],
+		"host": "^10.10.0"
+	},
+	"scripts": {
+		"build": "directus-extension build",
```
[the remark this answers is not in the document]
elliot-sabitov: more common variation for other extensions that may not require the 2 app/api separations: ```"scripts": {
		"build": "directus-extension build",
		"build-dev": "directus-extension build --no-minify --sourcemap -t interface -i src/index.ts -o ../../../../extensions/interfaces/name/index.js",
		"dev": "directus-extension build --watch --no-minify --sourcemap -t interface -i src/index.ts -o ../../../../extensions/interfaces/name/index.js",
		"link": "directus-extension link"
	}``` with the the directus extensions settings: ```"directus:extension": {
		"type": "interface",
		"path": "../../../../extensions/interfaces/name/index.js",
		"source": "src/index.ts",
		"host": "10...."
	},```

directus-extension-charging-stations-bundle/package.json
```diff
@@ -0,0 +1,50 @@
...
+		"directus-extension-bundle"
+	],
+	"type": "module",
+	"files": [
+		"dist"
+	],
+	"directus:extension": {
+		"type": "bundle",
+		"path": {
+			"app": "dist/app.js",
```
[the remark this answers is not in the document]
thanaParis: What do you mean? It works as-is, I'm not sure what you're suggesting.

DirectusExtensions/charging-stations-bundle/src/endpoints-charging-stations/index.ts
```diff
@@ -0,0 +1,38 @@
+// Copyright Contributors to the CitrineOS Project
+//
+// SPDX-License-Identifier: Apache 2.0
+
+import { defineEndpoint } from '@directus/extensions-sdk';
+
+export default defineEndpoint({
+    id: 'charging-stations',
+    handler: (router, { database }) => {
+        router.post('/update-station-status', async (req, res) => {
```
elliot-sabitov, line 10: @thanaParis why dont you just use one of the existing endpoints to update the record? Are you sure this needs to be a custom endpoint?

DirectusExtensions/charging-stations-bundle/src/hook-on-create-charging-station/index.ts
```diff
@@ -0,0 +1,35 @@
...
+import { defineHook } from '@directus/extensions-sdk';
+
+export default defineHook(({ action }, { env }) => {
+
+    action('ChargingStations.items.create', (input) => {
+        console.log("Subscribing " + input.key + " to connect and close events");
+        
+        const stationId = input.key;
+        const subscriptionUrl = `${env.CITRINEOS_URL}${env.CITRINEOS_SUBSCRIPTION_API_PATH}`;
+        const updateStationStatusUrl = `${env.DIRECTUS_URL}${env.DIRECTUS_CHARGING_STATION_UPDATE_STATUS_PATH}`;
```
elliot-sabitov, line 14: perhaps we can connect, if you are calling an endpoint to update the DB, then there may be a better approach where you can make DB queries directly from here

directus-extension-charging-stations-bundle/src/endpoints-charging-stations/index.ts
```diff
@@ -0,0 +1,38 @@
+// Copyright Contributors to the CitrineOS Project
+//
+// SPDX-License-Identifier: Apache 2.0
+
+import { defineEndpoint } from '@directus/extensions-sdk';
+
+export default defineEndpoint({
+    id: 'charging-stations',
+    handler: (router, { database }) => {
+        router.post('/update-station-status', async (req, res) => {
```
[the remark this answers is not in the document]
thanaParis: This is a good consideration for when we may have many endpoints in the future. Note, `endpoints-charging-stations` is not in the path because I declared an id for this endpoint. The path is `/charging-stations/update-station-status`, constructed from the id and the url in the router call.

DirectusExtensions/charging-stations-bundle/src/endpoints-charging-stations/index.ts
thanaParis: There is no existing endpoint that accepts the output of the subscription api.

directus-extension-charging-stations-bundle/package.json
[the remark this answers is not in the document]
thanaParis: > this also ensures that when you run locally it will include a sourcemap and a watcher while also compiling it to the correct extensions directory

This would be quite nice!
If I'm understanding you correctly, the only way to support this would be to break out the bundle into 3 separate extensions, then use the `directus:extension.path` param to place the built code directly into the file structure used by Directus. This wouldn't be able to build to both `Server` and `Swarm`, correct? Are there other ways to accomplish this?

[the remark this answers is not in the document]
elliot-sabitov: No this would be only for Directus. Looking at your PR without even considering the logic that you are trying to achieve, I think I have some points in terms of project structure and sharing code between the extensions as well as building and debugging that I would love to run by you!

Server/docker-compose.yml
```diff
@@ -61,7 +61,7 @@ services:
       - 8055:8055
     volumes:
       - ./data/directus/uploads:/directus/uploads
-      - ./data/directus/extensions:/directus/extensions
+      - ../directus-extension-charging-stations-bundle:/directus/extensions/directus-extension-charging-stations-bundle
```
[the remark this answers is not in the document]
elliot-sabitov: that isnt what I am saying - the bundle name stays as is, but you put all your bundles and other extensions in one container folder so that you can do `directus-extensions/*:/directus/extensions`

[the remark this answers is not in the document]
elliot-sabitov: it is best to have `/directus-extensions` in root where you can then place /bundles/charging-stations-bundle whatever else... when running locally, these can all build to the same `/extensions` directory that is mounted that can with sourcemaps then support breakpoints. This way, you can load any number of extensions quickly and have hot reload as directus can pull new extensions if anything in that one folder changes...

DirectusExtensions/charging-stations-bundle/src/hook-on-create-charging-station/index.ts
elliot-sabitov: this is already running on Directus BE service, why would you then call another directus BE service to then perform the logic rather then handing what you need here?

DirectusExtensions/charging-stations-bundle/src/endpoints-charging-stations/index.ts
elliot-sabitov: You can use `/items/ChargingStations` endpoints that are available in Directus by default to trigger:

```
await database('ChargingStations')
                    .where({ id: stationId })
                    .update({ isOnline });
```

so what is the benefit of having this as a separate endpoint?

thanaParis: You can review the structure of the subscription api in OcppRouter/router.ts. They don't match, unfortunately, and shouldn't since they serve a wider purpose than just providing Directus with updates.

DirectusExtensions/charging-stations-bundle/src/hook-on-create-charging-station/index.ts
thanaParis: When you review the code, you'll see that the directus url is not called here. It is given to CitrineOS's subscription api, which then calls it when there are relevant updates.

elliot-sabitov: I think we should have a quick discussion here - in this case I was talking about line 14 which is for the `updateStationStatusUrl`

DirectusExtensions/charging-stations-bundle/src/display-true-count/display.vue
```diff
@@ -0,0 +1,55 @@
+<template>
+	<span>{{ calculatedValue }}{{ totalPrefix }}{{ total }}{{ suffix }}</span>
+</template>
+
+<script lang="ts">
+// Copyright Contributors to the CitrineOS Project
+//
+// SPDX-License-Identifier: Apache 2.0
```
lydiazcheng, line 8: Very minor: how about moving this part to the top?

DirectusExtensions/charging-stations-bundle/src/display-true-count/display.vue
```diff
@@ -0,0 +1,55 @@
...
+		const total = props.showTotal ? props.value.length : null;
+
+		props.value.forEach(item => {
+			const columns = props.column.split('.');
+
+			columns.forEach(col => {
+				item = item[col];
+			});
+
+			if (item && (item === 'true' || item !== 'false')) {
```
lydiazcheng, line 46: `item === 'true'` is not necessary? 
```suggestion
			if (item && item !== 'false') {
```

lydiazcheng (approved):
The change looks good!

DirectusExtensions/charging-stations-bundle/src/display-true-count/display.vue
thanaParis: prevents the code from compiling, unfortunately.

PR #66, opened 2024-03-29
Feature/persistent admin api

thanaParis:
Alters Subscriptions to be persistent in SQL db, added GET and DELETE CRUD endpoints to api for Subscriptions now that they are db entities.

N.B.: The PUT Subscription was altered to a POST since the requests are not Idempotent (calling the endpoint repeatedly will create many new entries rather than updating the first created, due to the database id being generated in the data layer).

### DIFF
--- a/03_Modules/OcppRouter/src/module/api.ts
+++ b/03_Modules/OcppRouter/src/module/api.ts
@@ -39,7 +39,7 @@ export class AdminApi extends AbstractModuleApi<MessageRouterImpl> implements IA
      * @return {Promise<number>} The id of the created subscription.
      */
     @AsDataEndpoint(Namespace.Subscription, HttpMethod.Post)
-    async putSubscription(request: FastifyRequest<{ Body: Subscription }>): Promise<number> {
+    async postSubscription(request: FastifyRequest<{ Body: Subscription }>): Promise<number> {
         return this._module.subscriptionRepository.create(request.body as Subscription).then((subscription) => subscription?.id);
     }
 
--- a/03_Modules/OcppRouter/src/module/api.ts
+++ b/03_Modules/OcppRouter/src/module/api.ts
@@ -8,6 +8,7 @@ import { FastifyInstance, FastifyRequest } from 'fastify';
 import { ILogObj, Logger } from 'tslog';
 import { IAdminApi } from './interface';
 import { MessageRouterImpl, Subscription } from './router';
+import { ChargingStationKeyQuerystring, ModelKeyQuerystring } from '@citrineos/data';
 
 /**
  * Server API for the Certificates module.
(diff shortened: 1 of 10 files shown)


03_Modules/OcppRouter/src/module/router.ts
```diff
@@ -252,6 +262,71 @@ export class MessageRouterImpl extends AbstractMessageRouter implements IMessage
...
+                this._logger.debug(`Added onConnect callback to ${subscription.url} for station ${connectionIdentifier}`);
+            }
+            if (subscription.onClose) {
+                this._onCloseCallbacks.get(connectionIdentifier)!.push((info?: Map<string, string>) => this._subscriptionCallback({ stationId: connectionIdentifier, event: 'closed', info: info }, subscription.url));
+                this._logger.debug(`Added onClose callback to ${subscription.url} for station ${connectionIdentifier}`);
+            }
+            if (subscription.onMessage) {
+                this._onMessageCallbacks.get(connectionIdentifier)!.push(async (message: string, info?: Map<string, string>) => {
+                    if (!subscription.messageRegexFilter || new RegExp(subscription.messageRegexFilter).test(message)) {
+                        return this._subscriptionCallback({ stationId: connectionIdentifier, event: 'message', origin: MessageOrigin.ChargingStation, message: message, info: info }, subscription.url)
```
lydiazcheng, line 290: Do we need to differentiate the event 'onMessage' and 'sentMessage'?

lydiazcheng (approved):

03_Modules/OcppRouter/src/module/router.ts
```diff
@@ -34,20 +23,25 @@ export class MessageRouterImpl extends AbstractMessageRouter implements IMessage
     protected _handler: IMessageHandler;
     protected _networkHook: (identifier: string, message: string) => Promise<boolean>;
 
-    private _onConnectionCallbacks: ((identifier: string, info?: Map<string, string>) => Promise<boolean>)[] = [];
-    private _onCloseCallbacks: ((identifier: string, info?: Map<string, string>) => Promise<boolean>)[] = [];
-    private _onMessageCallbacks: ((identifier: string, message: string, info?: Map<string, string>) => Promise<boolean>)[] = [];
-    private _sentMessageCallbacks: ((identifier: string, message: string, error?: any, info?: Map<string, string>) => Promise<boolean>)[] = [];
+    // Structure of the maps: key = identifier, value = array of callbacks
```
ChrisWeissmann, line 26: Just to be sure here, and this is probably worth a comment:
Since web socket connection are stateful and tied to a single instance, we can put the callbacks into memory. And when a charger connects, it gets the callbacks from the DB and loads them into memory.
 
A small thing which probably isn't worth a deeper dive right now, but when sending the request to set up a callback, we need to be sure to know where the charger is connected to. Else, we could hit instance `A` but the charger is connected to instance `B` and since the callback state only gets refreshed on connection, instance `B` won't know about this until the charger drops and reconnects. But I think this is down the line.

ChrisWeissmann (approved):
I left some minor feedback. Other than that, looks good to me :)

03_Modules/OcppRouter/src/module/router.ts
thanaParis: Good question. It made sense to me to keep them separate since they occur at different places in the code. In terms of use case, I can see situations where you may only want one or the other, such as if you just want to monitor StatusNotificationRequest and the response (which is empty) is pointless.

03_Modules/OcppRouter/src/module/router.ts
thanaParis: Good point. I added a comment to the AdminApi class. I'm not sure what the correct way around this is, but I also agree we can leave it until later to fix.

03_Modules/OcppRouter/src/module/api.ts
```diff
@@ -29,87 +31,26 @@ export class AdminApi extends AbstractModuleApi<MessageRouterImpl> implements IA
...
-        return;
+    /**
+     * Creates a {@link Subscription}.
+     * Will always create a new entity and return its id.
+     *
+     * @param {FastifyRequest<{ Body: Subscription }>} request - The request object, containing the body which is parsed as a {@link Subscription}.
+     * @return {Promise<number>} The id of the created subscription.
+     */
+    @AsDataEndpoint(Namespace.Subscription, HttpMethod.Post)
+    async putSubscription(request: FastifyRequest<{ Body: Subscription }>): Promise<number> {
```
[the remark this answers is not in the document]
thanaParis: Fixed!

PR #67, opened 2024-04-01
Adding prettier and lint adjustments

elliot-sabitov:
feat: adding lint and prettier as dev dependencies in root package
feat: creating lint, lint-fix, and prettier run scripts
feat: creating github workflow to trigger lint
chore: adding lint and prettier run configs

### DIFF
--- a/package.json
+++ b/package.json
@@ -6,12 +6,17 @@
     "@types/node": "^20.11.20",
     "concurrently": "^8.2.2",
     "nodemon": "^3.1.0",
-    "typescript": "^5.3.3"
+    "typescript": "^5.3.3",
+    "prettier": "^3.2.5",
+    "eslint": "^8.48.0"
   },
   "scripts": {
     "install-all": "npm i --verbose",
     "clean": "rm -rf package-lock.json **/package-lock.json **/**/package-lock.json dist **/dist **/**/dist node_modules **/node_modules **/**/node_modules tsconfig.tsbuildinfo **/tsconfig.tsbuildinfo **/**/tsconfig.tsbuildinfo",
-    "build": "tsc --build --verbose"
+    "build": "tsc --build --verbose",
+    "lint": "npx eslint --ext .ts ./",
+    "lint-fix": "npx eslint --fix --ext .ts ./",
+    "prettier": "prettier --write ."
   },
   "dependencies": {
     "@citrineos/server": "1.0.0"
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -1,0 +1,28 @@
+name: Linter
+
+on:
+  pull_request:
+
+concurrency:
+  # Rerun check when the PR is updated, https://stackoverflow.com/a/72408109
+  group: ${{ github.workflow }}-${{ github.event.pull_request.number || github.ref }}
+  cancel-in-progress: true
+
+jobs:
+  linter:
+    name: Linter
+    runs-on: ubuntu-22.04 # todo: to prevent needing to run npm i we can have our own image where everything is installed
+    strategy:
+      matrix:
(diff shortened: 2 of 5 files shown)


lydiazcheng (commented):
Lint work flow failed with errors in the unrelated files of this PR.  How about we try to use  `eslint $(git diff --name-only --diff-filter=d HEAD **/*.ts | xargs)
` to check only the changed files? (I search this command from internet, haven't tried)
Another point looks confusing is that lint errors are only in the 01_Data module, such as "Strings must use singlequote". There are also this kind of error in other modules but are not reported.  🤔

elliot-sabitov:
> Lint work flow failed with errors in the unrelated files of this PR. How about we try to use `eslint $(git diff --name-only --diff-filter=d HEAD **/*.ts | xargs) ` to check only the changed files? (I search this command from internet, haven't tried) Another point looks confusing is that lint errors are only in the 01_Data module, such as "Strings must use singlequote". There are also this kind of error in other modules but are not reported. 🤔

@lydiazcheng the logs may not show full list of linting issues that were found. Most likely, to address the linting issues, we can use the `npm run lint-fix` command which should make the adjustments to fix all of the linting errors.

We can run the linter only on the files that were changed, but I figured that our goal is to make sure that everything is formatted properly, and every PR will need to ensure that all files are formatted so it seemed reasonable to lint the entire code base.

Please note that there is also `npm run prettier` command that was created. The vision here is that the devs would manually run `npm run lint-fix` and `npm run prettier` on their changes to make sure that we are in sync in terms of formatting. 

There can be ways to automate this so that they don't have to be triggered manually, but Thana, Christian and I agreed on this manual approach which I think is a good starting point

package.json
```diff
@@ -6,12 +6,17 @@
...
+    "typescript": "^5.3.3",
+    "prettier": "^3.2.5",
+    "eslint": "^8.48.0"
   },
   "scripts": {
     "install-all": "npm i --verbose",
     "clean": "rm -rf package-lock.json **/package-lock.json **/**/package-lock.json dist **/dist **/**/dist node_modules **/node_modules **/**/node_modules tsconfig.tsbuildinfo **/tsconfig.tsbuildinfo **/**/tsconfig.tsbuildinfo",
-    "build": "tsc --build --verbose"
+    "build": "tsc --build --verbose",
+    "lint": "npx eslint --ext .ts ./",
```
ChrisWeissmann, line 17: Linting doesn't seem to work for me. I get a `ESLint couldn't find the config "standard-with-typescript" to extend from. Please check that the name of the config is correct` It seems there is a difference in the `.eslint.json` between the directories. Do you have a best practice document somewhere that we can use and make the standard in citrine?

ChrisWeissmann (commented):
In the beginning I was also against the run configurations. But I think they could be a good idea to lower the friction of contribution. Probably should add something like this for vscode. They are very specific to use and offer a play button to build etc. They don't interfere with the general intelij settings. @thanaParis you still set on not adding RunConfigurations.

package.json
elliot-sabitov: @ChrisWeissmann you did `npm i` right? because I think in eslint we extend this standard from other npm packages

ChrisWeissmann: yea, I have run `npm i` The problem is in the `01_Data` 
> The config "standard-with-typescript" was referenced from the config file in "/Users/christian/Developer/citrineos-core/01_Data/.eslintrc.json".

In there we have `"extends": "standard-with-typescript",` while in the others we don't have that. 
There we have: `"extends": "eslint:recommended"` and in some places we have  `extends: ['eslint:recommended', 'plugin:@typescript-eslint/recommended'],`

Probably worth having one "right" config.

PR #70, opened 2024-04-05
Feature/module breakdown

elliot-sabitov:
feat: created transaction.event.service and adjusted transaction module to have all dependencies imported which required moving the sender/handler and logger into an injectable as we have previously done with config and cache, but was able to step through and it looks like it is injecting things properly

Screenshot 2024-04-05 at 3 35 07 PM

In the screenshot above you will see the module with all of the dependencies injected. Changes have been applied to all modules.

In addition, some of the dependencies that are being injected would change because right now, I am simply injecting the new `TransactionEventService` only to handle `_handleTransactionEvent` but other event handlers may have a mix of other services. I did not touch the other handlers yet.

Using this approach, things can me imported where needed so the modules can only import the services and services can import their appropriate repositories and feign clients etc... This will allow us to easily break down the handlers into multiple classes as creating classes will become significantly easier with being able to inject any dependency as arguments into the constructor as long as it is appropriate and makes sense and not like a circular dependency...


03_Modules/Transactions/src/module/module.ts
```diff
@@ -113,98 +109,7 @@ export class TransactionsModule extends BaseModule {
...
-          .then(messageConfirmation => this._logger.debug("Transaction response sent: ", messageConfirmation));
-      });
-    } else {
-      const response: TransactionEventResponse = {
-        // TODO determine how to set chargingPriority and updatedPersonalMessage for anonymous users
-      };
-      this.sendCallResultWithMessage(message, response)
-        .then(messageConfirmation => this._logger.debug("Transaction response sent: ", messageConfirmation));
-    }
+    await this.transactionEventService?.handleTransactionEvent(message, props);
```
elliot-sabitov, line 112: This is the main drive behind the changes. We have a module with multiple handlers, and here we can have a way to have multiple services handling the events which is greatly simplified with DI since things that are needed by the class can be imported into the constructor automatically

03_Modules/Transactions/src/module/services/transaction.event.service.ts
```diff
@@ -0,0 +1,183 @@
...
+import {
+  AuthorizationRepository,
+  TransactionEventRepository,
+} from "@citrineos/data";
+import { RabbitMqSender } from "@citrineos/util";
+
+@singleton()
+export class TransactionEventService {
+  constructor(
+    @inject(LoggerService) private readonly loggerService?: LoggerService,
```
elliot-sabitov, line 25: You can see that the new service automatically gets the needed logger, sender and repositories injected

01_Data/src/layers/sequelize/repository/Base.ts
```diff
@@ -3,91 +3,88 @@
...
-                model.setDataValue(k, value.getDataValue(k));
-            }
-            return model;
-        }
+import {ICrudRepository, injectable, LoggerService, SystemConfig, SystemConfigService} from '@citrineos/base';
+import { Model, Sequelize } from 'sequelize-typescript';
+import { DefaultSequelizeInstance } from '../util';
+import { inject } from 'tsyringe';
+
+@injectable()
```
elliot-sabitov, line 11: made @injectable and injecting the SystemConfig and Logger in constructor making it possible to make all repositories @injectable since they don't override the constructor

02_Util/src/index.ts
```diff
@@ -2,7 +2,7 @@
 // Copyright Contributors to the CitrineOS Project
 //
 // SPDX-License-Identifier: Apache 2.0
-
+import "reflect-metadata";
```
elliot-sabitov, line 5: this has to be early on in every package for tsyringe as well as our annotations and will probably be needed by other things so I added to all modules

Server/src/index.ts
```diff
@@ -76,346 +76,355 @@ export class CitrineOSServer {
...
+    //   this.getModuleConfig(EventGroup.Configuration),
+    //   this.getModuleConfig(EventGroup.EVDriver),
+    //   this.getModuleConfig(EventGroup.Monitoring),
+    //   this.getModuleConfig(EventGroup.Reporting),
+    //   this.getModuleConfig(EventGroup.SmartCharging),
+    //   this.getModuleConfig(EventGroup.Transactions),
+    // ].forEach((moduleConfig) => {
+      // this.initModule(moduleConfig);
+    // });
+    this.modules.push(this.transactionsModule as BaseModule);
```
elliot-sabitov, line 256: temporarily only initializing transactions module to because havent refactored the rest

Server/src/index.ts
```diff
@@ -76,346 +76,355 @@ export class CitrineOSServer {
...
-      const moduleConfig: ModuleConfig = this.getModuleConfig(this.eventGroup)
-      this.initModule(moduleConfig)
+      const moduleConfig: ModuleConfig = this.getModuleConfig(this.eventGroup);
+      // this.initModule(moduleConfig);
     }
   }
 
-  protected _createSender (): IMessageSender {
-    return new RabbitMqSender(this._config, this._logger)
+  protected _createSender(): IMessageSender {
```
elliot-sabitov, line 363: no longer need createSender/createHandler because now these get automatically injected

Server/src/index.ts
```diff
@@ -76,346 +76,355 @@ export class CitrineOSServer {
...
 
-  private addAjvFormats () {
+  private addAjvFormats() {
     addFormats(this._ajv, {
-      mode: 'fast',
-      formats: ['date-time']
-    })
-  }
-
-  private initLogger () {
```
elliot-sabitov, line 184: logger initialization happens in the LoggerService and is automatically injected

Server/src/index.ts
elliot-sabitov: we have flexibility using useFactory or just looking at the config in the provider to return depending on condition. In other words, we will be able to nicely switch between RabbitMq and Pubsub once we have each of them injectable first and then using tsyringe tools

elliot-sabitov:
I just updated the rest of the modules aside from OcppRouter to also use dependency injection - coming along smoothly, but the rest of the refactoring with breaking down the handlers into separate smaller pieces / services will have to happen gradually!

03_Modules/Transactions/src/module/services/transaction.event.service.ts
```diff
@@ -0,0 +1,194 @@
...
+        transactionEventResponse.idTokenInfo = idTokenInfo;
+      }
+    } else {
+      // IdTokenInfo.status is one of Blocked, Expired, Invalid, NoCredit
+      // N.B. Other non-Accepted statuses should not be allowed to be stored.
+      transactionEventResponse.idTokenInfo = idTokenInfo;
+    }
+  }
+
+  public sendCallResultWithMessage(
```
lydiazcheng, line 188: `sendCallResultWithMessage` has been defined in BaseModule.  Is there any better way to avoid this duplication?

03_Modules/Transactions/src/module/api.ts
```diff
@@ -46,12 +46,12 @@ export class TransactionsModuleApi extends AbstractModuleApi<TransactionsModule>
 
     @AsDataEndpoint(Namespace.TransactionEventRequest, HttpMethod.Get, TransactionEventQuerySchema)
     getTransactionEventsByStationIdAndTransactionId(request: FastifyRequest<{ Querystring: TransactionEventQuerystring }>): Promise<TransactionEventRequest[]> {
-        return this._module.transactionEventRepository.readAllByStationIdAndTransactionId(request.query.stationId, request.query.transactionId);
+        return this._module.transactionEventRepository?.readAllByStationIdAndTransactionId(request.query.stationId, request.query.transactionId)!;
```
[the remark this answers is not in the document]
elliot-sabitov: yes the ? and the ! are super useful and should be adopted as tools to help us with nullables. In this case, the repository is defined as `transactionEventRepository?: TransactionEventRepository` which enables DI to inject it, but the result is that TS is going to complain saying that `transactionEventRepository` may be `undefined` and so using `?` makes it run the code if it is not undefined. The `!` operator as I understand it is also useful in a similar fashion to get TS to behave assuming all went well as it should.

03_Modules/Transactions/src/module/services/transaction.event.service.ts
elliot-sabitov: this is a temporary workaround because in this case this is the `TransactionEventService` and it does not extend the BaseModule, while it is needed for the `sendTransactionEventConfirmation` method - but I can see a base service, or other ways of injecting the module to get this functionality across implemented in the near future for sure! I think with the introduction of dependency injection, we have a lot of flexibility in re-organizing how things are being intialized, and so this
```
return this.sender?.sendResponse(message, payload)!;
```
will potentially be moved elsewhere!

03_Modules/Transactions/src/module/api.ts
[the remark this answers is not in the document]
elliot-sabitov: in some cases, to get around this typescript `object may be undefined` I have also seen the `object as ObjectClass` syntax, but I have seen `!` used instead and it seems to work and is a bit cleaner

03_Modules/Transactions/src/module/services/transaction.event.service.ts
```diff
@@ -0,0 +1,530 @@
...
+   * @param {string} identifier - The identifier of the call.
+   * @param {string} tenantId - The tenant ID.
+   * @param {CallAction} action - The action to be performed.
+   * @param {OcppRequest} payload - The payload of the call.
+   * @param {string} [callbackUrl] - The callback URL for the call.
+   * @param {string} [correlationId] - The correlation ID of the call.
+   * @param {MessageOrigin} [origin] - The origin of the call.
+   * @return {Promise<IMessageConfirmation>} A promise that resolves to the message confirmation.
+   */
+  public sendCall(
```
elliot-sabitov, line 482: This comes from base, since this simply involves the cacheService and the sender (along with logger), I would say that it makes sense for this to move as we utilize DI more to adjust our structure to be service based. We have a lot of options of how this method can be shared, it can be in some Base class, it can be injected, it can be a static method on a class, but I think having this in Base makes sense since the modules had this in base

thanaParis:
Outdated.

PR #72, opened 2024-04-08
feat: adding jest and unit tests

elliot-sabitov:
feat: created few tests for transaction.event.mapper and transaction.event.service illustrating how things can be mocked
feat: adding jest.config to support Typescript in tests
chore: adding npm run test script in package.json and a run configuration to run tests


thanaParis:
Outdated.

PR #74, opened 2024-04-12
Certificate Part 2: Add Update (M)TLS Certificate Endpoint

lydiazcheng:
## Definitions
**Sub CA**: the certificate which is signed by root CA and is also the upper layer of CSMS cert and the charging station certificate. And it is sent together with charging station certificate as a chain in A02.
**Root CA**: the root cert from the external CA server, e.g., Let's encrypt.
**Certificate Chain**: a leaf cert following by sub CA cert but without root CA cert. When creating the http server, it needs a `cert` field configured with value in such a structure. In the OCPP 2.0.1, Part 2, the CertificateSignedRequest also need such a structured chain.
**Private key**: In this pr, there are 2 kinds of key. One is the key for CSMS cert and is used when create the http server to decrypt the content from the client side. Another kind of keys is the Sub CA key which signs the CSMS cert and the charging station certificate.

## Changes
* Add `IFileAcess` interface. 
* Extend directus util with Directus Files which allows get a file by file id
* Add a new endpoint which updates (m)tls related certs. This endpoint will update the tls cert, its key, as well as root CA in the http server without restarting the server. It also update the sub CA key which is used to sign the charging station certificate. 
    * require to give the server id as query param 
    * allows to update by given file id in request body

## Local Tests
1. set tls related configurations and the original certs are as in screenshot Screenshot 2024-05-15 at 9 37 17 AM
2.  call data endpoint to update certificates and keys Screenshot 2024-05-15 at 10 04 05 AM
3. Reload the web page, the certificates updated Screenshot 2024-05-15 at 9 53 23 AM
4. The update also replace files in asset folder. And there is a backup and rollback logic when update cert and key files. When there is an error in between, we can roll back to the original certs and keys.  Screenshot 2024-05-15 at 9 59 09 AM


thanaParis (approved):
Approved

PR #75, opened 2024-04-15
fix: issues with running docker compose when having node_modules and …

elliot-sabitov:
- fix: issues with running docker compose when having node_modules and dist files locally which previously messed up the docker compose volume mounting
- fix: adjusting docker compose to mount local package-lock.json file to an ignored file in docker container, and ensuring that other files are copied over while node_modules are still ignored. Using command to run npm i and rebuild for bcrypt and deasync then running start.
- fix: adjusted local.Dockerfile to ensure that WORKDIR is the root folder otherwise changing directories in docker compose command is difficult, so instead also creating npm run start:local-docker in root package.json.
- fix: ensuring nodemon properly ignores dist and node modules and doesnt needlessly trigger rebuilds
- fix: ensuring dist and node_modules are properly excluded in tsconfig
- fix: outDir path trailing slash not needed
- fix: ensuring package-lock.json is in gitignore so it does not get copied
- fix: swagger logoPath issues and missing variables in environment SystemConfigs
- chore: created npm run clean to remove build files
- chore: created npm run fresh to remove everything including node_modules, dist, package-lock, and tsconfig.buildinfo
- chore: excluding DirectusExtensions and Swarm behind a todo for now because Swarm causes build errors if dependencies changes but Swarm code is not updated when triggering npm run build


00_Base/package.json
```diff
@@ -34,7 +35,7 @@
     "class-transformer": "^0.5.1",
     "fastify": "^4.22.2",
     "reflect-metadata": "^0.1.13",
-    "tslog": "^4.9.2",
```
thanaParis, line 37: if this change is important enough to include, should we propagate it to the other package.json files?

elliot-sabitov: so this was one of the fixes we did last week for tslog and tsyringe... I just ended up copying and pasting the whole file and this was included... I can revert this one if you would like or propagate it to other packages... maybe we can just remove all carrots and tildas in one separate PR and I can just revert this here?

thanaParis: yes, please revert it so that it can have its own pr.

elliot-sabitov: reverted

thanaParis (approved):
Approved!

PR #77, opened 2024-04-18
Certificate Part 3: Add Generate Certificate Endpoint

lydiazcheng:
## Changes
This PR adds 2 new endpoints:
1. generate root certificate
   * either create a self signed root certificate and a sub CA
   * or only create a sub CA and sign it by an external ca root certificate (use acme client and Let'e encrypt root cert) 
2. install root certificate to charger.
   * if fileId is set in the request body, this endpoint will install the given file to charger
   * if not, the endpoint will get the external root CA according to the certificate type, e.g., if `certificateType` is `V2GRootCertificate`, it get root CA from hubject and install it.

## Local Test
**Test case 1:  create self signed root CA and sub CA**
1. call endpoint and get the generated certificate entities in the response body. Screenshot 2024-05-15 at 12 51 43 PM
2. generated certs and keys are stores in directus files Screenshot 2024-05-15 at 12 52 32 PM

**Test case 2: install root certificate to charger**
1. install self signed CSMSRootCertificate: Screenshot 2024-05-15 at 12 53 51 PM  Screenshot 2024-05-15 at 12 54 18 PM
2. install external CSMSRootCertificate from Lets encrypt  Screenshot 2024-05-15 at 12 55 57 PM Screenshot 2024-05-15 at 12 56 16 PM
4. install external V2GRootCertificate from hubject Screenshot 2024-05-15 at 12 57 25 PM Screenshot 2024-05-15 at 12 57 40 PM

**Test case 3: create sub CA signed by external CA server**
TBD

**Test case 4: Everest connect to Citrine using self signed root CA without no matched ciphers error**
TBD


03_Modules/Certificates/src/module/api.ts
```diff
@@ -228,6 +236,121 @@ export class CertificatesModuleApi
...
+      rootCAPem = externalRootCAPem;
+    }
+
+    // Send InstallCertificateRequest to the charger
+    const confirmation: IMessageConfirmation = await this.installCertificate(
+      certRequest.stationId,
+      certRequest.tenantId,
+      {
+        certificateType: certRequest.certificateType,
+        certificate: rootCAPem,
```
lydiazcheng, line 343: Here I change to send only the root CA which is different from what was decided in our previous discussion, i.e., send a sub CA and root CA chain in the installCertificate request. Since during testing with everest, I found they expect a single certificate. 
Screenshot 2024-05-07 at 5 34 34 PM
I think this also makes sense. Because in A02 update cert case, it expects for a certificate chain in the CertificateSignedRequest (leaf and the necessary sub CA certificates). And we might don't need to send the sub CA in install request.

thanaParis (approved):
Approved

PR #78, opened 2024-04-18
Feature/improved dockerfiles

thanaParis:
- .dockerignore now handles making sure unneeded files are not added to image
- dockerfile now uses node slim for final image
- server-citrine image size reduced from more than 2gb to ~550mb
- redundant build commands removed from npm scripts and nodemon.json
- DirectusExtension now has its own root files for tsconfig and docker build, so that it can use a different docker context from Server
- Some old unused files removed
- npm scripts for windows added, old scripts renamed to be unix-specific


lydiazcheng (commented):
The change looks good.

elliot-sabitov:
All changes look good. We just have to make sure that running docker compose after running node locally, doesnt result in any issues. Other than that, everything looks good

elliot-sabitov (approved):

PR #79, opened 2024-04-22
Certificate Part 1: Integrate CA Server Client

lydiazcheng:
## Changes
* Add CA client interfaces
  * `IV2GCertificateAuthorityClient`
  * `IChargingStationCertificateAuthorityClient`
* Add 2 Implementations of ca clients: Hubject, ACME (Let's encrypt)
  * Hubject: process V2G cert in A02
  * ACME: process Charging Station cert in A02. Also allow sign cert using Let's encrypt root ca cert. This function is used in PR: https://github.com/citrineos/citrineos-core/pull/77
* Add new fields for ca clients in system config. 
* Add a `CertificateAuthorityService` which instantiates CA clients.
* refactor `_handleSignCertificate`.
  * verify request and then accept or reject the request. (Currently this logic is commented out partially due to a OCTT failed test case)
  * sign csr and return a certificate chain (from leaf to sub ca cert except for root ca cert) to charger

## Local Tests
**Test case 1**  sign v2g certificate by hubject and return a cert chain contain signed cert followed by 2 sub CAs. Screenshot 2024-05-02 at 9 14 40 AM
**Test case 2**  sign charging station certificate by our sub CA and return a cert chain contain signed cert followed by the sub CA. (Change some code temperaly to only test the flow in A02 without testing a fully workable Profile 3 connection)  Screenshot 2024-05-02 at 9 13 44 AM
**Test case 3** verify organization name in the csr Screenshot 2024-05-02 at 10 07 00 AM


03_Modules/Certificates/src/module/service/CertificateAuthority.ts
```diff
@@ -0,0 +1,102 @@
...
+  private readonly _logger: Logger<ILogObj>;
+
+  constructor(config: SystemConfig, cache: ICache, logger?: Logger<ILogObj>) {
+    this._logger = logger
+      ? logger.getSubLogger({ name: this.constructor.name })
+      : new Logger<ILogObj>({ name: this.constructor.name });
+
+    this._acmeClient = new Acme(config, this._logger);
+    this._hubjectClient = new Hubject(config);
+    this._localCertificateAuthority = new Local(config, cache, this._logger);
```
thanaParis, line 30: I'm a little confused by this section.
If I'm following, we have a 'hubjectClient' which is for V2G Certs, and then 'localCA' or 'acmeClient' for ChargingStation Certs
but here you instantiate both acmeClient and localCA. From what i can see here, acmeClient is effectively unused.

Would the way this works be: 2 different kinds of CA clients (V2G & ChargingStation) with 1 implementation per kind depending on the SystemConfig (so, _either_ acme _or_ local, not both)?

lydiazcheng: > I'm a little confused by this section. If I'm following, we have a 'hubjectClient' which is for V2G Certs, and then 'localCA' or 'acmeClient' for ChargingStation Certs but here you instantiate both acmeClient and localCA. From what i can see here, acmeClient is effectively unused.
> 
> Would the way this works be: 2 different kinds of CA clients (V2G & ChargingStation) with 1 implementation per kind depending on the SystemConfig (so, _either_ acme _or_ local, not both)?

Hi Thana, sorry for this mess. 

The local client is used for signing charging station certificate. It uses our mtls private key (sub CA key) to sign the cert.

The acme client is used for the PR 3 which adds a generate cert endpoint. In that PR 3, the acme client is used to generate the Sub CA certificate signed by Let's encrypt. I initiate it in the CertificateAuthority as well so that api or module script can always only interact with CertificateAuthority rather than multiple clients.

Regarding the last question, for charging station certificate, we can only use local client because we need to use our own Sub CA certificate to sign it due to the common name limitation (unique serial number of charger). For V2G certificate, I am not sure whether it can be signed by acme client(Let's encrypt) as well. Or only hubject can sign and return the correct V2G root cert?

lydiazcheng: Besides, I am thinking about removing the local client and move its functions into the `CertificateAuthority`. Because it is not a real client like hubject and acme which stands for an external service. (I though it might be better organized when I initially added this local client.) And during working for the PR 2 (update tls certs endpoint), I also found the local client brings additional difficulty when updating the content inside the `_securityCaCertsKeyMap`.

03_Modules/Certificates/src/module/client/acme.ts
```diff
@@ -0,0 +1,223 @@
...
+    });
+
+    this._email = config.modules.certificates?.chargingStationCA.acme?.email;
+    const accountKey = fs.readFileSync(
+      config.modules.certificates?.chargingStationCA?.acme
+        ?.accountKeyFilePath as string,
+    );
+    const acmeEnv: string | undefined =
+      config.modules.certificates?.chargingStationCA?.acme?.env;
+    if (!acmeEnv && acmeEnv === 'production') {
```
yuvenus, line 61: can this if statement be simplified to `acmeEnv === 'production'` ? just wondering if this condition works with that first condition `!acmeEnv` 🤔

03_Modules/Certificates/src/module/module.ts
```diff
@@ -193,94 +166,52 @@ export class CertificatesModule extends AbstractModule {
...
-    let caCert: forge.pki.Certificate;
-    let caPrivateKey: forge.pki.rsa.PrivateKey;
-    let csr;
+
+    // Do Verification and send accept response before signing.
+    // Reject the request if the verification fails
     try {
-      csr = forge.pki.certificationRequestFromPem(message.payload.csr);
+      await this._verifySignCertRequest(csrString, stationId, certificateType);
+      // this.sendCallResultWithMessage(message, {
```
yuvenus, line 188: is this commented out temporarily? wondering if an additional TODO comment to be like "will be uncommented in another PR" is needed so that it's not confusing why there's commented out code 😅

03_Modules/Certificates/src/module/client/acme.ts
lydiazcheng: Thank you, Venus! You are right. I removed 1st condition.

thanaParis (approved):
Approved

PR #81, opened 2024-04-28
Remediate server-directus healthcheck

fabioluiz1:
See https://github.com/citrineos/citrineos/issues/20#issuecomment-2081319273

### DIFF
--- a/Server/docker-compose.yml
+++ b/Server/docker-compose.yml
@@ -136,7 +136,7 @@ services:
       DB_PASSWORD: 'citrine'
       WEBSOCKETS_ENABLED: 'true'
     healthcheck:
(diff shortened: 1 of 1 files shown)


volkert-fastned:
Thanks for finding the cause of the failed container and for opening a PR so quickly!

I'm. curious, though: why doesn't `localhost` work here? Shouldn't `localhost` always resolve to `127.0.0.1`?

thanaParis (approved):
Approved!

thanaParis:
> I'm curious, though: why doesn't localhost work here? Shouldn't localhost always resolve to 127.0.0.1?
 
[From the log in this comment](https://github.com/citrineos/citrineos/issues/20#issuecomment-2081319273)
It seems docker binds localhost to the IPv6 ::1, so that breaks whereas the explicit call to 127.0.01 works.

PR #82, opened 2024-04-29
WIP Feature/ocpi

elliot-sabitov:
Creating early for expusure


00_Base/src/interfaces/api/AbstractModuleApi.ts
```diff
@@ -199,13 +231,29 @@ export abstract class AbstractModuleApi<T extends IModule>
...
+    if (
+      !!schema &&
+      !!schema.headers &&
+      !!schema.headers.properties &&
+      !!schema.headers.properties.Authorization
+    ) {
+      _opts['preHandler'] = (this._server as unknown as any).auth([
+        (this._server as unknown as any).authorization,
+      ]);
+      if (!_opts['security']) {
```
elliot-sabitov, line 250: currently, nothing is using security, and when reviewing these few lines, we may not want to add the same AuthorazationSecurity in all cases where Authorization header is present, which can result in a bug. This code was created to help clunk the OCPI endpoints in one set and apply security across all of them. I am going to move this to as.ocpi.endpoint.ts class instead.

02_Util/src/util/swagger.ts
```diff
@@ -36,11 +46,13 @@ function OcppTransformObject({
             methodKey
           ] as OpenAPIV3.OperationObject;
           if (method) {
-            // Set tags based on path key
-            method.tags = pathKey
-              .split('/')
-              .slice(2, -1)
-              .map((tag) => tag.charAt(0).toUpperCase() + tag.slice(1));
+            // Set tags based on path key if tags were not passed in
+            if (!method.tags) {
```
elliot-sabitov, line 50: adjusting to instead have current logic as fallback and if tags are passed in then they are honored

02_Util/src/util/swagger.ts
```diff
@@ -76,23 +88,19 @@ function OcppTransformObject({
...
-        version: '1.1.1',
-      },
-    },
-    transformObject: OcppTransformObject,
-  });
-
+) => {
   const swaggerUiOptions: any = {
     routePrefix: systemConfig.util.swagger?.path,
+    securityDefinitions: {
```
elliot-sabitov, line 97: need to create authorization swaggerDefinition, but we may have multiple so we may need to move this to OCPI module right @thanaParis ? This is what I was saying is a bit convoluted, because we need the OCPI endpoints to be able to have this authorization securityDefinition

02_Util/src/util/swagger.ts
```diff
@@ -117,4 +125,126 @@ export function initSwagger(
...
+    throw new UnauthorizedException('Token not found in headers');
+  }
+  return token;
+};
+
+const registerFastifyAuth = async (server: FastifyInstance) => {
+  await server.register(FastifyAuth as any).after();
+  console.log((server as any).authorization);
+
+  server.decorate(
```
elliot-sabitov, line 187: This decorator callback function will run and authorize the request. This is the way for fastify to have authorization headers available for parsing. Currently left token hardcoded, but will look to implement logic that checks token against credentials module

02_Util/src/util/swagger.ts
```diff
@@ -117,4 +125,126 @@ export function initSwagger(
...
+  server: FastifyInstance,
+) => {
+  server.register(fastifySwagger, {
+    openapi: {
+      info: {
+        title: 'CitrineOS Central System API',
+        description: 'Central System API for OCPP 2.0.1 messaging.',
+        version: '1.1.1',
+      },
+      servers: [
```
elliot-sabitov, line 218: needed to add `servers` so that we can hit the endpoints via Swagger UI

02_Util/src/util/swagger.ts
```diff
@@ -117,4 +125,126 @@ export function initSwagger(
...
+        {
+          url: `http://${systemConfig.centralSystem.host}:${systemConfig.centralSystem.port}`,
+          description: 'TODO',
```
elliot-sabitov, line 221: left a todo as we may look into adding multiple environments and descriptions

02_Util/src/util/swagger.ts
```diff
@@ -117,4 +125,126 @@ export function initSwagger(
...
+      ],
+      components: {
+        securitySchemes: {
+          authorization: {
+            type: 'http',
+            scheme: 'bearer',
+          },
+        },
+      },
+      tags: Object.values(OcpiTag).map((tag) => {
```
elliot-sabitov, line 232: currently including all OCPI tags again in the same fastify instance - we may need better separation here

02_Util/src/util/swagger.ts
```diff
@@ -117,4 +125,126 @@ export function initSwagger(
...
+  });
+};
+
+export async function initSwagger(
+  systemConfig: SystemConfig,
+  server: FastifyInstance,
+) {
+  registerFastifySwagger(systemConfig, server);
+  registerSwaggerUi(systemConfig, server);
+  await registerFastifyAuth(server);
```
elliot-sabitov, line 249: adjusted previous implementation to handle the
```
await server.register(FastifyAuth as any).after();
```
await because otherwise the `authorization` and `.auth` are not available when creating the endpoints in `AbstractModuleApi`

03_Modules/Ocpi/src/apis/BaseApi.ts
```diff
@@ -0,0 +1,362 @@
+import { OcpiResponse } from '@citrineos/base';
```
elliot-sabitov, line 1: generated base file for ocpi triggers (receiver endpoints)

03_Modules/Ocpi/src/apis/ChargingProfilesControllerApi.ts
```diff
@@ -0,0 +1,120 @@
...
+
+    if (requestParameters.responseUrl != null) {
+      queryParameters['response_url'] = requestParameters.responseUrl;
+    }
+
+    const headerParameters: HTTPHeaders =
+      validateAndgetOcpiHeaders(requestParameters);
+
+    setAuthHeader(headerParameters);
+    return await this.request({
```
elliot-sabitov, line 49: created all Api classes to follow similar format to perform the `receiver` requests utilizing Base and helpers

03_Modules/Ocpi/src/modules/temp/credentials.api.ts
```diff
@@ -0,0 +1,139 @@
...
+      finalVersionRepository,
+    );
+  }
+
+  @AsDataEndpoint(
+    '/ocpi/:versionId/credentials',
+    HttpMethod.Get,
+    undefined,
+    undefined,
+    targetConstructorToSchema(VersionIdParam),
```
elliot-sabitov, line 53: `targetConstructorToSchema` is used in @AsOcpiEndpoint as well but here, we need to do it manually as to not to infringe on existing functionality, although there may be a better way to make sure that the schema is generated closer to fastify. Additionally, with fastify not having a very robust setup with `class-validator` which is being used to generate the schema via the annotations - I have seen some of the annotations like `@Equals(VersionNumber.TWO_DOT_TWO_DOT_ONE)` fail in this setup because class validator isnt actually getting a chance to run because we are only using it to generate the schema via `targetConstructorToSchema` - @thanaParis I would love to use these annotations more extensively, as I think it would add value to our code base - what are your thoughts here, this does work as is - but can be improved...

03_Modules/Ocpi/src/modules/temp/exceptions/global.exception.handler.ts
```diff
@@ -0,0 +1,27 @@
+import { FastifyError, FastifyReply, FastifyRequest } from 'fastify';
+import { ILogObj, Logger } from 'tslog';
+import {
+  ExceptionHandler,
+  HttpStatus,
+  UnauthorizedException,
+} from '@citrineos/base';
+import { NotFoundException } from './not.found.exception';
+
+export class GlobalExceptionHandler implements ExceptionHandler {
```
elliot-sabitov, line 10: Setting a `GlobalExceptionHandler` via `setErrorHandler` in `EverythingElseApi` for now, but should be repurposed to have a more robust error handling setup

03_Modules/Ocpi/src/modules/temp/exceptions/not.found.exception.ts
```diff
@@ -0,0 +1,6 @@
+export class NotFoundException extends Error {
```
elliot-sabitov, line 1: added some exceptions, and the `UnauthorizedException` had to be added to `Base` - the exceptions are handled by the `GlobalExceptionHandler` to return the appropriate response

03_Modules/Ocpi/src/modules/temp/module.ts
```diff
@@ -0,0 +1,57 @@
...
+  IMessageSender,
+  SystemConfig,
+} from '@citrineos/base';
+import {RabbitMqReceiver, RabbitMqSender} from '@citrineos/util';
+import {ILogObj, Logger} from 'tslog';
+
+/**
+ * Component that handles transaction related messages.
+ */
+export class OcpiModule extends AbstractModule {
```
elliot-sabitov, line 21: placeholder module

03_Modules/Ocpi/src/modules/temp/service/credentials.service.ts
```diff
@@ -0,0 +1,210 @@
...
+
+export class CredentialsService {
+  constructor(
+    private _logger: Logger<ILogObj>,
+    private credentialsRepository: CredentialsRepository,
+    private versionRepository: VersionRepository,
+  ) {
+  }
+
+  private tempCreateCredentials() {
```
elliot-sabitov, line 31: will remove this method which was only used temporarily

03_Modules/Ocpi/src/modules/temp/service/credentials.service.ts
```diff
@@ -0,0 +1,210 @@
...
+    const version = versions.data?.find((v) => v.version === versionId);
+    if (!version) {
+      throw new Error('todo'); // todo error handling
+    }
+    const versionDetails = await versionsControllerApi.getVersion({
+      authorization: token,
+      versionId: versionId,
+    });
+    if (!versionDetails) {
+      throw new Error('todo'); // todo error handling
```
elliot-sabitov, line 191: will fix error handling now that it is set up

PR #84, opened 2024-05-02
Fix undefind status error

lydiazcheng:
When using `VariableAttributeQuerystring` to query any entities in DeviceModel without set `status` value, it returns error as in screenshot. This PR is to fix this error.
Screenshot 2024-05-02 at 9 08 13 AM

Local Test
Screenshot 2024-05-02 at 10 51 05 AM

### DIFF
--- a/01_Data/src/layers/sequelize/repository/DeviceModel.ts
+++ b/01_Data/src/layers/sequelize/repository/DeviceModel.ts
@@ -290,9 +290,9 @@ export class DeviceModelRepository extends SequelizeRepository<VariableAttribute
     return {
       where: {
         ...(queryParams.stationId ? { stationId: queryParams.stationId } : {}),
-        ...(queryParams.type !== null ? { type: queryParams.type } : {}),
+        ...(queryParams.type ? { type: queryParams.type } : {}),
         ...(queryParams.value ? { value: queryParams.value } : {}),
-        ...(queryParams.status !== null ? { status: queryParams.status } : {}),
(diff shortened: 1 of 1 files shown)


01_Data/src/layers/sequelize/repository/DeviceModel.ts
```diff
@@ -290,9 +290,9 @@ export class DeviceModelRepository extends SequelizeRepository<VariableAttribute
     return {
       where: {
         ...(queryParams.stationId ? { stationId: queryParams.stationId } : {}),
-        ...(queryParams.type !== null ? { type: queryParams.type } : {}),
```
thanaParis, line 293: The original intent of these checks was to permit explicit searching for cases where the type or status was null. I.e., _undefined_ type (missing from query) would be interpreted as any type, but null type (present in query as null) would be interpreted as no type.

This functionality doesn't look like it works as written either. Is there a way we can fix it that provides the intended functionality?

lydiazcheng: Thanks for the feedback, Thana! I updated the code. Here are some tests. 
1. filter type by null, only one attribute has null type returns:  Screenshot 2024-05-02 at 1 43 37 PM
2. undefined type in the query then all attributes return Screenshot 2024-05-02 at 1 44 51 PM
3. The message endpoint works  Screenshot 2024-05-02 at 1 55 10 PM

thanaParis (commented):
Approved

thanaParis (approved):

PR #85, opened 2024-05-07
Accept "NoAuthorization" Authorize Requests

langiole:
Authorize requests that specify a type of "NoAuthorization" will now result in a "Accepted" response:

Screenshot 2024-05-07 at 12 33 37


thanaParis (approved):
~~Approved!~~ see below

03_Modules/EVDriver/src/module/module.ts
```diff
@@ -167,6 +167,12 @@ export class EVDriverModule extends AbstractModule {
             // TODO determine how/if to set personalMessage
           },
         };
+
+        if (message.payload.idToken.type === IdTokenEnumType.NoAuthorization) {
```
thanaParis, line 171: IdTokenEnumType is not imported according to the build result. Please add it to the import statements.

thanaParis (changes requested):
Changes needed for broken build.

thanaParis (approved):
Approved!

PR #87, opened 2024-05-08
Feature/crud events

thanaParis:
Adding crud events to CrudRepository.

- ICrudRepository > CrudRepository, interface to abstract class in order to extend EventEmitter to emit lifecycle events
- 'created', 'updated', 'deleted' events
- Updated repository classes to emit events
- Changed Sequelize repository implementation names to include Sequelize
- SequelizeRepository from abstract class to class


lydiazcheng (commented):
Hi Thana, I am reading this change and have a question, could you please share the context of this change? Is it an improvement or a step 0 for a new feature?

thanaParis:
> Hi Thana, I am reading this change and have a question, could you please share the context of this change? Is it an improvement or a step 0 for a new feature?

Both.

> ICrudRepository > CrudRepository, interface to abstract class in order to extend EventEmitter to emit lifecycle events
> 'created', 'updated', 'deleted' events
> Updated repository classes to emit events

These new lifecycle events are valuable improvements. We are adding them now in order to support OCPI data pushes in the near future.

00_Base/src/interfaces/repository.ts
```diff
@@ -3,21 +3,66 @@
 //
 // SPDX-License-Identifier: Apache 2.0
 
+import { EventEmitter } from 'events';
+
+export type CrudEvent<T> = {
```
elliot-sabitov, line 8: should CrudEvent be a generic class?

00_Base/src/interfaces/repository.ts
```diff
@@ -3,21 +3,66 @@
...
+  deleted: [T[]];
+};
+
 /**
  * Represents a generic CRUD repository.
  *
  * @template T - The type of the values stored in the repository.
  */
-export interface ICrudRepository<T> {
+export abstract class CrudRepository<T> extends EventEmitter {
```
elliot-sabitov, line 19: what made you choose to extend EventEmitter instead of just having an event emitter on the repository class as a property? My thinking is you can just have something like 
```
export abstract class CrudRepository<T> extends EventEmitter {
   ...
   eventEmitter = new EventEmitter();
   ...
   public async create(value: T, namespace?: string): Promise<T> {
    const result = await this._create(value, namespace);
    this.eventEmitter.emit('created', [result]);
    return result;
  }
}
```
and I think then you dont need these:
```
on<K extends keyof CrudEvent<T>>(
    event: K,
    listener: (...args: CrudEvent<T>[K]) => void,
  ): this {
    return super.on(event, listener as (...args: any[]) => void);
  }
  
  emit<K extends keyof CrudEvent<T>>(
    event: K,
    ...args: CrudEvent<T>[K]
  ): boolean {
    return super.emit(event, ...args);
  }
  ```

File: 00_Base/src/interfaces/repository.ts [hunk 3]
```diff
@@ -61,46 +134,108 @@ export interface ICrudRepository<T> {
...
-  updateByQuery(
-    value: T,
+  public async updateAllByQuery(
+    value: Partial<T>,
     query: object,
     namespace?: string,
-  ): Promise<T | undefined>;
+  ): Promise<T[]> {
+    const result = await this._updateAllByQuery(value, query, namespace);
+    this.emit('updated', result);
```
elliot-sabitov, line 167: should we also emit the Class as well as the object or some sort of type along or do you think the object itself will be sufficient? My thought process here is looking at some utility behind redux and when the event is emitted with just the object, it means that you need to have a good parser to have multiple handlers for the different classes that are being emitted. I think you can pass the Class itself along as well (similar to how we do it in java like ClassName.class but with different syntax) - and I think if you pass class along then you can just have a simple switch down the line to parse the emitted event and be able to quickly pass the payload to the correct handler.

It is kind of hard to explain, so I will be happy to clarify!

01_Data/src/layers/sequelize/repository/Authorization.ts
```diff
@@ -26,10 +47,12 @@ export class AuthorizationRepository extends SequelizeRepository<Authorization>
         ...value.idTokenInfo,
       });
       if (authorizationModel.idTokenInfoId) {
-        let savedIdTokenInfo = (await this.s.models[IdTokenInfo.MODEL_NAME].findOne({
-          where: { id: authorizationModel.idTokenInfoId },
-          include: [{ model: IdToken, include: [AdditionalInfo] }],
-        })) as IdTokenInfo;
+        let savedIdTokenInfo = await this.idTokenInfo
+          .readAllByQuery({
```
elliot-sabitov, line 51: since we are doing:
```
.then((rows) => rows[0]);
```
should we have an interface for `this.idTokenInfo.readfirstByQuery`?

01_Data/src/layers/sequelize/repository/DeviceModel.ts
```diff
@@ -227,29 +275,29 @@ export class DeviceModelRepository extends SequelizeRepository<VariableAttribute
...
+    return await super.deleteAllByQuery(this.constructQuery(query));
   }
 
   async findComponentAndVariable(componentType: ComponentType, variableType: VariableType): Promise<[Component | null, Variable | null]> {
-    const component = await Component.findOne({
+    const component = await this.component.readAllByQuery({
       where: { name: componentType.name, instance: componentType.instance ? componentType.instance : null },
-    });
-    const variable = await Variable.findOne({
+    }).then((components) => components.length > 0 ? components[0] : null);
```
elliot-sabitov, line 292: similar situation here where maybe we can benefit from having `findFirst` helper

elliot-sabitov: to eiliminate / centralize occurenses of logic like this 
```.then((variableCharacteristics) => variableCharacteristics.length > 0 ? variableCharacteristics[0] : undefined);
```
and have it in one place in the helpers / base

01_Data/src/layers/sequelize/repository/Subscription.ts
```diff
@@ -3,10 +3,18 @@
...
 
+import { Sequelize } from 'sequelize-typescript';
 import { SequelizeRepository, Subscription } from '..';
 import { ISubscriptionRepository } from '../../..';
+import { SystemConfig } from '@citrineos/base';
+import { Logger, ILogObj } from 'tslog';
+
+export class SequelizeSubscriptionRepository extends SequelizeRepository<Subscription> implements ISubscriptionRepository {
+
+  constructor(config: SystemConfig, logger?: Logger<ILogObj>, namespace = Subscription.MODEL_NAME, sequelizeInstance?: Sequelize) {
```
elliot-sabitov, line 14: instead of passing namespace around constructors cant we just set it on the class like so:
```
export class SequelizeSubscriptionRepository extends SequelizeRepository<Subscription> implements ISubscriptionRepository {
   namespace = Subscription.MODEL_NAME
...
}
```
and I think when the base will use `this.namespace` it should pick up the right value right? This way we don't have to pass this around and can just define it on the repo as a property which may be a little cleaner - thoughts?

elliot-sabitov:
Overall looks good - left a few comments on the code for potentially improvements, let me know if you wish to discuss any in more detail and feel free to ignore / proceed since all the comments were points for consideration and nothing critical from my end

00_Base/src/interfaces/repository.ts
elliot-sabitov: this way we can have builder(s), helper methods, (de)serializer(s), annotation(s) etc... and should be more stable I think in general with classes since these represent data objects

01_Data/src/layers/sequelize/repository/Boot.ts
```diff
@@ -3,74 +3,56 @@
...
-        return await boot.save();
-      }
-    });
+    return await this.updateByKey({ lastBootTime }, key);
   }
 
-  async readByKey(key: string): Promise<Boot> {
-    return await super.readByKey(key, Boot.MODEL_NAME);
+  async readByKey(key: string): Promise<Boot | undefined> {
+    return await this.readByKey(key);
```
lydiazcheng, line 47: Will this keep calling itself?

lydiazcheng (commented):
The changes looks good and helpful! Only some questions and minor point.

01_Data/src/layers/sequelize/repository/Boot.ts
elliot-sabitov: I am a bit confused here as well 🤔

00_Base/src/interfaces/repository.ts
thanaParis: CrudEvent is used exclusively to provide typing for the supported events.
In a code block such as below, your IDE will only permit the events in CrudEvent to be used as values for eventName, and it will automatically provide the typing for the callback argument.

```
const variableStatus = new SequelizeRepository<VariableStatus>(config, VariableStatus.MODEL_NAME, logger, sequelizeInstance);
const eventName = 'created';
variableStatus.on(eventName, async (variableStatuses) => { 
    // typeof  variableStatuses = VariableStatus[]
});
```

I.E. if you changed eventName to 'create' it would not compile, and if you are preventing implicit 'any' typing this will still lint okay since the callback argument (variableStatuses) is explicitly typed.
This works because it is generic, since it can utilize the same generic typing as CrudRepository implementations when used in the function signatures of on() and emit().

00_Base/src/interfaces/repository.ts
thanaParis: That would no longer supply typing for the emitted events, or require the use of CrudEvent keys. This would make it harder to develop with and less safe simultaneously. [See this comment for more detail.](https://github.com/citrineos/citrineos-core/pull/87#discussion_r1602183957)

00_Base/src/interfaces/repository.ts
thanaParis: There's no worry of any of this. The typing is supplied from the function signature.

01_Data/src/layers/sequelize/repository/Authorization.ts
thanaParis: `this.idTokenInfo.readfirstByQuery` could be a very dangerous function to exist, as it might promote usage that should break (i.e. if there should only be one variable attribute that matches and two are found, an error should be thrown).

I put a comment on line 55 explicitly to make it clear why it was safe, in this scenario, to assume one array element. It seems it wasn't clear enough; I will update this code so it breaks if there's more than one element returned.

`readOnlyByQuery`, a version of `readfirstByQuery` that throws an error if more than one entry matches the query, could be a useful function to exist. Maybe a future refactor could implement it if it seems it would simplify the code.

01_Data/src/layers/sequelize/repository/Authorization.ts
```diff
@@ -140,16 +152,20 @@ export class AuthorizationRepository extends SequelizeRepository<Authorization>
...
-      savedIdTokenModel = (await this.s.models[IdToken.MODEL_NAME].findOne({
-        where: { id: savedIdTokenId },
-        include: [AdditionalInfo],
-      })) as IdToken;
+      savedIdTokenModel = await this.idToken
+        .readAllByQuery({
+          where: { id: savedIdTokenId },
+          include: [AdditionalInfo],
+        })
+        .then((rows) => rows[0]);
```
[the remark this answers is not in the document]
thanaParis: In order for this to happen, there would need to be a bizarre bug or race condition in the SQL database since savedIdTokenId is a foreign key. I'll add a check regardless.

01_Data/src/layers/sequelize/repository/Authorization.ts
```diff
@@ -3,19 +3,40 @@
...
+
+export class SequelizeAuthorizationRepository extends SequelizeRepository<Authorization> implements IAuthorizationRepository {
+  idToken: CrudRepository<IdToken>;
+  idTokenInfo: CrudRepository<IdTokenInfo>;
+  additionalInfo: CrudRepository<AdditionalInfo>;
+
+  constructor(
+    config: SystemConfig,
+    logger?: Logger<ILogObj>,
+    namespace = Authorization.MODEL_NAME,
```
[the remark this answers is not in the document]
thanaParis: Excellent point. Will change.

01_Data/src/layers/sequelize/repository/Boot.ts
```diff
@@ -3,74 +3,56 @@
...
 import { VariableAttribute } from '../model/DeviceModel';
-import { Op } from 'sequelize';
+import { SequelizeRepository } from '..';
+import { Logger, ILogObj } from 'tslog';
+import { Sequelize } from 'sequelize-typescript';
+
+export class SequelizeBootRepository extends SequelizeRepository<Boot> implements IBootRepository {
+  variableAttributes: CrudRepository<VariableAttribute>;
+
+  constructor(config: SystemConfig, logger?: Logger<ILogObj>, namespace = Boot.MODEL_NAME, sequelizeInstance?: Sequelize, variableAttributes?: CrudRepository<VariableAttribute>) {
```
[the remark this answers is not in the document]
thanaParis: Excellent point. Will change.

01_Data/src/layers/sequelize/repository/Boot.ts
thanaParis: whoops. my bad, will fix!

01_Data/src/layers/sequelize/repository/Subscription.ts
thanaParis: I changed it to be assigned in the super constructor as Zihe mentioned; I think I like that better than having it be a field though they both are better 👍

01_Data/src/layers/sequelize/repository/DeviceModel.ts
thanaParis: I agree! Thanks for the suggestion; I've added a 'readOnlyOneByQuery' method which throws an error if more than one result is returned, and replaced situations like the above with a call to it.

PR #88, opened 2024-05-09
Certificate Part 4: Handle Get15118EVCertificateRequest

lydiazcheng:
### Changes
The PR is to implement the Get15118EVCertificateRequest handler which is based on description of use cases, M01 and M02 in OCPP 2.0.1 Part 2.
It uses hubject to get the 15118 EV Certificate.

### Tests
1. Test parse the successful response body and get the exiResponse. (Since hubject does not provide a request example that can get a 200 response in its staging env, I did not manage to test the whole flow but manually inject the success response in the code to test the parsing.)  Screenshot 2024-05-10 at 9 46 44 AM
2. Test get cert failed by giving the invalid schema version and hubject return 400 response. Screenshot 2024-05-10 at 9 53 13 AM


thanaParis (approved):
Approved!!

PR #89, opened 2024-05-10
Fix: Avoid to overwrite mutability and other fields in stored VariableAttribute when call set variable

lydiazcheng:
### Goal
This PR is to fix an issue with SetVariables overwriting data that it does not contain (i.e. mutability is reset).
The cause of the issue is that, when storing VariableAttribute in db, we always set default values for some fields, e.g., mutability, when they are omitted in the given object and then store these default values in db. So the stored values are overwrite by default values. This way of setting default values works for use case handle NotifyReport req since OCPP 2.0.1 expect us to do so. But we need a fix for use case SetVariable where default values should not overwrite stored values in db.

### Changes
1. In `createOrUpdateDeviceModelByStationId`, replace upsert with create and update so that default values are used when creation. When updating, it depends on values in the given entity. We don't set any default values when any fields omitted.
2. In usages of `createOrUpdateDeviceModelByStationId` we make sure to set the correct values in the entity and give it to `createOrUpdateDeviceModelByStationId`.
3. In this way, db layer don't need to know which use cases (NotifyReport or SetVariable) it is or whether use default values. 

### Tests
1. Prepare: Manually change values of a VariableAttribute in db Screenshot 2024-05-10 at 3 49 35 PM
2. Call SetVariable from postman Screenshot 2024-05-10 at 3 50 30 PM
3. Other fields are not overwritten except for `value` Screenshot 2024-05-10 at 3 52 30 PM
4. Send a notifyReport req and omit `mutability`, `consistent` and `persistent` fields. Screenshot 2024-05-10 at 3 54 24 PM
5. The saved VariableAttribute are overwritten with default values.  Screenshot 2024-05-10 at 3 56 56 PM

### DIFF
--- a/01_Data/src/layers/sequelize/repository/DeviceModel.ts
+++ b/01_Data/src/layers/sequelize/repository/DeviceModel.ts
@@ -35,20 +35,32 @@ export class DeviceModelRepository extends SequelizeRepository<VariableAttribute
 
     return await Promise.all(
       value.variableAttribute.map(async (variableAttribute) => {
-        // Even though defaults are set on the VariableAttribute model, those only apply when creating an object
-        // So we need to set them here to ensure they are set correctly when updating
-        const [savedVariableAttribute, _variableAttributeCreated] = await VariableAttribute.upsert({
-          stationId,
-          variableId: variable.id,
-          componentId: component.id,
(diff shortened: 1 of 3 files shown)


thanaParis (approved):
Approved!

PR #92, opened 2024-05-14
Remove Redis from docker-compose.yml as default cache is the In-Memory cache

yuvenus:
- Removed the Redis instance from docker-compose.yml to simplify it, as the In-Memory cache is the default cache
- Updated a small typo in the README regarding the `01_Data` folder

Merges into rc-1.2.0 but can retarget as needed. Thank you!


lydiazcheng (approved):

thanaParis (approved):
Approved

PR #95, opened 2024-05-17
Adjust parsing of Basic authorization credentials to handle colons in passwords

falanadamian:
This PR resolves inaccuracies in Basic authorization credentials extraction when passwords contain colons. 
Currently, credentials are split at every colon, resulting in inaccurate password extraction. For example, "username:pass:word" yields "pass" as the password instead of "pass:word".


ChrisWeissmann (approved):
Looks good to me 🚀

PR #98, opened 2024-05-21
fix: don't use env variables

ChrisWeissmann:
I in my last commit I moved to the matrix strategy and didn't notice you weren't allowed to use env variable.

fix: matrix strategy doesn't have access to env vars so need to use literal string


thanaParis (approved):
Approved

PR #103, opened 2024-05-22
(1.3.0) Tenant Module

yuvenus:
**Will re-target to rc-1.3.0 branch once it's available!**

This PR adds the Tenant module without business logic, to be used in the future when we have business logic to add.

* Added new folder to `03_Modules` called `Tenant` to house all Tenant-module related files in the same structure as the other modules
  * Tenant version is at 1.3.0 to match with release candidate
* Updated `Server/docker-compose.yml` and `Server/docker-compose.swarm.yml` to spin up new Tenant module
* Updated `Swarm` folder-related modules to include Tenant module
  * Can be reverted if not necessary (as I see that items in the Swarm folder haven't been updated too recently)
*  Updated RabbitMQ `receiver.ts` to disallow queue creation if a module has no actions
  * Added a clarifying comment to this since chargers won't have actions and should get a queue
  * This was done because otherwise RabbitMQ will erroneously feed all messages to the Tenant module


03_Modules/Tenant/src/module/api.ts
```diff
@@ -0,0 +1,56 @@
...
+import { FastifyInstance } from 'fastify';
+import { ILogObj, Logger } from 'tslog';
+
+/**
+ * Server API for the Tenant module.
+ */
+export class TenantModuleApi
+  extends AbstractModuleApi<TenantModule>
+  implements ITenantModuleApi
+{
```
yuvenus, line 18: it seems the github actions linter complains here, but i ran the `npm run lint-fix` command locally and arrived at this formatting 🤔 should there be another local command i run instead?

yuvenus:
Closing PR to move branch over to the actual repo instead of a fork!

PR #104, opened 2024-05-23
(1.3.0) Tenant Module

yuvenus:
This PR adds the Tenant module without business logic, to be used in the future when we have business logic to add.

* Added new folder to `03_Modules` called `Tenant` to house all Tenant-module related files in the same structure as the other modules
  * Tenant version is at 1.3.0 to match with release candidate
* Updated `Server/docker-compose.yml` and `Server/docker-compose.swarm.yml` to spin up new Tenant module
* Updated `Swarm` folder-related modules to include Tenant module
  * Can be reverted if not necessary (as I see that items in the Swarm folder haven't been updated too recently)
*  Updated RabbitMQ `receiver.ts` to disallow queue creation if a module has no actions
  * Added a clarifying comment to this since chargers won't have actions and should get a queue
  * This was done because otherwise RabbitMQ will erroneously feed all messages to the Tenant module


Server/docker-compose.swarm.yml
```diff
@@ -220,6 +220,28 @@ services:
       - 8089:8089
     expose:
       - 8089
+    volumes:
+      - ./:/usr/tenant
+      - /usr/tenant/node_modules
+    environment:
+      CITRINEOS_UTIL_DIRECTUS_USERNAME: 'admin@citrineos.com'
```
yuvenus, line 227: i saw in another PR the username/password was going to be replaced with a CITRINEOS_UTIL_DIRECTUS_TOKEN -- if i should also update here then i can do so, let me know!

Server/src/config/envs/swarm.docker.ts
```diff
@@ -48,10 +48,15 @@ export function createDockerConfig() {
         host: '0.0.0.0',
         port: 8088,
       },
+      tenant: {
+        endpointPrefix: 'tenant',
+        host: '0.0.0.0',
+        port: 8089,
```
yuvenus, line 54: i updated the port because i wanted to keep the alphabetical ordering of the modules -- if it's preferred to not keep the alphabetical ordering and instead add the tenant to the "end" of the list, thereby making it port 8090, i can do so, let me know!

03_Modules/Tenant/package.json
```diff
@@ -0,0 +1,33 @@
+{
+  "name": "@citrineos/tenant",
+  "version": "1.3.0",
```
yuvenus, line 3: version is 1.3.0 to match release candidate, but if there's a separate version convention i can update, let me know!

lydiazcheng (approved):

Server/docker-compose.swarm.yml
yuvenus: ah, makes sense! for consistency i've updated to also use the token in the tenant module docker config, now that those changes are merged in main 🙌

Server/src/config/envs/swarm.docker.ts
ChrisWeissmann: I think keeping the alphabetical ordering is good. But changing the ports seems risky, as others and integration tests could already depend on the ports of certain modules.

00_Base/src/config/types.ts
```diff
@@ -79,6 +79,13 @@ export const systemConfigInputSchema = z.object({
         port: z.number().int().positive().default(8081).optional(),
       })
       .optional(),
+    tenant: z
+      .object({
+        endpointPrefix: z.string().default(EventGroup.Tenant).optional(),
+        host: z.string().default('localhost').optional(),
+        port: z.number().int().positive().default(8081).optional(),
```
ChrisWeissmann, line 86: This might be a stupid questions, but wouldn't we want the default to be 8080? Why is the default 8081?

Server/src/config/envs/swarm.docker.ts
yuvenus: makes sense! then i've kept the alphabetical ordering, but the ports will still be in order of when the modules were created (meaning tenant now inherits 8090 and transaction returns to 8089) 👍

00_Base/src/config/types.ts
yuvenus: if i'm understanding correctly, this is the config for the websocket which is why it's 8081 🤔

03_Modules/Tenant/package.json
```diff
@@ -0,0 +1,33 @@
...
+    "ocpp_v201"
+  ],
+  "license": "Apache-2.0",
+  "devDependencies": {
+    "@types/deasync-promise": "^1.0.0",
+    "eslint": "^8.48.0",
+    "typescript": "^5.0.4"
+  },
+  "dependencies": {
+    "@citrineos/base": "1.1.1",
```
ChrisWeissmann, line 28: Probably should update these dependancies here.

yuvenus: yup good point, updating this fixed the build also, woohoo!

ChrisWeissmann (approved):
Looks good to me

PR #105, opened 2024-05-28
Fix/query by pk

ChrisWeissmann:
Small fix to base of  read by key, which now finds by any primary key instead of only `id`. Also made delete by key transactional and able to delete by a primary key that is not `id`


.github/workflows/push-release-tagged-server.yml
thanaParis, line 1: Are all these changes meant to be here too?

00_Base/json-schema-processor.js
thanaParis, line 1: I don't think these changes are intentional?

.github/workflows/push-release-tagged-server.yml
thanaParis: okay, I think the commits got screwed up by some upstream merges
these changes are correct, they seem to have been undone in rc-1.2.0 by accident. We'll use this pr to fix them.

thanaParis (approved):
Approved.

PR #109, opened 2024-05-31
fix: fixing systemConfig route to return configuration object without failure

danielyoo227:
Fixes the previous error on calling systemConfig endpoint:
```
{
    "statusCode": 500,
    "error": "Internal Server Error",
    "message": "method.call(...).catch is not a function"
}
```

- changing addDataRoute call to specify method type to a Promise
- changing addDataRoute parameter methods for systemConfig namespace to return a promise


lydiazcheng (approved):
The change looks good. Also run it locally and endpoints work.

PR #110, opened 2024-06-03
Add .nvmrc to automatically specify the node version

ChrisWeissmann:
Since we mention only allowing node 18+ we should add an `.nvmrc` to automatically switch to the correct node this for users that have nvm installed.

- chore: add .nvmrc to automatically specify the node version


thanaParis (approved):
approved!

PR #111, opened 2024-06-03
Resolve onChange when value is updated in MemoryCache

falanadamian:
Current MemoryCache implementation does not resolve "onChange" promises when the value associated with a namespace key changes. This PR adds the invocation of "onChange" callback function to resolve promises returned to listeners upon value change. (This is how I understand the "onChange" functionality is expected to work?)


thanaParis (approved):
This was definitely an oversight. Approved!

Issue #17, opened 2023-11-09
Runnig docker compose fails to start citrine service


couryrr-afs:
I ran through the setup for the project. When I run the `docker compose up -d` command it looks like everything ran fine. Even checking via Docker Desktop all of the services are green. I was having issues getting a connection and dug into it. I found that the citrine service was started but stuck in an error state. 

I can kill the service and run the local version just fine.

```
2023-11-09 13:50:50 /usr/server/node_modules/bindings/bindings.js:126
2023-11-09 13:50:50   err = new Error(
2023-11-09 13:50:50         ^
2023-11-09 13:50:50 Error: Could not locate the bindings file. Tried:
2023-11-09 13:50:50  → /usr/server/node_modules/deasync/build/deasync.node
2023-11-09 13:50:50  → /usr/server/node_modules/deasync/build/Debug/deasync.node
2023-11-09 13:50:50  → /usr/server/node_modules/deasync/build/Release/deasync.node
2023-11-09 13:50:50  → /usr/server/node_modules/deasync/out/Debug/deasync.node
2023-11-09 13:50:50  → /usr/server/node_modules/deasync/Debug/deasync.node
2023-11-09 13:50:50  → /usr/server/node_modules/deasync/out/Release/deasync.node
2023-11-09 13:50:50  → /usr/server/node_modules/deasync/Release/deasync.node
2023-11-09 13:50:50  → /usr/server/node_modules/deasync/build/default/deasync.node
2023-11-09 13:50:50  → /usr/server/node_modules/deasync/compiled/18.18.2/linux/arm64/deasync.node
2023-11-09 13:50:50  → /usr/server/node_modules/deasync/addon-build/release/install-root/deasync.node
2023-11-09 13:50:50  → /usr/server/node_modules/deasync/addon-build/debug/install-root/deasync.node
2023-11-09 13:50:50  → /usr/server/node_modules/deasync/addon-build/default/install-root/deasync.node
2023-11-09 13:50:50  → /usr/server/node_modules/deasync/lib/binding/node-v108-linux-arm64/deasync.node
2023-11-09 13:50:50     at bindings (/usr/server/node_modules/bindings/bindings.js:126:9)
2023-11-09 13:50:50     at Object.<anonymous> (/usr/server/node_modules/deasync/index.js:30:31)
2023-11-09 13:50:50     at Module._compile (node:internal/modules/cjs/loader:1256:14)
[...]

PR #113, opened 2024-06-04
Replace upsert

ChrisWeissmann:
Replacing upsets as they seem to exhibit unreliable behavior causing bugs. Also fixed naming overlap for repository methods that not calling the actual super.method.

### DIFF
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -36,6 +36,19 @@
       }
     ],
     "@typescript-eslint/indent": "off",
+    "@typescript-eslint/member-delimiter-style": [
+      "error",
+      {
+        "multiline": {
+          "delimiter": "semi",
+          "requireLast": true
+        },
+        "singleline": {
+          "delimiter": "semi",
+          "requireLast": false
+        }
+      }
+    ],
     "@typescript-eslint/member-ordering": "warn",
     "@typescript-eslint/naming-convention": [
       "error",
@@ -47,6 +60,7 @@
       }
     ],
     "@typescript-eslint/no-empty-function": "off",
+    "@typescript-eslint/no-empty-interface": "error",
     "@typescript-eslint/no-inferrable-types": [
       "error",
       {
@@ -64,6 +78,9 @@
     "@typescript-eslint/no-unused-expressions": "error",
     "@typescript-eslint/no-use-before-define": "error",
     "@typescript-eslint/prefer-function-type": "error",
+    "@typescript-eslint/quotes": ["error", "single"],
+    "@typescript-eslint/semi": ["error", "always"],
+    "@typescript-eslint/type-annotation-spacing": "error",
     "@typescript-eslint/unified-signatures": "error",
     "@typescript-eslint/no-explicit-any": "off",
     "@typescript-eslint/strict-boolean-expressions": "off",
@@ -78,9 +95,11 @@
       }
     ],
     "arrow-body-style": "error",
+    "brace-style": ["error", "1tbs"],
     "constructor-super": "error",
     "curly": "error",
     "dot-notation": "off",
+    "eol-last": "error",
     "eqeqeq": ["error", "smart"],
     "guard-for-in": "error",
     "id-denylist": "off",
@@ -125,6 +144,7 @@
     "no-restricted-imports": ["error", "rxjs/Rx"],
     "no-shadow": "off",
     "no-throw-literal": "error",
+    "no-trailing-spaces": "error",
     "no-undef-init": "error",
     "no-underscore-dangle": "off",
     "no-unused-expressions": "off",
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -37,7 +37,7 @@
     ],
     "@typescript-eslint/indent": "off",
     "@typescript-eslint/member-delimiter-style": [
-      "error",
+      "off",
       {
         "multiline": {
           "delimiter": "semi",
@@ -60,7 +60,6 @@
       }
     ],
     "@typescript-eslint/no-empty-function": "off",
-    "@typescript-eslint/no-empty-interface": "error",
     "@typescript-eslint/no-inferrable-types": [
       "error",
       {
@@ -78,9 +77,9 @@
     "@typescript-eslint/no-unused-expressions": "error",
     "@typescript-eslint/no-use-before-define": "error",
     "@typescript-eslint/prefer-function-type": "error",
-    "@typescript-eslint/quotes": ["error", "single"],
-    "@typescript-eslint/semi": ["error", "always"],
-    "@typescript-eslint/type-annotation-spacing": "error",
+    "@typescript-eslint/quotes": ["off", "single"],
+    "@typescript-eslint/semi": ["off", "always"],
+    "@typescript-eslint/type-annotation-spacing": "off",
     "@typescript-eslint/unified-signatures": "error",
     "@typescript-eslint/no-explicit-any": "off",
     "@typescript-eslint/strict-boolean-expressions": "off",
@@ -95,11 +94,11 @@
       }
     ],
     "arrow-body-style": "error",
-    "brace-style": ["error", "1tbs"],
+    "brace-style": ["off", "1tbs"],
     "constructor-super": "error",
     "curly": "error",
     "dot-notation": "off",
-    "eol-last": "error",
+    "eol-last": "off",
     "eqeqeq": ["error", "smart"],
     "guard-for-in": "error",
     "id-denylist": "off",
@@ -144,7 +143,7 @@
     "no-restricted-imports": ["error", "rxjs/Rx"],
     "no-shadow": "off",
     "no-throw-literal": "error",
-    "no-trailing-spaces": "error",
+    "no-trailing-spaces": "off",
     "no-undef-init": "error",
     "no-underscore-dangle": "off",
     "no-unused-expressions": "off",
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -36,19 +36,6 @@
       }
     ],
     "@typescript-eslint/indent": "off",
-    "@typescript-eslint/member-delimiter-style": [
-      "error",
-      {
-        "multiline": {
-          "delimiter": "semi",
-          "requireLast": true
-        },
-        "singleline": {
-          "delimiter": "semi",
-          "requireLast": false
-        }
-      }
-    ],
     "@typescript-eslint/member-ordering": "warn",
     "@typescript-eslint/naming-convention": [
       "error",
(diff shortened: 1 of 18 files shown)


.eslintrc.json
```diff
@@ -144,7 +125,6 @@
     "no-restricted-imports": ["error", "rxjs/Rx"],
     "no-shadow": "off",
     "no-throw-literal": "error",
-    "no-trailing-spaces": "error",
     "no-undef-init": "error",
     "no-underscore-dangle": "off",
     "no-unused-expressions": "off",
```
lydiazcheng, line 130: Personally, I feel these removed rules are very helpful to maintaining a good formatted code base, especially, like no trailing spaces, delimiter, single quotes and so on. Instead of removing the rules, I think it would be better to add eslint disable, like`/* eslint-disable @typescript-eslint/member-ordering */` at the code that we intend to ignore the check due to a specific reason. If we indeed want to get rid of a rule, how about trying replace "error" with "off"? Then we could reset it when we need it again in the future.

00_Base/src/interfaces/repository.ts
```diff
@@ -62,7 +62,6 @@ export abstract class CrudRepository<T> extends EventEmitter {
     this.emit('created', [result]);
```
lydiazcheng, line 62: Minor: How about using /* eslint-disable @typescript-eslint/member-ordering */? In this case, we don't need to change the existed code and methods for the same db operation can be stay together.

lydiazcheng: Could you please also help fix the warning, Forbidden non-null assertion in the `Authorization`? The fix can be like
```
if (updatedValue !== undefined && savedIdTokenInfo) {
      // Null can still be used to remove data
     savedIdTokenInfo.setDataValue(k, valueIdTokenInfo.getDataValue(k));
}
```
Because `savedIdTokenInfo!` means that we are sure `savedIdTokenInfo` is always not null, then add undefine/null check should not make difference.

.eslintrc.json
ChrisWeissmann: The reason I got rid of them is because we are using eslint in combination with prettier. We extend prettier in the eslint config, as described in the prettier docs here: https://prettier.io/docs/en/integrating-with-linters 
However there are a few rules in our config that pertain to style, not to linting and cause conflict between prettier and eslint. I have set to automatically format according to our prettierrc any time I save. But this then causes an error with eslint. According to the prettier docs found here: https://github.com/prettier/eslint-config-prettier you can run `npx eslint-config-prettier path/to/main.js` that will tell you which eslint rules conflict with prettier. So I removed those in the commit https://github.com/citrineos/citrineos-core/pull/113/commits/615d88b7af2e3ed0bac12bce8f2eee3f754caa9a

00_Base/src/interfaces/repository.ts
ChrisWeissmann: I am open to ignoring the eslint rule here. Personally, I would prefer a repetitive structure which eslint enforces. Even if the underlying `_query` belong together, when looking at the code, you have it in an IDE, so you can just click jump to the impl. 

In regards to the non-null assertion in `Authorization` I was a bit hesitant due to the commented line `// Null can still be used to remove data` which I interpret as we want to explicitly allow null here. So if we add a null check, we will not be allowing null to execute. I don't have the full context on functionality, as I was just fixing minor errors and didn't take the time to deep dive into this section. @thanaParis any thoughts?

01_Data/src/layers/sequelize/repository/Certificate.ts
```diff
@@ -7,28 +7,30 @@ import { ICertificateRepository } from '../../../interfaces';
...
+          commonName: certificate.commonName,
+        },
+        transaction,
+      });
+      if (!savedCert) {
+        await certificate.save({ transaction });
+      } else {
+        await certificate.update(certificate, { transaction });
+      }
+      return await certificate.reload({ transaction });
```
lydiazcheng, line 33: Do we need to send the emit event here? Sry I am a bit confused about when to emit and when not.

00_Base/src/interfaces/repository.ts
thanaParis: > I am open to ignoring the eslint rule here. Personally, I would prefer a repetitive structure which eslint enforces. Even if the underlying _query belong together, when looking at the code, you have it in an IDE, so you can just click jump to the impl.

I could go either way on this. If more people feel the eslint method of organization is better, we should change it.

> In regards to the non-null assertion in Authorization I was a bit hesitant due to the commented line // Null can still be used to remove data which I interpret as we want to explicitly allow null here.

Yes, if a field has been removed vs a field was not included, the first is null the second is undefined. So explicitly allowing null is important here.

01_Data/src/layers/sequelize/repository/DeviceModel.ts
```diff
@@ -65,15 +64,28 @@ export class SequelizeDeviceModelRepository extends SequelizeRepository<Variable
...
+      await this.s.transaction(async (transaction) => {
+        const savedVariableCharacteristics = await this.s.models[VariableCharacteristics.MODEL_NAME].findOne({
+          where: {
+            variableId: variable.id,
+          },
+          transaction,
+        });
+
+        if (!savedVariableCharacteristics) {
+          await vc.save({ transaction });
```
lydiazcheng, line 84: Same emit question here. If we need emit, then we might use CrudRepository here as well as for update.
```suggestion
          await this.variableCharacteristics.create(vc);
```

01_Data/src/layers/sequelize/repository/DeviceModel.ts
```diff
@@ -65,15 +64,28 @@ export class SequelizeDeviceModelRepository extends SequelizeRepository<Variable
...
-          variableId: variable.id,
-        }),
-      );
-      dataType = variableCharacteristics.dataType;
+      dataType = value.variableCharacteristics.dataType;
+      const vc = VariableCharacteristics.build({
+        ...value.variableCharacteristics,
+        variableId: variable.id,
+      });
+      console.log(`variableCharacteristics: ${JSON.stringify(vc)}`);
```
lydiazcheng, line 74: Maybe using logger? I am curious whether console logs also work in Cloud Watch.

01_Data/src/layers/sequelize/repository/Certificate.ts
```diff
@@ -7,28 +7,30 @@ import { ICertificateRepository } from '../../../interfaces';
...
-      commonName: certificate.commonName,
-      validBefore: certificate.validBefore,
-      signatureAlgorithm: certificate.signatureAlgorithm,
-      countryName: certificate.countryName,
-      isCA: certificate.isCA,
-      pathLen: certificate.pathLen,
-      certificateFileId: certificate.certificateFileId,
-      privateKeyFileId: certificate.privateKeyFileId,
+    return await this.s.transaction(async (transaction) => {
+      // TODO what to search by
```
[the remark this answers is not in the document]
lydiazcheng: > @lydiazcheng could you fix this todo? I imagine you know what we could search by to find the certificate

Hi Thana, the fix is ready in this PR: https://github.com/citrineos/citrineos-core/pull/117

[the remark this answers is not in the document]
ChrisWeissmann: @thanaParis @lydiazcheng #17 has been merged into this branch 🎉

01_Data/src/layers/sequelize/repository/Certificate.ts
ChrisWeissmann: Hmm. I think you are correct that it should emit here and that we should rather use CRUD repository. Before, it use the models like I did here. Which didn't emit. But I think we should be consistent and use the CRUD repo methods. Making the adjustment.

ChrisWeissmann: I have gone over my changes again. So the pattern should be that if we are directly creating/updating we use the CRUD repo to emit. If we instead are creating, then attaching more info to the same object, we would want to emit the full object, not just the half one where we don't have all fields set yet. Example would be the Boot one.

01_Data/src/layers/sequelize/repository/MessageInfo.ts
```diff
@@ -30,14 +30,26 @@ export class SequelizeMessageInfoRepository extends SequelizeRepository<MessageI
...
+      const messageInfo = MessageInfo.build({
         stationId: stationId,
         componentId: componentId,
         ...message,
         active: true,
-      }),
-    );
-    return savedMessageInfo;
+      });
+      // TODO is this searching correctly?
```
lydiazcheng, line 40: As describing in the OCPP 2.0.1, the id of MessageInfoType is unique within an exchange context. It is defined within the OCPP context as a positive Integer value (greater or equal to zero). My understanding is that the id is unique for each charger. So stationId and message.id should be used for finding a unique one. (An composite unique constraint on these 2 field is also set in the messageInfo table.)

01_Data/src/layers/sequelize/repository/Tariff.ts
```diff
@@ -23,14 +23,20 @@ export class SequelizeTariffRepository extends SequelizeRepository<Tariff> imple
...
   async createOrUpdateTariff(tariff: Tariff): Promise<Tariff> {
-    const [storedTariff, _tariffCreated] = await this.upsert(
-      Tariff.build({
-        stationId: tariff.stationId,
-        unit: tariff.unit,
-        price: tariff.price,
-      }),
-    );
-    return storedTariff;
+    // TODO check if searchg correclty
```
lydiazcheng, line 26: Yes, it would work for the current tariff workaround. I found the unique constraint is missing in the Tariff model. Could you please help add it in 01_Data/src/layers/sequelize/model/Tariff/Tariffs.ts? It looks like,
```
  @Index
  @Column({
    type: DataType.STRING,
    unique: 'stationId_unit'
  })
  declare stationId: string;

  @Column({
    type: DataType.STRING,
    unique: 'stationId_unit'
  })
  declare unit: TariffUnitEnumType;
```

ChrisWeissmann: Sure will add the constraints to the model :)

01_Data/src/layers/sequelize/repository/MessageInfo.ts
ChrisWeissmann: Got it, changed to search by message.id

lydiazcheng (approved):

00_Base/src/interfaces/repository.ts
ChrisWeissmann: The people have spoken. We will be going stricter and try to automate these things :)

PR #114, opened 2024-06-05
Database querying issues

falanadamian:
Some specialized repositories offer methods for filtering by *Querystring objects. Currently, these methods override some base repository methods (inherited from SequelizeRepository and CrudRepository), which accept Sequelize's FindOptions object for filtering. However, some of these methods are internally used by base repositories and expect Sequelize's FindOptions object instead of *Querystring object. As a result, some queries are not applying any filtering or criteria. 

This PR fixes the issue by simply renaming these methods. Another approach could be adjusting all clients to pass appropriate *Querystring objects, but renaming methods seems safer and more flexible for now.


ChrisWeissmann (approved):
Looks good to me. I might have to do some naming adjustments on my PR to avoid conflict.

thanaParis (approved):
Perfect! Approved.

PR #115, opened 2024-06-05
Create API to update password for charging stations

falanadamian:
This PR introduces an API for updating passwords on charging stations. 

The implementation currently follows a naive approach: upon receiving a request to update the password, we first send a request to the charging station (if specified by client). If the request is successful, we update the database accordingly. 
However, if an error occurs between sending the request and updating the database (e.g. application crash) or while updating the database, the system would become inconsistent.

I opted for simplicity in current implementation, considering that this API will probably be rarely used by admin users; in case of an error, perhaps manual retrying of the operation by admins could be enough?
Additionally, (default) waiting time for a response from the charging station is 10 seconds, potentially blocking API clients for that duration. Should we consider making this operation asynchronous to avoid blocking?
Moreover, if the charging station responds after the waiting time has elapsed, we would not update the database, although the charger might have been successfully updated. Is this acceptable?

Maybe we should track password update process in the database and act accordingly when we receive event associated with that process (identified by uuid) from charging station? @thanaParis should I enhance this flow to ensure eventual consistency?


03_Modules/Configuration/src/module/api.ts
```diff
@@ -310,6 +325,125 @@ export class ConfigurationModuleApi
...
+        this._logger.warn(`Failed updating password on ${stationId} station`, error);
+        return {success: false, payload: `Failed updating password on ${stationId} station`};
+      }
+    }
+    const variableAttributes = await this.updatePasswordForStation(password, stationId);
+    this._logger.debug(`Successfully updated password for ${stationId} station`);
+    return {success: true, payload: `Updated ${variableAttributes.length} attributes`};
+  }
+
+  private async updatePasswordOnStation(password: string, stationId: string, callbackUrl?: string): Promise<void> {
```
falanadamian, line 360: Should I move this logic to a service layer instead? (I am uncertain about the convention; I have observed that many APIs are implemented directly within the API layer?)

03_Modules/Configuration/src/module/api.ts
```diff
@@ -310,6 +325,125 @@ export class ConfigurationModuleApi
...
+
+    const setVariablesResponse: SetVariablesResponse = JSON.parse(responseJsonString);
+    const passwordUpdated = setVariablesResponse.setVariableResult
+        .every(result => result.attributeStatus == SetVariableStatusEnumType.Accepted);
+    if (!passwordUpdated) {
+      throw new Error(`Failure updating password on ${stationId} station`);
+    }
+  }
+
+  private async updatePasswordForStation(password: string, stationId: string): Promise<VariableAttribute[]> {
```
falanadamian, line 403: Should I move this logic to a service layer instead?

03_Modules/Configuration/src/module/api.ts
```diff
@@ -310,6 +325,125 @@ export class ConfigurationModuleApi
...
+      throw new Error(`Failed sending request to ${stationId} station for updating password`);
+    }
+
+    const responseJsonString = await cacheCallbackPromise;
+    if (!responseJsonString) {
+      throw new Error(`${stationId} station did not respond in time for updating password`);
+    }
+
+    const setVariablesResponse: SetVariablesResponse = JSON.parse(responseJsonString);
+    const passwordUpdated = setVariablesResponse.setVariableResult
```
falanadamian, line 396: Will this check suffice, or should we specifically verify that the status is accepted for SecurityCtrlr and BasicAuthPassword?

00_Base/src/ocpp/model/schemas/UpdateChargingStationPasswordRequestSchema.json
```diff
@@ -0,0 +1,21 @@
+{
```
falanadamian, line 1: Is this the appropriate location for placing this file?

00_Base/src/ocpp/persistence/namespace.ts
```diff
@@ -36,4 +36,5 @@ export enum Namespace {
   VariableMonitoringStatus = 'VariableMonitoringStatus',
   VariableStatus = 'VariableStatus',
   VariableType = 'Variable',
+  PasswordType = 'Password',
```
falanadamian, line 39: Is it fine to create a new namespace?

00_Base/src/ocpp/model/types/UpdateChargingStationPasswordRequest.ts
```diff
@@ -0,0 +1,12 @@
+// Copyright (c) 2023 S44, LLC
```
falanadamian, line 1: Is this the appropriate location for placing this file?

01_Data/src/interfaces/queries/UpdateChargingStationPasswordQuery.ts
```diff
@@ -0,0 +1,11 @@
+// Copyright Contributors to the CitrineOS Project
```
falanadamian, line 1: Is this the appropriate location for placing this file?

03_Modules/Configuration/src/module/api.ts
ChrisWeissmann: I think a service layer should be introduced and moved to there. API layer can keep logic on validating the request.

00_Base/src/ocpp/model/schemas/UpdateChargingStationPasswordRequestSchema.json
thanaParis: No. This folder is for the automatically generated files that are output by json-schema-processor. We should create a new folder for this class, probably underneath the 'persistence' folder.

00_Base/src/ocpp/persistence/namespace.ts
thanaParis: Yes!

01_Data/src/interfaces/queries/UpdateChargingStationPasswordQuery.ts
thanaParis: Yes!

03_Modules/Configuration/src/module/api.ts
thanaParis: I'm not sure what you mean. This check will already return false if there were any issue with SecurityCtrlr or BasicAuthPassword. It looks good 👍

03_Modules/Configuration/src/module/api.ts
thanaParis: There is no such concept that's been universally applied, as of yet. It's fine here for now.

thanaParis:
I want to see this pr accompanied by an updated quickstart guide on the github.io page. The primary purpose of adding this functionality is to make security profiles 1 and 2 more accessible when new to the project & protocol. When we have a pr open to update the guide, we can merge them together 👍

thanaParis (changes requested):
Mostly looks good! Some files need to be moved, and a second pr needs to be opened on the io repo.

03_Modules/Configuration/src/module/api.ts
falanadamian: Sounds good!

falanadamian:
> I want to see this pr accompanied by an updated quickstart guide on the github.io page. The primary purpose of adding this functionality is to make security profiles 1 and 2 more accessible when new to the project & protocol. When we have a pr open to update the guide, we can merge them together 👍

@thanaParis makes sense, should I update the quickstart guide or is someone already working on it?

00_Base/src/ocpp/model/schemas/UpdateChargingStationPasswordRequestSchema.json
falanadamian: Got it, I moved it under persistence/schemas, thanks @thanaParis!

thanaParis:
> @thanaParis makes sense, should I update the quickstart guide or is someone already working on it?

Please do! No one else is already working on it. [This is the guide I mean](https://github.com/citrineos/citrineos.github.io/blob/main/quickstart.md).

falanadamian:
> Please do! No one else is already working on it. [This is the guide I mean](https://github.com/citrineos/citrineos.github.io/blob/main/quickstart.md).

@thanaParis I created a PR to update the quickstart guide:
https://github.com/citrineos/citrineos.github.io/pull/22
Please let me know if I should include some more information there.

falanadamian:
@thanaParis @ChrisWeissmann just a quick thought – would it make sense to validate that a password must always be present in the request if setOnCharger is true? It probably doesn't make much sense to update the device model with an autogenerated password but not update the charger itself at the same time? I can add that validation real quick if that works for you?

ChrisWeissmann:
@falanadamian yea, I think that is a good point and good thing to add.

thanaParis (approved):
Approved

PR #116, opened 2024-06-05
Smart Charging Part 1: Set Charging Profile

lydiazcheng:
## Changes
1. Add validation for ChargingProfileType in `02_Util/src/util/validator.ts`. This method validates constraints defined in ChargingProfileType and its nested objects. The method can be used in different modules.
2. Implement K01 SetChargingProfile use case.
   - Create new table and repository for persist charging profiles
   - Validate use case specific constraints.
3. Implement partial NotifyEVChargingNeeds to support the SetChargingProfile flow.

### Test Charging Profile with purpose, TxProfile with received NotifyEVChargingNeedsRequest - Validation Succeed
1. Prepare: Send a `TransactionEventRequest` and a `NotifyEVChargingNeedsRequest` to citrine and the entities are stored in DB. 
2. Send a Charging Profile purpose, TxProfile and transactionId. We expect the charging need is found using the transactionId and the validation success. Screenshot 2024-06-11 at 3 38 26 PM Screenshot 2024-06-11 at 3 38 43 PM
### Test send a ChargingProfile with a stackLevel - transactionId combination that already exists in another ChargingProfile (with different id) with purpose TxProfile (K01.FR.39) - Validation Failed Screenshot 2024-06-11 at 2 53 15 PM


03_Modules/SmartCharging/src/module/api.ts
```diff
@@ -91,12 +97,372 @@ export class SmartChargingModuleApi
...
+          identifier,
+          chargingProfile.transactionId,
+        ))
+      ) {
+        return {
+          success: false,
+          payload: `Transaction ${chargingProfile.transactionId} not found on station ${identifier}.`,
+        };
+      }
+      if (request.evseId <= 0) {
```
ChrisWeissmann, line 168: Nitpick. Quick comment of requirement reference
```suggestion
      // OCPP 2.0.1 Part 2 K01.FR.16
      if (request.evseId <= 0) {
```

03_Modules/SmartCharging/src/module/api.ts
```diff
@@ -91,12 +97,372 @@ export class SmartChargingModuleApi
     CallAction.SetChargingProfile,
     SetChargingProfileRequestSchema,
   )
-  setChargingProfile(
+  async setChargingProfile(
```
ChrisWeissmann, line 100: I think we might be missing `K01.FR.26` 
> When a `SetChargingProfileRequest` is received with a value for `chargingRateUnit`, that is not configured in the configuration variable `ChargingScheduleChargingRateUnit`.

03_Modules/SmartCharging/src/module/module.ts
```diff
@@ -224,6 +280,12 @@ export class SmartChargingModule extends AbstractModule {
     props?: HandlerProperties,
   ): void {
     this._logger.debug('SetChargingProfile response received:', message, props);
+    const response: SetChargingProfileResponse = message.payload;
+    if (response.status === ChargingProfileStatusEnumType.Rejected) {
```
ChrisWeissmann, line 284: Should we associate the response to the object we saved when we made the request?

ChrisWeissmann (commented):
Great job on this tough use case! I left a few comments when going through the spec.

03_Modules/SmartCharging/src/module/module.ts
lydiazcheng: > Should we associate the response to the object we saved when we made the request?

There is no identifier in the SetChargingProfileResponse which enables us to find the corresponding stored object. There is a plan to find out a solution to match the response and request and refactor the current code. But since we are busy on other topics, this improvement hasn't been done.

lydiazcheng (commented):
Thanks for the review, @ChrisWeissmann ! I change the code.

03_Modules/SmartCharging/src/module/api.ts
```diff
@@ -91,12 +100,322 @@ export class SmartChargingModuleApi
...
+          return -1;
+        }
+        return 0;
+      });
+
+      for (const chargingSchedulePeriod of chargingSchedule.chargingSchedulePeriod) {
+        if (chargingSchedulePeriod.phaseToUse) {
+          // OCPP 2.0.1 Part 2 K01.FR.19
+          if (chargingSchedulePeriod.numberPhases !== 1) {
+            throw new Error(
```
ChrisWeissmann, line 396: This doesn't look like it's gets caught anywhere, like the ones in the validator. So this won't get formatted into the response format.

03_Modules/SmartCharging/src/module/api.ts
```diff
@@ -91,12 +100,322 @@ export class SmartChargingModuleApi
...
+              `ChargingSchedule ${chargingSchedule.id}: PhaseToUse SHALL only be set with numberPhases = 1.`,
+            );
+          }
+          // OCPP 2.0.1 Part 2 K01.FR.20
+          if (
+            acPhaseSwitchingSupported.length === 0 ||
+            !acPhaseSwitchingSupported[0].value
+          ) {
+            throw new Error(
```
ChrisWeissmann, line 405: This doesn't look like it's gets caught anywhere, like the ones in the validator. So this won't get formatted into the response format. 
Probably should just return in the correct format.

03_Modules/SmartCharging/src/module/api.ts
lydiazcheng: Thank you, Christian! I fixed them.

ChrisWeissmann (approved):
Looks good to me now 🚀

PR #119, opened 2024-06-07
Smart Charging Part 2: Get and Clear Charging Profile

lydiazcheng:
## Changes
1. Implement Get and Clear Charging Profile (K09 & K10)
2. Fix ChargingProfile model and repository to make it work with new use cases

### Test Get Charging Profiles - Basic flow
1. Citrine sends GetChargingProfilesRequest Screenshot 2024-06-07 at 2 01 31 PM  Screenshot 2024-06-07 at 2 02 31 PM
2. Charger sends ReportChargingProfilesRequest and profiles are saved in DB. The isActive set to true. Screenshot 2024-06-07 at 2 27 04 PM Screenshot 2024-06-07 at 2 27 45 PM
### Test Get Charging Profiles - Profile ids length exceed max limit
The max limit is 2 but give 5 charging profile ids in the request Screenshot 2024-06-07 at 5 35 56 PM
### Test trigger GetChargingProfilesRequest by Set Charging Profile response
1. Citrine send set charging profile req Screenshot 2024-06-07 at 5 41 59 PM
2. Charger send a Set Charging Profile Accept response with the same request id and it receives a GetChargingProfilesRequest Screenshot 2024-06-07 at 5 39 11 PM
### Test trigger GetChargingProfilesRequest by Clear Charging Profile response
1. Citrine send clear charging profile req
2. Charger send back an Accept response use the same req id.
3. When Citrine receive this response it send GetChargingProfiles to get the latest status of charging profiles and the existed profile is set to `isActive` false Screenshot 2024-06-07 at 5 52 45 PM Screenshot 2024-06-07 at 5 53 01 PM


03_Modules/SmartCharging/src/module/api.ts
```diff
@@ -59,12 +60,30 @@ export class SmartChargingModuleApi
...
     ClearChargingProfileRequestSchema,
   )
-  clearChargingProfile(
+  async clearChargingProfile(
     identifier: string,
     tenantId: string,
     request: ClearChargingProfileRequest,
     callbackUrl?: string,
   ): Promise<IMessageConfirmation> {
+    if (!request.chargingProfileId && !request.chargingProfileCriteria) {
```
ChrisWeissmann, line 69: What about the case `include one or more of the fields stackLevel, evseId and chargingProfilePurpose` at `K10.FR.02` Would that already be covered here?

03_Modules/SmartCharging/src/module/api.ts
```diff
@@ -78,12 +97,34 @@ export class SmartChargingModuleApi
...
     GetChargingProfilesRequestSchema,
   )
-  getChargingProfile(
+  async getChargingProfile(
     identifier: string,
     tenantId: string,
     request: GetChargingProfilesRequest,
     callbackUrl?: string,
   ): Promise<IMessageConfirmation> {
+    if (
```
ChrisWeissmann, line 106: If ChargingProfileId is not set. Don't we have to validate that one of the other conditions is met according to `K09.FR.03` :

> The CSMS SHALL specify in chargingProfile criteria in GetChargingProfilesRequest either:
> - a (list of) chargingProfileId(s) OR
> - one or more of the fields stackLevel, chargingLimitSource, chargingProfilePurpose.

03_Modules/SmartCharging/src/module/module.ts
```diff
@@ -228,12 +233,24 @@ export class SmartChargingModule extends AbstractModule {
...
     message: IMessage<ReportChargingProfilesRequest>,
     props?: HandlerProperties,
-  ): void {
+  ): Promise<void> {
     this._logger.debug('ReportChargingProfiles received:', message, props);
 
+    const chargingProfiles = message.payload
+      .chargingProfile as ChargingProfileType[];
+    for (const chargingProfile of chargingProfiles) {
+      await this._chargingProfileRepository.createOrUpdateChargingProfile(
```
ChrisWeissmann, line 245: Take a look at K09 the `Remark`:

> The charging profiles report can be split over multiple ReportChargingProfilesRequest messages, this can be because charging profiles for different charging sources need to be reported, or because there is just to much data for one message. To indicate that more reports will follow the flag tbc can be used.

 I think we might run into a problem here if the charging profile is split over multiple messages. (`tbc` is set `true`) in the message. It will overwrite the existing db entry. 
A solution direction split on if `tbc` in the db persisted object  is true and then merge/extend/stack incoming with saved in DB and if the income one has `tbc` false, be sure to persist this.

lydiazcheng: > Take a look at K09 the `Remark`:
> 
> > The charging profiles report can be split over multiple ReportChargingProfilesRequest messages, this can be because charging profiles for different charging sources need to be reported, or because there is just to much data for one message. To indicate that more reports will follow the flag tbc can be used.
> 
> I think we might run into a problem here if the charging profile is split over multiple messages. (`tbc` is set `true`) in the message. It will overwrite the existing db entry. A solution direction split on if `tbc` in the db persisted object is true and then merge/extend/stack incoming with saved in DB and if the income one has `tbc` false, be sure to persist this.

I might have a different understanding on the remark. The charger stores a list of charging profiles. A single charging profile object will not be split.  But the size of charging profile list can be too large to send them all in a single ReportChargingProfilesRequest or charging profiles are from different sources. In these 2 cases, charger can split and send part of charging profiles in one ReportChargingProfilesRequest and left ones in other ReportChargingProfilesRequests.

03_Modules/SmartCharging/src/module/api.ts
lydiazcheng: Thanks a lot for the helpful review, @ChrisWeissmann ! I fixed the code based on your feedback.

03_Modules/SmartCharging/src/module/module.ts
ChrisWeissmann: Ok. I dove in again and I think I get it now. So because of this constraint: `unique: 'stationId_evseId_id',` on the database object and that we save each profile with context of its `chargingLimitSource`, we know that it won't be overridden.

ChrisWeissmann (approved):
Thanks for addressing my comments 🚀

thanaParis (approved):
Thank you for mentioning the use cases in the comments, it made it much easier to follow.

Approved!

PR #124, opened 2024-06-12
Fix/Lint Fix

lydiazcheng:
The context of this PR is that, I run `npm run lint-fix` before commit my local changes to have the correct format, but then these unrelated files are fixed as well. And I have to manually reset these unrelated changes. To get rid of the manual effort in the future, I create this PR. It contains only format changes.


ChrisWeissmann (approved):
Yay for less changes and more standards in the future 🤖

yuvenus (approved):

PR #125, opened 2024-06-12
Smart Charging Part 4: Get Composite Schedule

lydiazcheng:
## Changes
1. Implement K08 - Get Composite Schedule
2. Persist CompositeSchedule
### Test send GetCompositeScheduleRequest and receive GetCompositeScheduleResponse - Req Success
1. Prepare evse and variable ChargingScheduleChargingRateUnit 
2. Send GetCompositeScheduleRequest Screenshot 2024-06-12 at 3 58 47 PM
3. Charger receive request and send a response Screenshot 2024-06-12 at 3 59 04 PM
4. CompositeSchedule stored in db. Screenshot 2024-06-12 at 5 20 30 PM
### Test send invalid value in GetCompositeScheduleRequest - Req Failed
Screenshot 2024-06-12 at 5 07 35 PM Screenshot 2024-06-12 at 5 07 09 PM


01_Data/src/interfaces/repositories.ts
```diff
@@ -133,4 +134,5 @@ export interface IChargingProfileRepository extends CrudRepository<ChargingProfi
   createOrUpdateChargingProfile(chargingProfile: ChargingProfileType, stationId: string, evseId?: number, chargingLimitSource?: ChargingLimitSourceEnumType, isActive?: boolean): Promise<ChargingProfile>;
   createChargingNeeds(chargingNeeds: NotifyEVChargingNeedsRequest, stationId: string): Promise<ChargingNeeds>;
   findChargingNeedsByEvseDBIdAndTransactionDBId(evseDBId: number, transactionDataBaseId: number): Promise<ChargingNeeds | undefined>;
+  createCompositeSchedule(compositeSchedule: CompositeScheduleType, stationId: string): Promise<CompositeSchedule>;
```
ChrisWeissmann, line 137: Out of interest, why did we add it to the ChargingProfileRepository interface? The implementation doesn't really use that repository at all. Or are we structuring it towards our modules?

03_Modules/SmartCharging/src/module/api.ts
```diff
@@ -547,12 +527,40 @@ export class SmartChargingModuleApi
...
+  async getCompositeSchedule(
     identifier: string,
     tenantId: string,
     request: GetCompositeScheduleRequest,
     callbackUrl?: string,
   ): Promise<IMessageConfirmation> {
+    // OCPP 2.0.1 Part 2 K08.FR.05
+    const evse =
+      await this._module.deviceModelRepository.findEvseByIdAndConnectorId(
+        request.evseId,
```
ChrisWeissmann, line 539: If the incoming request has the EVSEId 0, for getting the composite schedule of the entire station/grid connection as described in `K08.FR.03` this will cause an issue right?

lydiazcheng: Nice catch! Thank you!

01_Data/src/interfaces/repositories.ts
lydiazcheng: > Or are we structuring it towards our modules?

Yes, but not exactly toward the modules. According to the OCPP, CompositeSchedule is calculated based on ChargingProfiles. They belongs to the same topic. So to keep the code clean, I put them inside the same data model folder and use `ChargingProfileRepository` for CompositeSchedule as well.

ChrisWeissmann (approved):
Looks good to me 🚀

thanaParis (approved):
Well done!

Approved.

PR #127, opened 2024-06-13
OCPI Commands Module

langiole:
Minor changes made to core to support OCPI commands module.

Includes bug fix, new query, and changes to local config.

OCPI REPO PR: https://github.com/citrineos/citrineos-ocpi/pull/4


01_Data/src/layers/sequelize/repository/TransactionEvent.ts
```diff
@@ -193,4 +193,8 @@ export class SequelizeTransactionEventRepository extends SequelizeRepository<Tra
       })
       .then((row) => row as MeterValue[]);
   }
+
+  findByTransactionId(transactionId: string): Promise<Transaction | undefined> {
+    return this.transaction.readOnlyOneByQuery({ where: { transactionId } });
```
lydiazcheng, line 198: We need both transactionId and stationId to find only one transaction since in OCPP 2.0.1, transactionId is generated by each charging station. To find a station id, I think we can use session id and search in session and location when these 2 modules are ready. 🤔

Server/src/config/envs/local.ts
```diff
@@ -121,8 +121,8 @@ export function createLocalConfig() {
       },
     },
     logLevel: 2, // debug
-    maxCallLengthSeconds: 5,
-    maxCachingSeconds: 10,
+    maxCallLengthSeconds: 30,
+    maxCachingSeconds: 30,
```
lydiazcheng, line 125: Are these changes temp to testing?

langiole: I was thinking these be permanent change.

It seems useful to have longer times so that we have enough time to manually send a response.

But maybe we can have input from the team.

01_Data/src/layers/sequelize/repository/TransactionEvent.ts
langiole: Gotcha. I see in OCPP spec that it says this ID only needs to be unique for the charging station.

Server/src/config/envs/local.ts
lydiazcheng: Got it. Makes sense.

thanaParis:
Where should this PR be pointed? there's no longer a pr for feature/ocpi-restructure-v2 as far as I can tell
@langiole @elliot-sabitov

langiole:
Closing since I believe this is an outdated PR.

PR #128, opened 2024-06-17
ensuring device model persistence 

thanaParis:
making sure device model updates are persisted in NotifyEvent and StatusNotification, and NotifyMonitoringReport.

### DIFF
--- a/03_Modules/Transactions/src/module/module.ts
+++ b/03_Modules/Transactions/src/module/module.ts
@@ -24,6 +24,7 @@ import {
   MeterValuesRequest,
   MeterValuesResponse,
   ReadingContextEnumType,
+  ReportDataType,
   SampledValueType,
   StatusNotificationRequest,
   StatusNotificationResponse,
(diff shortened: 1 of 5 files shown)


03_Modules/Transactions/src/module/module.ts
```diff
@@ -401,12 +402,40 @@ export class TransactionsModule extends AbstractModule {
...
+            id: statusNotificationRequest.evseId,
+            connectorId: statusNotificationRequest.connectorId,
+          },
+        },
+        {
+          name: 'AvailabilityState',
+        },
+      );
+    if (!component || !variable) {
+      throw new Error("Missing component or variable for status notification.");
```
yuvenus, line 427: does this imply that the evse should have already sent an availability state for the connector at start up, thereby guaranteeing that the availability state should exist by the time the next status notification is sent?

i'm wondering since a few lines below the `createOrUpdateDeviceModelByStationId` is being used, so wouldn't we want it to create the availability state if it didn't exist already?

yuvenus (approved):
looks good to me! just one clarifying question 👍

lydiazcheng (approved):

03_Modules/Transactions/src/module/module.ts
thanaParis: There is a bit of a problem... 
Each component is uniquely identified by the combination of name + instance, which means the relation to EvseType is informative, not descriptive--i.e. you can't relate the same component to two different EvseTypes in order to make two Connector components. In order to be distinct, one of both of the Connector components need to have different instance values.

This is where the problem comes in: OCPP 2.0.1 is prescriptive about the component _name_ but not about the component _instance_. It leaves the specifics of the instances to the implementer. This means we can't unilaterally create the Connector for the charging station because we'd need to make up an instance or assume instance is null (often incorrectly since it's normal for stations to have more than one connector).
Therefore we're forced to throw this error, which I agree is suboptimal. I think this is the sort of thing we'd ideally bring to the OCA working group for consideration because for proper backwards compatibility, in order to support bootstrapping 1.6J implementations towards 2.0.1, we'd want to be able to handle this situation without an error, which requires the protocol to be prescriptive about the instance values, at least for the Evse & Connector components.

yuvenus: got it, makes sense to me! unfortunate that that's the case but it is what it is 😅

PR #132, opened 2024-06-18
Adjust nodemon to rebuild

elliot-sabitov:
feat: adjusting nodemon to trigger build and adjusting build command to use tsc instead of compile so that only modules that have file changes can be rebuilt quickly during development.


ChrisWeissmann (approved):

langiole:
This needs to go in ASAP!! 

I really want this feature in the code 😅

elliot-sabitov:
FYI, I think I have found a way where we can significantly reduce the time that it takes to go from npm run start-unix to Citrine Core actually running with OCPI. After some digging, I see that we are using nodemon with ts-node to run locally. Unfortunately, as a result, just triggering the exec command that we have:
node --inspect=0.0.0.0:9229 --nolazy -r ts-node/register
and waiting for it to hit a console log as the first line in /Server/index.ts, would result in like a 10-20 second wait. Following that, the application would also take another 5-15 seconds.

I tried adjusting to instead run the application via just node.
"exec": "npm run build --prefix ../ && node dist/index"
and as a result, the total time it takes for Citrine (w OCPI) to start goes down from like 20-30 seconds to just a few seconds. I think this way is most performant for our development because we are already using tsc to build the TS code we call npm run build which is already very performant as it only rebuilds modules that have had changes. The result of the npm run build / tsc command is the dist  folder that is generated containing the JS code that was transpiled from TS, so we can simply just run node dist/index to then start the application.

What is important to note with this change is that when running Citrine for local development, we would now be running JS and to have breakpoints working nicely, we would have to make sure that we have the sourcemaps and types (d.ts) files generated correctly, but that should be doable.

I am thinking about making a PR for this change but wanted to gather feedback first to make sure that no one had any concerns. In my past experience, using tsc in combination with running the built JS code as described above, has worked fairly well…

It is also worth noting that your computer performance will most likely improve with using what is described above because ts-node has to load stuff into memory, while running build JS should have no additional overhead.

All of the above is my understanding, but please fact check in case I am mistaken!

danielyoo227 (approved):

langiole:
@elliot-sabitov status of this PR?

elliot-sabitov:
feat: adjusting run start and nodemon to no longer use ts-node but instead use node to directly run the built TS files which speeds up the time that it takes the server to (re)start
feat: ensuring that assets are copied when calling npm run build otherwise logo and certs are not present which results in an error

ChrisWeissmann:
@elliot-sabitov If we decide to take out ts-node, we should do it all around, e.g. from the package.json and it lives on in the README. :)

Server/tsconfig.json
```diff
@@ -1,6 +1,6 @@
 {
   "extends": "../tsconfig.build.json",
-  "include": ["src/**/*.ts", "src/**/*.json"],
+  "include": ["src/**/*.ts", "src/**/*.json", "src/assets"],
```
langiole, line 3: Potentially don't need this additional include

PR #134, opened 2024-06-20
Feature/ocpi updates

elliot-sabitov:
Changes that are needed to core to align with new OCPI implementation


Server/src/index.ts
```diff
@@ -312,17 +312,179 @@ export class CitrineOSServer {
...
-      this.getModuleConfig(EventGroup.EVDriver),
-      this.getModuleConfig(EventGroup.Monitoring),
-      this.getModuleConfig(EventGroup.Reporting),
-      this.getModuleConfig(EventGroup.SmartCharging),
-      this.getModuleConfig(EventGroup.Transactions),
-    ].forEach((moduleConfig) => {
-      this.initModule(moduleConfig);
-    });
+
+    if (this._config.modules.certificates) {
```
langiole, line 316: Not sure I really see these changes below as an enhancement.

Seems like we just expanded the for loop. Would've preferred to just keep OCPI related changes instead of doing any other enhancements.

But if I am in the minority, I am fine to proceed with these changes.

00_Base/src/interfaces/messages/index.ts
```diff
@@ -32,6 +32,15 @@ export enum EventGroup {
   Reporting = 'reporting',
   SmartCharging = 'smartcharging',
   Transactions = 'transactions',
+  Cdrs = 'cdrs',
```
langiole, line 35: We might want to prefix these with "Ocpi".

E.g. OcpiCdrs = 'OcpiCdrs'

00_Base/src/interfaces/api/exceptions/unauthorized.exception.ts
langiole, line 1: Was this added because ocpi uses it?

If so, why not keep in ocpi repo?

langiole:
@elliot-sabitov some conflicts that need to be resolved.

I would also ask if we can please merge this in ASAP so as to not keep adding to it.

Server/src/index.ts
elliot-sabitov: yeah we expanded the loop because the different module classes need different repository classes passed in during initialization, and previously we werent passing the repos in and so new repos were being created. We could have still done it with the loop I suppose but we wanted to move quickly and this is more explicit I suppose, but I think if we move towards DI then much of this would be adjusted most likely

00_Base/src/interfaces/messages/index.ts
elliot-sabitov: perhaps that can be done in a followup PR?

00_Base/src/config/types.ts
```diff
@@ -459,6 +464,10 @@ export const systemConfigSchema = z
     logLevel: z.number().min(0).max(6),
     maxCallLengthSeconds: z.number().int().positive(),
     maxCachingSeconds: z.number().int().positive(),
+      ocpiServer: z.object({
```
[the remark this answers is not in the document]
elliot-sabitov: will run lint fix

00_Base/src/interfaces/api/exceptions/unauthorized.exception.ts
elliot-sabitov: I think it seemed sound to keep an UnauthorizedException in the core base, but I suppose if that will never be used elsewhere then it makes sense to move this to OCPI repo...

elliot-sabitov:
I fixed the conflicts and lint issues. The github workflows are failing because the OCPI npm packages are not yet published to NPM

langiole (approved):

Server/src/index.ts
```diff
@@ -313,18 +345,143 @@ export class CitrineOSServer {
...
+    }
+  }
+
+  private startOcpiServer(host: string, port: number) {
+    const ocpiServer = new OcpiServer(
+      this._config as OcpiServerConfig,
+      this._cache,
+      this._logger,
+      this.getOcpiModuleConfig(),
+      this._repositoryStore,
```
langiole, line 482: Seemingly this additional argument requires changes in ocpi repo?

Can we please have a PR up for this change in ocpi repo? Not seeing one at the moment.

elliot-sabitov: @langiole this is already in `main`: 
```
export class OcpiServer extends KoaServer {
  koa!: Koa;
  readonly serverConfig: OcpiServerConfig;
  readonly cache: ICache;
  readonly logger: Logger<ILogObj>;
  readonly ocpiSequelizeInstance: OcpiSequelizeInstance;
  readonly modules: OcpiModule[] = [];
  readonly repositoryStore: RepositoryStore;
  
  constructor(
    serverConfig: OcpiServerConfig,
    cache: ICache,
    logger: Logger<ILogObj>,
    modulesConfig: OcpiModuleConfig[],
    repositoryStore: RepositoryStore,
  ) {
  ...
  ```
  I think you are talking about these changes right?

File: 01_Data/src/layers/sequelize/repository/Base.ts [hunk 5]
```diff
@@ -19,14 +19,30 @@ export class SequelizeRepository<T extends Model<any, any>> extends CrudReposito
...
-  async readByKey(key: string): Promise<T | undefined> {
+  async readByKey(key: string | number): Promise<T | undefined> {
     return await this.s.models[this.namespace].findByPk(key).then((row) => row as T);
   }
 
   async readAllByQuery(query: object): Promise<T[]> {
     return await this.s.models[this.namespace].findAll(query as FindOptions<any>).then((row) => row as T[]);
   }
 
+  async readAllBySqlString(sqlString: string): Promise<object[]> {
```
thanaParis, line 30: Is this something we should consider adding to CrudRepository as well? Or will the syntax be too sequelize-specific?

PR #135, opened 2024-06-24
Feature/persisting timestamps

thanaParis:
Adding timestamp to IMessageContext.
Adding timestamp to VariableAttribute collection.

Timestamps for VariableAttribute that more closely reflect when the modification was made as opposed to when the db entity was updated are needed. This PR approaches that goal from two angles: using the 'generatedAt' timestamp in OCPP messages where available, and using a new timestamp on the IMessageContext in most other places (some Message API calls just use `new Date()`). The IMessageContext.timestamp field will have other uses for measuring CitrineOS performance and message tracking in the future, as it is created when the message is created.

In addition to the above, there is an improvement to overall message routing. MessageOrigin was previously inconsistent in usage, applying in some cases to the sender of the Call (so, Heartbeat-related messages would all have origin 'Charging Station') and in some cases to the sender of the message (so HeartbeatRequest has origin 'Charging Station' and HeartbeatResponse has origin 'Central System'). We've found the first behavior is just confusing, so now the latter will always apply. Additionally, 'CentralSystem' is OCPP 1.6J terminology, so it's been replaced.


yuvenus (approved):
nice, approved!

01_Data/src/layers/sequelize/model/DeviceModel/VariableAttribute.ts
```diff
@@ -78,6 +78,14 @@ export class VariableAttribute extends Model implements VariableAttributeType {
   })
   declare constant?: boolean;
 
+  @Column({
+    type: DataType.DATE,
+    get() {
+      return this.getDataValue('timestamp').toISOString();
+    },
+  })
+  declare timestamp: string;
```
ChrisWeissmann, line 87: Nitpick: The naming here could be not clear enough. Since this extent sequelize Model there are already 2 timestamps automatically added (createdAt and updatedAt). With those 2 fields in mind, I would get confused if I were new a looking at the DB. Wondering what the generic timestamp represents.

ChrisWeissmann (commented):
I like the idea! It gives us more transparency on if certain messages get through much better!  🚀 
Left a comment on the naming.

01_Data/src/layers/sequelize/model/DeviceModel/VariableAttribute.ts
thanaParis: fixed!

ChrisWeissmann (approved):
Looks good to me 🚀

PR #137, opened 2024-06-26
ensuring callErrors are processed when able, and do not crash citrine

thanaParis:
CallError handling has not worked as expected. This PR aims to fix some of the problems seen in testing which prevented CallErrors from being sent, caused exceptions that should have been CallErrors to crash citrine, or sent CallErrors when they should not have been sent.


langiole (approved):

lydiazcheng (approved):
The change looks good!

PR #139, opened 2024-06-30
fixing status notification handling

thanaParis:
- fixed issue where 'undefined' instance on component broke status notification handling
- made incomplete device model (missing connector component/availability state connector variable) no longer break statusNotification handling but instead emit warn log
- added new StatusNotification collection
- created many-to-one relationship between StatusNotification and ChargingStation
- now we save StatusNotification & relation when handling message

### DIFF
--- a/01_Data/src/layers/sequelize/repository/DeviceModel.ts
+++ b/01_Data/src/layers/sequelize/repository/DeviceModel.ts
@@ -346,10 +346,10 @@ export class SequelizeDeviceModelRepository extends SequelizeRepository<Variable
 
   async findComponentAndVariable(componentType: ComponentType, variableType: VariableType): Promise<[Component | undefined, Variable | undefined]> {
     const component = await this.component.readOnlyOneByQuery({
-      where: { name: componentType.name, instance: componentType.instance ? componentType.instance : undefined },
+      where: { name: componentType.name, instance: componentType.instance ? componentType.instance : null },
     });
     const variable = await this.variable.readOnlyOneByQuery({
-      where: { name: variableType.name, instance: variableType.instance ? variableType.instance : undefined },
+      where: { name: variableType.name, instance: variableType.instance ? variableType.instance : null },
     });
     if (variable) {
       const variableCharacteristics = await this.variableCharacteristics.readOnlyOneByQuery({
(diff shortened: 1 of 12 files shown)


01_Data/src/layers/sequelize/repository/DeviceModel.ts
```diff
@@ -346,10 +346,10 @@ export class SequelizeDeviceModelRepository extends SequelizeRepository<Variable
 
   async findComponentAndVariable(componentType: ComponentType, variableType: VariableType): Promise<[Component | undefined, Variable | undefined]> {
     const component = await this.component.readOnlyOneByQuery({
```
yuvenus, line 348: in addition the `undefined` issue fixed here, we noticed `readOnlyOneByQuery` would fail if it found two component entries, which happened because i had an EVSE with two connectors, and therefore each component entry had a different `evseDatabaseId` 🤔 

i see down below in the transaction module that problem is solved in there -- would a similar logic need to be applied to additionally search by evseId/connectorId (if that's even possible here) to prevent this issue or is that a concern for a separate PR?

Server/src/config/envs/docker.ts
```diff
@@ -53,7 +53,7 @@ export function createDockerConfig() {
         username: 'citrine',
         password: 'citrine',
         storage: '',
-        sync: false,
+        sync: true,
```
yuvenus, line 56: is this a permanent thing that should happen or just temporary for testing? if the latter, can this be reverted?

01_Data/src/layers/sequelize/repository/DeviceModel.ts
thanaParis: > which happened because i had an EVSE with two connectors, and therefore each component entry had a different evseDatabaseId

This is a protocol breach, so we don't need to worry about that case. Two components with the same name and instance cannot exist.

This method does not support searching for a component by name only to find different instances. Undefined is treated as null, so there is no argument that can be given which would do that. This is why I had to resort to a query directly on an instance of CrudRepository<Component> to get what we need in the StatusNotification handler.

yuvenus (approved):

PR #142, opened 2024-07-02
Tariffs module

falanadamian:
This PR enhances the Tariffs model and updates the Tariffs API accordingly.


00_Base/src/util/model/CountryCode.ts
```diff
@@ -0,0 +1,267 @@
...
+    return (COUNTRY_CODES as readonly string[]).includes(value); // TODO: optimize
+}
+
+export function countryCode(value: string): CountryCode {
+    assert(isCountryCode(value), `Invalid country code: ${value}`);
+    return value;
+}
+
+// ISO 3166-1 alpha-2 country codes
+export const COUNTRY_CODES = [
```
falanadamian, line 15: Should we limit the country codes to only those that we plan to support for now?

00_Base/src/util/model/Currency.ts
```diff
@@ -0,0 +1,186 @@
...
+    return (CURRENCIES as readonly string[]).includes(value);  // TODO: optimize
+}
+
+export function currency(value: string): Currency {
+    assert(isCurrency(value), `Invalid currency: ${value}`);
+    return value;
+}
+
+// ISO 4217 currencies
+const CURRENCIES = [
```
falanadamian, line 15: Should we limit the currencies to only those that we plan to support for now?

00_Base/src/util/model/Language.ts
```diff
@@ -0,0 +1,200 @@
...
+    return (LANGUAGES as readonly string[]).includes(value); // TODO: optimize
+}
+
+export function language(value: string): Language {
+    assert(isLanguage(value), `Invalid language: ${value}`);
+    return value;
+}
+
+// ISO 639-1 language codes
+const LANGUAGES = [
```
falanadamian, line 15: Should we limit the languages to only those that we plan to support for now?

01_Data/src/layers/sequelize/model/Tariff/TariffElement.ts
```diff
@@ -0,0 +1,175 @@
...
+    allowNull: false,
+    validate: {
+      notNull: true,
+      atLeastOne: (value: PriceComponent[]) => assert(notEmpty(value), `priceComponents cannot be empty`),
+      hasNoDuplicates: (value: PriceComponent[]) => assert(!hasDuplicates(value, 'type'), 'priceComponents cannot contain duplicates'),
+    }
+  })
+  declare priceComponents: PriceComponent[];
+
+  @Column(DataType.VIRTUAL)
```
falanadamian, line 54: I think it's preferable to maintain tariff restrictions within the tariff elements table instead of creating a separate table for them. Individual restrictions are private within this class, and a virtual data type was used to present restrictions as a single object from the TariffElement.

01_Data/src/layers/sequelize/model/Tariff/TariffElement.ts
```diff
@@ -0,0 +1,175 @@
...
+
+  @DateOnlyColumn()
+  private declare startDate?: DateOnly;
+
+  @DateOnlyColumn()
+  private declare endDate?: DateOnly;
+
+  @Column({
+    type: DataType.DECIMAL(19, 4),
+    get(this: Tariff) {
```
falanadamian, line 105: It seems Sequelize treats DECIMAL type as a string during data retrieval, hence this getter mapping. (I used it for every other DECIMAL column I added)

falanadamian: Also, I was considering whether we should store and manage money in minor currency units to avoid floating-point errors with JS numbers? Instead of developing our own wrapper class, using a library like Dinero.js could be an option?
I'm also curious if you think it's worthwhile to create similar wrappers for energy (storing and calculating in Watthour instead of Kilowatthour), power, and electrical current. For now, I keep it as plain numbers.

01_Data/src/layers/sequelize/model/Tariff/Tariffs.ts
```diff
@@ -1,32 +1,295 @@
...
+    type: DataType.STRING(36),
+    validate: {
+      isTariffId: (value: string) => assert(isTariffId(value), `Invalid id: ${value}`),
+    }
+  })
+  declare id: TariffId;
 
   @Index
+  @Column(DataType.STRING)
+  declare stationId: string; // TODO: ??
```
falanadamian, line 65: Not sure about stationId and how to handle it; it seems like we might want to refer to the connector instead?

01_Data/src/layers/sequelize/model/Tariff/Tariffs.ts
```diff
@@ -1,32 +1,295 @@
...
   static readonly MODEL_NAME: string = Namespace.Tariff;
 
-  /**
-   * Fields
-   */
+  public static newInstance(data: TariffData): Tariff {
+    return Tariff.build({...data}, {include: [TariffElement]});
+  }
+
+  @PrimaryKey
```
falanadamian, line 54: I understand we want a composite primary key of (tariffId, countryCode, partyId)?

01_Data/src/layers/sequelize/model/Tariff/Tariffs.ts
```diff
@@ -1,32 +1,295 @@
...
+
+  @Column({
+    type: DataType.DECIMAL(19, 4),
+    get(this: Tariff) {
+      return parseFloat(this.getDataValue('minPriceInclVat'));
+    }
+  })
+  private declare minPriceInclVat?: number;
+
+  @Column(DataType.VIRTUAL)
```
falanadamian, line 133: Similarly to restrictions for TariffElement, I believe it's better to maintain min price and max price as separate columns in the Tariff table.

01_Data/src/layers/sequelize/repository/Tariff.ts
```diff
@@ -22,30 +30,36 @@ export class SequelizeTariffRepository extends SequelizeRepository<Tariff> imple
...
+      if (!savedTariff) {
+        const createdTariff = await tariff.save({ transaction });
+        this.emit('created', [createdTariff]);
+        return createdTariff;
       }
-      const createdTariff = await tariff.save({ transaction });
-      this.emit('created', [createdTariff]);
-      return createdTariff;
+
+      // TODO: optimize
```
falanadamian, line 47: Keeping it simple for now; we can consider optimizations later, if that works for everyone?

01_Data/src/layers/sequelize/repository/Tariff.ts
```diff
@@ -22,30 +30,36 @@ export class SequelizeTariffRepository extends SequelizeRepository<Tariff> imple
...
+      await TariffElement.destroy({where: {tariffId: savedTariff.id}, transaction});
+      await TariffElement.bulkCreate(tariff.elementsData.map(element => ({...element, tariffId: savedTariff.id}), {transaction}));
+      const updatedTariff = await savedTariff.set(tariff.data).save({transaction});
+
+      this.emit('updated', [updatedTariff]);
+      return updatedTariff;
     });
   }
 
   async readAllByQuerystring(query: TariffQueryString): Promise<Tariff[]> {
```
falanadamian, line 57: Should I extend filtering with countryCode and partyId?

01_Data/src/layers/sequelize/repository/Tariff.ts
```diff
@@ -22,30 +30,36 @@ export class SequelizeTariffRepository extends SequelizeRepository<Tariff> imple
...
     return super.readAllByQuery({
       where: {
-        ...(query.stationId ? { stationId: query.stationId } : {}),
-        ...(query.unit ? { unit: query.unit } : {}),
-        ...(query.id ? { id: query.id } : {}),
+        ...(query.stationId && { stationId: query.stationId }),
+        // ...(query.unit && { unit: query.unit }), // TODO: tariff dimension type?
```
falanadamian, line 61: Not sure how to handle the unit here; should I filter by tariff dimension type? Is it needed?

01_Data/src/layers/sequelize/types/DateOnlyColumn.ts
```diff
@@ -0,0 +1,18 @@
+import {Column, DataType, Model} from "sequelize-typescript";
+import {DateOnly, safelyMap} from "@citrineos/base";
+import {TariffElement} from "../model/Tariff/TariffElement";
+
+// TODO: custom type would be preferable
```
falanadamian, line 5: I tried registering a custom type in Sequelize for it but couldn't get it to work. I think a custom type would be a more elegant solution. Perhaps we can revisit it when we have more time?

03_Modules/Transactions/src/module/api.ts
```diff
@@ -115,13 +118,72 @@ export class TransactionsModuleApi
...
-      Querystring: CreateOrUpdateTariffQueryString;
+      Body: any;
     }>,
   ): Promise<Tariff> {
-    return this._module.tariffRepository.createOrUpdateTariff(request.body);
+    const tariff = this.buildTariff(plainToInstance(UpsertTariffRequest, request.body));
+    return await this._module.tariffRepository.upsertTariff(tariff);
+  }
+
+  // TODO: move to service layer
```
falanadamian, line 130: Will move it to service layer once we have DI in place if that works?

03_Modules/Transactions/src/module/module.ts
```diff
@@ -513,7 +513,9 @@ export class TransactionsModule extends AbstractModule {
         { totalKwh: totalKwh },
         { where: { id: transactionDbId }, returning: false },
       );
-      totalCost = this._roundCost(totalKwh * tariff.price);
+      // TODO: fix
+      let price = tariff.elements[0].priceComponents.find(component => component.isEnergy())!.price;
```
falanadamian, line 517: Is it okay to fetch the price from the energy component of the first element? I'm uncertain about how to proceed at this point.

PR #144, opened 2024-07-03
fix: IdTokenInfo Id being overwritten

ChrisWeissmann:
fix: IdTokenInfo is being set as null for updates causing them to now be saved correctly. Set to undefined to avoid being overwritten.


thanaParis (approved):
Sweet fix. Approved

PR #153, opened 2024-07-08
Enhance password validation

falanadamian:
This PR adds validation to ensure that a password is present in the request if setOnCharger is true. 
This prevents updating the device model with an autogenerated password without simultaneously updating the charger.

PR with quickstart updates:
https://github.com/citrineos/citrineos.github.io/pull/26


thanaParis (approved):
Approved!

PR #157, opened 2024-07-09
Map totalKwh during create / update on Transaction

danielyoo227:
Changing the logic for the totalKwh mapping to be done within one transaction update so that only one event is triggered for a finalized transaction for it to be processed into a CDR.

### DIFF
--- a/01_Data/src/layers/sequelize/repository/TransactionEvent.ts
+++ b/01_Data/src/layers/sequelize/repository/TransactionEvent.ts
@@ -7,11 +7,13 @@ import {
   type ChargingStateEnumType,
   CrudRepository,
   type EVSEType,
+  IdTokenEnumType,
   type IdTokenType,
-  MeasurandEnumType,
-  MeterValueType,
-  ReadingContextEnumType,
-  SampledValueType, SystemConfig, TransactionEventEnumType, type TransactionEventRequest, IdTokenEnumType } from '@citrineos/base';
+  MeterValueUtils,
+  SystemConfig,
+  TransactionEventEnumType,
+  type TransactionEventRequest
+} from '@citrineos/base';
 import { type ITransactionEventRepository } from '../../../interfaces';
 import { MeterValue, Transaction, TransactionEvent } from '../model/TransactionEvent';
 import { SequelizeRepository } from './Base';
@@ -122,7 +124,7 @@ export class SequelizeTransactionEventRepository extends SequelizeRepository<Tra
         }
       }
 
-      event = await event.save({ transaction: sequelizeTransaction });
+      event = await event.save({transaction: sequelizeTransaction});
 
       if (value.meterValue && value.meterValue.length > 0) {
         await Promise.all(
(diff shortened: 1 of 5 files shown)


01_Data/src/layers/sequelize/repository/TransactionEvent.ts
```diff
@@ -143,11 +155,74 @@ export class SequelizeTransactionEventRepository extends SequelizeRepository<Tra
...
+      });
+
+    // sort the map based on timestamps
+    valuesMap = new Map(
+      [...valuesMap.entries()].sort((v1, v2) => v1[0] - v2[0]),
+    );
+    const sortedValues = Array.from(valuesMap.values());
+
+    let totalKwh = 0;
+    for (let i = 1; i < sortedValues.length; i++) {
```
thanaParis, line 214: I would do it this way:
totalKwh = sortedValues[sortedValues.length-1] - sortedValues[0]
same answer, easier to understand and more performant

01_Data/src/layers/sequelize/repository/TransactionEvent.ts
```diff
@@ -143,11 +155,74 @@ export class SequelizeTransactionEventRepository extends SequelizeRepository<Tra
...
+      {totalKwh}
+    );
+
+    return totalKwh;
+  }
+
+  private getTotalKwh(meterValues: MeterValueType[]): number {
+    const contexts: ReadingContextEnumType[] = [
+      ReadingContextEnumType.Transaction_Begin,
+      ReadingContextEnumType.Sample_Periodic,
```
thanaParis, line 171: technically, you don't need Sample_Periodic, only begin and end. will make the rest of this easier.

01_Data/src/layers/sequelize/repository/TransactionEvent.ts
```diff
@@ -143,11 +155,74 @@ export class SequelizeTransactionEventRepository extends SequelizeRepository<Tra
...
+      ReadingContextEnumType.Transaction_End,
+    ];
+
+    let valuesMap = new Map();
+
+    meterValues
+      .filter(
+        (meterValue) =>
+          meterValue.sampledValue[0].context &&
```
thanaParis, line 180: when absent, context is Sample.Periodic

From spec, definition of context field: 
> Type of detail value: start, end or sample.
Default = "Sample.Periodic"

so you should actually make it:
> !meterValue.sampledValue[0].context ||

01_Data/src/layers/sequelize/repository/TransactionEvent.ts
```diff
@@ -143,11 +155,74 @@ export class SequelizeTransactionEventRepository extends SequelizeRepository<Tra
...
+          contexts.indexOf(meterValue.sampledValue[0].context) !== -1,
+      )
+      .forEach((meterValue) => {
+        const sampledValues = meterValue.sampledValue as SampledValueType[];
+        const overallValue = sampledValues.find(
+          (sampledValue) =>
+            sampledValue.phase === undefined &&
```
thanaParis, line 187: this is a dangerous way to compare, it could come in as null due to Sequelize's handling of absent fields
I'd stick to a truthy-falsy comparison: 
> !sampledValue.phase &&

01_Data/src/layers/sequelize/repository/TransactionEvent.ts
danielyoo227: Thanks! Will resolve all 🙌

01_Data/src/layers/sequelize/repository/TransactionEvent.ts
```diff
@@ -143,11 +155,74 @@ export class SequelizeTransactionEventRepository extends SequelizeRepository<Tra
...
+            MeasurandEnumType.Energy_Active_Import_Register,
+        );
+        if (
+          overallValue &&
+          overallValue.unitOfMeasure?.unit?.toUpperCase() === 'KWH'
+        ) {
+          valuesMap.set(Date.parse(meterValue.timestamp), overallValue.value);
+        } else if (
+          overallValue &&
+          overallValue.unitOfMeasure?.unit?.toUpperCase() === 'WH'
```
thanaParis, line 198: spec:
>  Optional. Unit of the value. Default = "Wh" if the (default)
measurand is an "Energy" type.

```
let powerOfTen = overallValue.unitOfMeasure?.multiplier ?? 0;
const unit = overallValue.unitOfMeasure?.unit;
switch (unit) {
  case 'kWh':
    break;
  case 'Wh':
  case undefined:
    powerOfTen -= 3
  default:
    throw new Error("Unknown unit for Energy.Active.Import.Register: " + unit);
}
overallValue.value = overallValue.value * 10 ** powerOfTen;
```

thanaParis (approved):
Approved

PR #163, opened 2024-07-10
Feature/everest helper

elliot-sabitov:
feat: creating convenience `npm run start-everest` helper that will perform docker compose up using the newly created docker-compose and Dockerfile in the /Server/everest folder. Dockerfile mimics ghcr.io/everest/everest-demo/manager image but also installs sqlite and performs a SQL query to set the CSMS URL. The Dockerfile supports a `EVEREST_TARGET_URL` build argument that can be used to specify the target that everest should point to. The docker compose file sets this argument as well and is used by npm to start the everest server

### DIFF
--- a/Server/everest/Dockerfile
+++ b/Server/everest/Dockerfile
@@ -20,5 +20,9 @@ RUN sqlite3 /ext/source/build/dist/share/everest/modules/OCPP201/device_model_st
 
 RUN rm /ext/source/build/dist/etc/everest/certs/ca/v2g/*.*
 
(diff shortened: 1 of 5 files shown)


elliot-sabitov:
Moved to WIP because noticed an issue

elliot-sabitov:
Also enhancing helper by adding http-server to serve Everest OCPP logs on port 8888 by creating a start.sh script that turned out to be the simplest approach to ensure that the sil script and the http-server are triggered simultaneously

PR #165, opened 2024-07-15
correcting origin for call errors during message handling

thanaParis:
When errors are thrown during Call handling, CallError should be emitted with correct origin (CSMS)


lydiazcheng (approved):
The fix looks good!

PR #166, opened 2024-07-15
Fixing issue where Transaction is never set to inactive

thanaParis:


01_Data/src/layers/sequelize/repository/TransactionEvent.ts
```diff
@@ -79,18 +79,21 @@ export class SequelizeTransactionEventRepository extends SequelizeRepository<Tra
 
       const updatedTransaction = Transaction.build({
         stationId,
-        isActive: value.eventType !== TransactionEventEnumType.Ended,
```
lydiazcheng, line 82: Hi Thana, why this way doesn't work?

lydiazcheng (approved):
Approved!

01_Data/src/layers/sequelize/repository/TransactionEvent.ts
thanaParis: the default value on the column overwrote this logic. Modifying the column was enough to fix the immediate issue, but I also wanted to make sure that in the case where messages are received out of order the Transaction stays inactive no matter what, so I refactored this logic as well.

PR #168, opened 2024-07-17
Feature/authorization interface

thanaParis:
adding an authorization interface for additional authorization steps


02_Util/src/authorization/authorization.ts
```diff
@@ -0,0 +1,21 @@
...
+     * The Partial IdTokenInfoType is used to update the IdTokenInfo in the AuthorizeResponse or TransactionEventResponse.
+     * If IdTokenInfoType.status is not Accepted, the authorization process will stop and the rejection put into the response.
+     * The order of Authorizers can lead to different outputs; instantiate them in the order they should be called.
+     * 
+     * @param {Authorization} authorization The authorization object associated with the idToken in the request. No modifications should be made to this object.
+     * @param {IMessageContext} context
+     * 
+     * @returns {Promise<Partial<IdTokenInfoType>>} Updates to the IdTokenInfo
+    **/
+    authorize(authorization: Authorization, context: IMessageContext, ): Promise<Partial<IdTokenInfoType>>;
```
langiole, line 20:
```suggestion
    authorize(authorization: Authorization, context: IMessageContext): Promise<Partial<IdTokenInfoType>>;
```

03_Modules/EVDriver/src/module/module.ts
```diff
@@ -392,6 +402,15 @@ export class EVDriverModule extends AbstractModule {
...
                 }
               }
+
+              for (const authorizer of this._authorizers) {
+                if (response.idTokenInfo.status !== AuthorizationStatusEnumType.Accepted) {
+                  break;
+                }
+                const result: Partial<IdTokenType> = await authorizer.authorize(authorization, context);
+                Object.assign(response.idTokenInfo, result);
+              }
```
langiole, line 412: This is just personal preference.

```suggestion
if (response.idTokenInfo.status === AuthorizationStatusEnumType.Accepted) {  
  for (const authorizer of this._authorizers) {
    const result: Partial<IdTokenType> = await authorizer.authorize(authorization, context);
    Object.assign(response.idTokenInfo, result);
  }
}
```

langiole (approved):
Minor formatting feedback. Approved otherwise.

thanaParis:
> The code looks good! Just to have a better understanding, is the plan to instantiate this interface in the OCPI so that we can inject the result of token validation into OCPP? Why do we need the context info?

Yes that's the plan! The context info is needed for the stationId and tenantId

03_Modules/EVDriver/src/module/module.ts
thanaParis: This doesn't work; if there's a rejection in the for loop it no longer breaks.

03_Modules/Transactions/src/module/module.ts
```diff
@@ -493,6 +507,19 @@ export class TransactionsModule extends AbstractModule {
             // TODO: allow for a 'real time auth' type call to fetch token status.
```
langiole, line 507: Is this TODO still needed?

@thanaParis @elliot-sabitov

PR #173, opened 2024-07-19
Feature/session cdr transaction fetch refactor

danielyoo227:
Adding additional mapping required to map OCPI objects based on Core objects


01_Data/src/layers/sequelize/model/Authorization/IdToken.ts
```diff
@@ -28,4 +29,7 @@ export class IdToken extends Model implements IdTokenType {
   declare type: IdTokenEnumType;
 
   declare customData?: CustomDataType;
+
+  @HasOne(() => Authorization)
+  declare authorization: Authorization;
```
thanaParis, line 34: It is possible that an IdToken is a IdTokenInfo.groupId token. Linking it this way does not allow for this distinction. This will need to be changed.

01_Data/src/layers/sequelize/model/TransactionEvent/Transaction.ts
```diff
@@ -38,9 +40,15 @@ export class Transaction extends Model implements TransactionType {
...
 
-  @HasMany(() => TransactionEvent)
+  @HasMany(() => TransactionEvent,
+    { as: Transaction.TRANSACTION_EVENTS_ALIAS, foreignKey: 'transactionDatabaseId' })
   declare transactionEvents?: TransactionEventRequest[];
 
+  // required only for filtering, should not be used to pull transaction events
+  @HasMany(() => TransactionEvent,
+    { as: Transaction.TRANSACTION_EVENTS_FILTER_ALIAS, foreignKey: 'transactionDatabaseId' })
+  declare transactionEventsFilter?: TransactionEventRequest[];
```
thanaParis, line 50: Could you elaborate on why this field is needed? I've looked over both PRs and I have yet to understand fully...

thanaParis (dismissed):

01_Data/src/layers/sequelize/model/TransactionEvent/Transaction.ts
elliot-sabitov: not sure why it doent show in the diff, but the EVENTS_FILTER_ALIAS is being used in the `TransactionQueryBuilder`:

```
buildQuery(params: QueryParams, endedOnly?: boolean): any {
    const modelsToInclude: any[] = [
      this.createMspFilters(params.mspCountryCode, params.mspPartyId),
      this.createCpoFilters(params.cpoCountryCode, params.cpoPartyId),
      MeterValue,
      Evse,
      ...(endedOnly
        ? [
            {
              model: TransactionEvent,
              as: Transaction.TRANSACTION_EVENTS_FILTER_ALIAS,
              attributes: ['eventType'],
              where: {
                eventType: TransactionEventEnumType.Ended,
              },
            },
          ]
        : []),
    ];
```

elliot-sabitov: why do we need a separate alias though, cant we just use the same one @danielyoo227 ?

elliot-sabitov:
Dismissing Thana's review for now but will circle back. Merging the PR for now to eliminate build errors with latest 0.2.0 code. We can make additional adjustments in a follow up PR (if needed) 🙇

PR #174, opened 2024-07-19
WIP: introducing an integration between TSLog and ELK 

elliot-sabitov:
**Problem:**
It is difficult to debug the application because there are ongoing messages and setting breakpoints causes queue build ups and other issues. Sifting through the console is also challenging since there are a lot of incoming messages.

**Proposal:**
Introducing an integration between TSLog and ELK (Elasticsearch, Logstash, Kibana). Adjusted docker compose to run the 3 services and adjusted TSLog implementation in Server to add a transport to log to Logstash. Running docker compose up now enables us to navigate to `5601` and use default credentials to log in and view logs provided the index pattern for tslog-* is created.

Screenshot 2024-07-19 at 6 01 04 PM


Server/docker-compose.yml
```diff
@@ -87,7 +87,7 @@ services:
         condition: service_healthy
       amqp-broker:
         condition: service_healthy
-      directus:
```
thanaParis, line 90: Why is this replacing the directus condition? We still rely on directus being up in order to start CitrineOS. We should add the logstash condition to the list instead.

Server/src/index.ts
```diff
@@ -302,13 +303,28 @@ export class CitrineOSServer {
...
   private initLogger() {
-    return new Logger<ILogObj>({
+    const logger = new Logger<ILogObj>({
       name: 'CitrineOS Logger',
       minLevel: systemConfig.logLevel,
       hideLogPositionForProduction: systemConfig.env === 'production',
       // Disable colors for cloud deployment as some cloud logging environments such as cloudwatch can not interpret colors
       stylePrettyLogs: process.env.DEPLOYMENT_TARGET !== 'cloud',
     });
+    const logstashHost = 'logstash'; // WARNING: This will only work within docker
```
thanaParis, line 313: This seems like a serious issue, right? what happens when you start locally?

thanaParis:
I appreciate the current setup is hard to debug, but I'm wondering what specific benefits there are to this change, as it introduces multiple new points of failure. A tool like Docker desktop already lets you view all the logs and search them. You mention issues with breakpoints causing queue buildup--does this help with that in any way or is it just a nicer way of viewing the logs already visible in the docker container's console?

elliot-sabitov:
> I appreciate the current setup is hard to debug, but I'm wondering what specific benefits there are to this change, as it introduces multiple new points of failure. A tool like Docker desktop already lets you view all the logs and search them. You mention issues with breakpoints causing queue buildup--does this help with that in any way or is it just a nicer way of viewing the logs already visible in the docker container's console?

This is intended as a quick bandaid to be able to have a nicer way to look at the logs which may be helpful in certain cases where the log streams become unmanageable. The issue is that if there are lets say hundreds or thousands of lines of logs produced per x amount of time, then when you are displaying those logs on your computer, whether from a container or not, it will need to take up memory on your machine, and in extreme cases, you can have a situation where either your buffer size is too small and is dropping old logs that you still care about, or your allowable buffer size for the logs is large, and it ends up taking more memory from your machine. 

Disregarding the memory though, this is predominatly a nicety to view, filter / search logs better for local development and testing. Additionally, I was thinking maybe we can create views that we can then make sure are included so that we can launch the app and right away see some visualizations that may be helpful. For example, you can quickly see top 10 errors that occurred recently sorted by order of occurrence right away without having to look at the logs, potentially saving time and allowing us to catch issues earlier... You can probably even create charge specific visualizations where you can select a charger and quickly see only logs pertaining to that single charger, which may be more helpful in cases where we are testing swarm?

Server/docker-compose.yml
elliot-sabitov: right, I think that was a misunderstanding - now that you point it out, we need it in COS because we are creating flows right? At that moment, I was thinking that Directus is a separate layer and COS should be able to run without it being up... You are also correct, that yes if we do need Directus to run COS, then I can add that back. The `logstash` is being added below, so adding `directus` back should be all we need...

Server/src/index.ts
elliot-sabitov: then it just logs to the console.. because this is just an additional transport. You do make a good point though, this was kind of a quick intro PR to get thoughts from the team on where we want to take it - in this case, if you run locally, your local config would specify where you are running logstash or exclude it entirely.. so if you have logstash running locally, you can use `localhost` here

elliot-sabitov:
Closing PR as it is outdated but can be used as a reference

PR #175, opened 2024-07-22
Added handling for unknown tokens

langiole:
Added proper handling for unknown tokens on transaction start

### DIFF
--- a/03_Modules/Transactions/src/module/module.ts
+++ b/03_Modules/Transactions/src/module/module.ts
@@ -403,22 +403,20 @@ export class TransactionsModule extends AbstractModule {
       // old transactions should be marked inactive and an alert should be raised (this can only happen in the field with charger bugs or missed messages)
 
       // Check for ConcurrentTx
-      return await this._transactionEventRepository
-        .readAllActiveTransactionsByIdToken(transactionEvent.idToken)
-        .then((activeTransactions) => {
-          // Transaction in this TransactionEventRequest has already been saved, so there should only be 1 active transaction for idToken
-          if (activeTransactions.length > 1) {
-            const groupIdToken =
-              transactionEventResponse.idTokenInfo?.groupIdToken;
-            transactionEventResponse.idTokenInfo = {
-              status: AuthorizationStatusEnumType.ConcurrentTx,
-              groupIdToken: groupIdToken,
-              // TODO determine how/if to set personalMessage
-            };
-          }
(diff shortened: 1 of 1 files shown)


langiole:
Server was crashing on unknown tokens as shown here:
image

It seems as though an error thrown within `Promise.then` will not be caught by the try/catch all.

The fix was to adjust the code to not use `Promise.then`.

We also adjusted the code such that unknown token will result in a `TransactionEventResponse` with `idTokenStatus === Unknown` instead of a `CallError`.

03_Modules/Transactions/src/module/module.ts
```diff
@@ -412,6 +305,127 @@ export class TransactionsModule extends AbstractModule {
...
+    const transactionEventResponse: TransactionEventResponse = {
+      idTokenInfo: {
+        status: AuthorizationStatusEnumType.Unknown,
+        // TODO determine how/if to set personalMessage
+      },
+    };
+
+    if (authorizations.length !== 1) {
+      return transactionEventResponse;
+    } else if (authorizations.length > 1) {
```
lydiazcheng, line 324: The execution never comes into this condition since it returns when `authorizations.length !== 1`. If we follow the idea in the original code, i.e., only continue when the length is 1, I think we can remove this part.

03_Modules/Transactions/src/module/module.ts
```diff
@@ -412,6 +305,127 @@ export class TransactionsModule extends AbstractModule {
...
+        );
+      }
+
+      // TODO there should only be one active transaction per evse of a station.
+      // old transactions should be marked inactive and an alert should be raised (this can only happen in the field with charger bugs or missed messages)
+
+      // Check for ConcurrentTx
+      return await this._transactionEventRepository
+        .readAllActiveTransactionsByIdToken(transactionEvent.idToken)
+        .then((activeTransactions) => {
```
lydiazcheng, line 412: Minor: it would be perfect if we can get rid of the Promise.then here as well.

03_Modules/Transactions/src/module/module.ts
langiole: Done!

03_Modules/Transactions/src/module/module.ts
langiole: Done!

lydiazcheng (approved):

PR #177, opened 2024-07-22
(Release 1.4) Fix: Subscriptions API

lydiazcheng:
## Changes
- Fix post endpoint that accepts subscription bodies without stationIds
- Fix post endpoint does not appear correctly in OpenAPI (missing request body)
- Fix error log when push to subscription fails is unclear
- Add schema for get and del endpoints.

⚠️ The fix belongs to Release 1.4. Need to change the target branch before merging.

## Tests
1. If missing stationId Screenshot 2024-07-22 at 1 26 14 PM
2. If all subscription events are false Screenshot 2024-07-22 at 3 51 01 PM
3. If post to subscription url returns not ok response  Screenshot 2024-07-22 at 3 52 31 PM
4. If fetch fails Screenshot 2024-07-22 at 3 53 14 PM
5. Swagger open API: Screenshot 2024-07-23 at 10 36 47 AM


03_Modules/OcppRouter/src/module/api.ts
```diff
@@ -55,16 +58,35 @@ export class AdminApi
...
   async postSubscription(
     request: FastifyRequest<{ Body: Subscription }>,
   ): Promise<number> {
+    if (
+      !request.body.onClose &&
+      !request.body.onConnect &&
+      !request.body.onMessage &&
+      !request.body.sentMessage
+    ) {
+      throw new Error(
```
ChrisWeissmann, line 76: Nitpick: shouldn't we be returning a HTTP 400 (instead of a 500) in this case? The client is sending us something that isn't allowed.

ChrisWeissmann (approved):
Left a nitpick. But it's not mission critical.

03_Modules/OcppRouter/src/module/api.ts
lydiazcheng: It looks like below now,
Screenshot 2024-07-24 at 11 10 38 AM

PR #178, opened 2024-07-23
Bugfix: making @fastify/swagger static to 8.14.0

yuvenus:
The new version of @fastify/swagger@8.15.0 changes the type declarations of the `swaggerObject` and `openapiObject` and causes build errors. This PR is the quick fix to this issue while we properly handle dependency changes like this.

### DIFF
--- a/02_Util/package.json
+++ b/02_Util/package.json
@@ -38,7 +38,7 @@
     "@citrineos/base": "1.3.1",
     "@citrineos/data": "1.3.1",
     "@directus/sdk": "^15.0.3",
-    "@fastify/swagger": "^8.10.1",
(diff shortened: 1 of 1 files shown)


lydiazcheng (approved):
The target branch should be rc-1.3.2?  Sorry I don't have the merge permission.

danielyoo227 (approved):

yuvenus:
> The target branch should be rc-1.3.2? Sorry I don't have the merge permission.

since main branch can't even build because of this problem, i targeted to main 😱 can definitely retarget though if it makes sense somewhere else :D

elliot-sabitov (approved):

PR #180, opened 2024-07-24
OCPI Tokens Patch support

danielyoo227:
Adding transaction support for creating authorization so that we can create / patch a token in a transaction when updating OCPP's authorization object.

Also exporting out the Sequelize.Transaction as a separate object so that we can integrate this in other libraries that depend on core/data


01_Data/src/index.ts
```diff
@@ -3,6 +3,8 @@
 //
 // SPDX-License-Identifier: Apache 2.0
 
+import { Transaction as SequelizeTransaction } from 'sequelize';
```
yuvenus, line 6: dumb question... shouldn't this already be exportable as `SequelizeTransaction`, rather than needing to do `import { Transaction as SequelizeTransaction }` ?

yuvenus (approved):
looks good! just a quick comment 👍

01_Data/src/index.ts
elliot-sabitov: @yuvenus I think this is referring to the SequelizeTransaction coming from `sequelize` - used for ensuring that a group of DB queries can be executed in a single transaction. I think maybe this was done to try and reduce the confusion between this and the Transaction models that we have repressing the charging session

PR #182, opened 2024-07-25
(Release 1.4) Test introduction

ChrisWeissmann:
Started to refactor boot notification to extract some logic from the handler to make it testable.

Added initial Jest setup and some tests.

### DIFF
--- a/03_Modules/Configuration/jest.config.js
+++ b/03_Modules/Configuration/jest.config.js
@@ -1,8 +1,0 @@
-/** @type {import('ts-jest').JestConfigWithTsJest} **/
-module.exports = {
-  testEnvironment: 'node',
-  transform: {
-    '^.+.tsx?$': ['ts-jest', {}],
-  },
-  displayName: 'Configuration Module',
-};
--- a/03_Modules/Configuration/jest.config.js
+++ b/03_Modules/Configuration/jest.config.js
@@ -1,0 +1,8 @@
+/** @type {import('ts-jest').JestConfigWithTsJest} **/
+module.exports = {
+  testEnvironment: 'node',
+  transform: {
+    '^.+.tsx?$': ['ts-jest', {}],
+  },
+  displayName: 'Configuration Module',
+};
--- a/03_Modules/Configuration/src/module/bootService.ts
+++ b/03_Modules/Configuration/src/module/bootService.ts
@@ -1,0 +1,91 @@
+import { Boot, IBootRepository } from '@citrineos/data';
+import {
+  BootConfig,
+  BootNotificationResponse,
+  RegistrationStatusEnumType,
+} from '@citrineos/base';
+
+export class BootService {
+  protected _bootRepository: IBootRepository;
+
+  constructor(bootRepository: IBootRepository) {
+    this._bootRepository = bootRepository;
+  }
+
+  determineBootStatus(
+    bootConfig: Boot | undefined,
+    configuration: any,
+  ): RegistrationStatusEnumType {
+    let bootStatus = bootConfig
+      ? bootConfig.status
+      : configuration.unknownChargerStatus;
+
+    if (bootStatus === RegistrationStatusEnumType.Pending) {
+      let needToGetBaseReport = configuration.getBaseReportOnPending;
+      let needToSetVariables = false;
+      if (bootConfig) {
+        if (
+          bootConfig.getBaseReportOnPending !== undefined &&
+          bootConfig.getBaseReportOnPending !== null
+        ) {
+          needToGetBaseReport = bootConfig.getBaseReportOnPending;
+        }
+        if (
+          bootConfig.pendingBootSetVariables &&
+          bootConfig.pendingBootSetVariables.length > 0
+        ) {
+          needToSetVariables = true;
+        }
+      }
+      if (
+        !needToGetBaseReport &&
+        !needToSetVariables &&
+        configuration.autoAccept
+      ) {
+        bootStatus = RegistrationStatusEnumType.Accepted;
+      }
+    }
+
+    return bootStatus;
+  }
+  createBootNotificationResponse(
+    bootConfig: Boot | undefined,
+    bootStatus: RegistrationStatusEnumType,
+    configuration: any,
+  ): BootNotificationResponse {
+    return {
+      currentTime: new Date().toISOString(),
+      status: bootStatus,
+      statusInfo: bootConfig?.statusInfo,
+      interval:
+        bootStatus === RegistrationStatusEnumType.Accepted
+          ? bootConfig?.heartbeatInterval || configuration.heartbeatInterval
+          : bootConfig?.bootRetryInterval || configuration.bootRetryInterval,
+    };
+  }
+
+  async updateBootConfig(
+    bootNotificationResponse: BootNotificationResponse,
+    stationId: string,
+  ): Promise<Boot> {
+    let bootConfigDbEntity: Boot | undefined =
+      await this._bootRepository.readByKey(stationId);
+    if (!bootConfigDbEntity) {
+      const unknownChargerBootConfig: BootConfig = {
+        status: bootNotificationResponse.status,
+        statusInfo: bootNotificationResponse.statusInfo,
+      };
+      bootConfigDbEntity = await this._bootRepository.createOrUpdateByKey(
+        unknownChargerBootConfig,
+        stationId,
+      );
+    }
+    if (!bootConfigDbEntity) {
+      throw new Error('Unable to create/update BootConfig...');
+    } else {
+      bootConfigDbEntity.lastBootTime = bootNotificationResponse.currentTime;
+      await bootConfigDbEntity.save();
+    }
+    return bootConfigDbEntity;
+  }
+}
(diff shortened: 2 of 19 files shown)


elliot-sabitov (approved):

03_Modules/Configuration/jest.config.js
```diff
@@ -0,0 +1,8 @@
+/** @type {import('ts-jest').JestConfigWithTsJest} **/
```
lydiazcheng, line 1: Very minor: I found jest also support typescript in their [doc](https://jestjs.io/docs/next/configuration). It would be more consist if we write this config script in ts.

03_Modules/Configuration/src/module/bootService.ts
```diff
@@ -0,0 +1,91 @@
...
+        !needToSetVariables &&
+        configuration.autoAccept
+      ) {
+        bootStatus = RegistrationStatusEnumType.Accepted;
+      }
+    }
+
+    return bootStatus;
+  }
+  createBootNotificationResponse(
```
lydiazcheng, line 51: Minor: Miss a new line between this method and the previous one.

lydiazcheng (approved):
The new test structure looks good!

03_Modules/Configuration/jest.config.js
ChrisWeissmann: fair! will adjust

03_Modules/Configuration/test/module/bootService.test.ts
```diff
@@ -0,0 +1,105 @@
...
+    });
+
+    it('should return Accepted status when bootConfig.status is pending and no actions are needed but autoAccept is true', () => {
+      const bootConfig = aValidBootConfig(
+        (item) => (item.getBaseReportOnPending = false),
+      );
+      const configWithAutoAccept = aValidConfiguration(
+        (item) => (item.autoAccept = true),
+      );
+      console.log(configWithAutoAccept);
```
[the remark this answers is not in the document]
ChrisWeissmann: ah yes. I must have overlooked that from my local testing. Will remove

03_Modules/Configuration/test/module/bootService.test.ts
```diff
@@ -0,0 +1,105 @@
...
+  describe('determineBootStatus', () => {
+    it('should return unknownChargerStatus if bootConfig is undefined', () => {
+      runDetermineBootStatusTest(
+        undefined,
+        aValidConfiguration(),
+        RegistrationStatusEnumType.Rejected,
+      );
+    });
+
+    it('should return bootConfig.status if bootConfig is defined and not pending', () => {
```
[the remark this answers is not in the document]
ChrisWeissmann: Sounds good. Will add an example for this.

03_Modules/Configuration/src/module/services.ts
langiole, line 1: Can this file not be renamed to something like `DeviceModelService.ts` 😅 ?

03_Modules/Configuration/src/module/services.ts
```diff
@@ -73,4 +73,152 @@ export class DeviceModelService {
       return Number(itemsPerMessageGetVariablesAttributes[0].value);
     }
   }
+
+  async updateDeviceModel(
```
langiole, line 77: I dumped this method into chatgpt and it came up with this:
```
async updateDeviceModel(
  chargingStation: any,
  stationId: string,
  timestamp: string,
): Promise<void> {
  const attributes = [
    { component: 'ChargingStation', variable: 'Model', value: chargingStation.model },
    { component: 'ChargingStation', variable: 'VendorName', value: chargingStation.vendorName },
    { component: 'Controller', variable: 'FirmwareVersion', value: chargingStation.firmwareVersion },
    { component: 'ChargingStation', variable: 'SerialNumber', value: chargingStation.serialNumber },
    { component: 'DataLink', variable: 'IMSI', value: chargingStation.modem?.imsi },
    { component: 'DataLink', variable: 'ICCID', value: chargingStation.modem?.iccid }
  ];

  const promises = attributes
    .filter(attr => attr.value !== undefined) // Filter out attributes with undefined values
    .map(attr => this._deviceModelRepository.createOrUpdateDeviceModelByStationId(
      {
        component: { name: attr.component },
        variable: { name: attr.variable },
        variableAttribute: [
          {
            type: AttributeEnumType.Actual,
            value: attr.value,
            mutability: MutabilityEnumType.ReadOnly,
            persistent: true,
            constant: true,
          },
        ],
      },
      stationId,
      timestamp,
    ));

  await Promise.all(promises); // Wait for all promises to resolve
}
```

langiole: Would probably be nice to have this as a bulk db update.

03_Modules/Configuration/test/providers/bootConfig.ts
langiole, line 1: Pretty neat optimization from chatgpt:

You can define a utility function:
```
type UpdateFunction<T> = (item: T) => void;

const applyUpdateFunction = <T>(item: T, updateFunction?: UpdateFunction<T>): T => {
  if (updateFunction) {
    updateFunction(item);
  }
  return item;
};
```

And then update your providers to look like:
```
export const aValidSetVariableResult = (
  updateFunction?: UpdateFunction<SetVariableResultType>,
): SetVariableResultType => {
  const item: SetVariableResultType = {
    attributeStatus: SetVariableStatusEnumType.Accepted,
    component: {
      name: faker.lorem.word(),
      instance: faker.string.uuid(),
    },
    variable: {
      name: faker.lorem.word(),
      instance: faker.string.uuid(),
    },
    customData: {
      vendorId: faker.string.uuid(),
    },
    attributeType: AttributeEnumType.Actual,
    attributeStatusInfo: {
      reasonCode: faker.lorem.word(),
      additionalInfo: faker.lorem.sentence(),
    },
  };

  return applyUpdateFunction(item, updateFunction); <- the update
};
```

So you eliminate the need to duplicate the `if (updateFunction) { ... }` condition for each provider.

03_Modules/Configuration/test/providers/bootConfig.ts
```diff
@@ -0,0 +1,77 @@
...
+  const item: Boot = {
+    id: faker.string.uuid(),
+    lastBootTime: faker.date.recent().toISOString(),
+    heartbeatInterval: faker.number.int({ min: 30, max: 3600 }),
+    bootRetryInterval: faker.number.int({ min: 30, max: 3600 }),
+    status: RegistrationStatusEnumType.Pending,
+    getBaseReportOnPending: faker.datatype.boolean(),
+    variablesRejectedOnLastBoot: [aValidSetVariableResult()],
+    bootWithRejectedVariables: faker.datatype.boolean(),
+  } as Boot;
```
langiole, line 54: This `as Boot` should not be needed here.

03_Modules/Configuration/src/module/bootService.ts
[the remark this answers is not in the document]
ChrisWeissmann: So in this case you are suggesting a `BootNotificationService`?  And inside, it would also update the device model, instead of calling the device model service?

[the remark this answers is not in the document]
langiole: I am suggesting you create a `BootNotificationHandler.ts`, and put all the logic that is needed to fulfill this request within it.

It would contain this piece of code that is currently in `module.ts`:
```
  @AsHandler(CallAction.BootNotification)
  protected async _handleBootNotification(
    message: IMessage<BootNotificationRequest>,
    props?: HandlerProperties,
  ): Promise<void> {
```

The idea is that you have a handler ts file per message instead of putting numerous handlers into a single `module.ts`.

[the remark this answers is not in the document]
langiole: It can still call the `DeviceModelService.ts` like you already have.

03_Modules/Configuration/src/module/services.ts
ChrisWeissmann: I can see that refactor of being good. Will adjust

03_Modules/Configuration/test/providers/bootConfig.ts
ChrisWeissmann: yea, nice one!

03_Modules/Configuration/test/providers/bootConfig.ts
ChrisWeissmann: I think it's needed, else there are issues with it being a class instance instead of an object. And it can't be a class instance here as we want to avoid having to deal with the squelize part of it. 
With out the `as Boot` it freaks out that it's missing the sequelize added things such as `$get`

03_Modules/Configuration/src/module/module.ts
```diff
@@ -106,6 +107,7 @@ export class ConfigurationModule extends AbstractModule {
   protected _bootRepository: IBootRepository;
   protected _deviceModelRepository: IDeviceModelRepository;
   protected _messageInfoRepository: IMessageInfoRepository;
+  protected bootService: BootNotificationService;
```
yuvenus, line 110: very minor, should this be `_bootService` to match with all the other repositories and service names? 🤔

yuvenus (approved):
hooray tests! did you happen to run the linter locally btw? seems like it's warning about some unused imports that could be removed 👀

03_Modules/Configuration/src/module/module.ts
ChrisWeissmann: Sure thing!

PR #185, opened 2024-07-26
Chore/lint fix

elliot-sabitov:


langiole:
I'm still seeing version bumping going on in this PR

lydiazcheng:
> I'm still seeing version bumping going on in this PR

Should be related to the target branch. We might use rc-1.3.2 instead of main. 🤔

00_Base/src/index.ts
```diff
@@ -134,11 +134,9 @@ import {
 } from './ocpp/model/index';
 import { CallAction } from './ocpp/rpc/message';
 
-export interface OcppRequest {
-}
+export interface OcppRequest {}
```
lydiazcheng, line 137: Just a question, when I run lint-fix, it will format the code in the way of red line 137 and 138. Maybe you have an idea of which setting cause this?

elliot-sabitov: 😮 That is very strange, are you on a windows machine?

elliot-sabitov:
adjusted to point to rc-1.3.2

00_Base/src/index.ts
lydiazcheng: 😢 No, I use Mac M1.

langiole: Just ran with linter and got no changes.

langiole (approved):

lydiazcheng (approved):

00_Base/src/index.ts
lydiazcheng: Should be sth wired only on my machine. Never mind.

PR #191, opened 2024-07-26
(Release 1.4) Feature: Reservation

lydiazcheng:
⚠️ The fix belongs to Release 1.4. Need to change the target branch before merging.

## Changes
Implements use cases in Reservation in OCPP 2.0.1 Part5 Certification Profiles.


01_Data/src/layers/sequelize/model/CallMessage.ts
```diff
@@ -0,0 +1,21 @@
...
+
+  @Index
+  @Column({
+    type: DataType.STRING,
+    unique: true,
+  })
+  declare correlationId: string;
+
+  @Column(DataType.INTEGER)
+  declare reservationId?: number;
```
ChrisWeissmann, line 20: Why make this optional? If I understand this correctly, the only point is to be able to find the reservationId via correlationId

ChrisWeissmann (approved):

01_Data/src/layers/sequelize/model/CallMessage.ts
lydiazcheng: Hi Christian, thanks for reviewing! 
My idea is that we may extend this table to use it for other models in the same way as well, e.g, adding a optional `chargingProfileId` col. In this case, reservationId should be optional.

ChrisWeissmann: Ah got it. Would you mind adding a comment on your intent for the class? That it will be extended for other correlationId related lookups?

PR #192, opened 2024-07-26
Feature: Smart charging ISO15118

lydiazcheng:
## Purpose
The task is to implement all use cases in **ISO 15118 Smart Charging** in **ISO 15118 support** in OCPP 2.0.1 V3 Part 5

After checking, most use cases are implemented in other tasks, e.g., Set charging profile, or don't need new implementation, e.g., ISO 15118 signed meter values

So only **Renegotiating a Charging Schedule** (including K16 and K17) is in this PR.

## Changes
1. Implement Renegotiating a Charging Schedule, K16 and K17
2. Add Smart Charging interface so that we can integrate with external smart charging service.
3. Fix some query codes in charging profile and transaction repositories. Remove unique constrains from ChargingNeeds table since OCPP protocol doesn't require allow only one needs per transaction.
4. Fix few lint issues.

## Tests
**Case 1: Renegotiating a Charging Schedule - Initiated by EV (Ref: OCPP 2.0.1 V3 Part 6 TC_K_57_CS)**
0. Prepare a ongoing transaction.
1. The Charging Station sends a NotifyEVChargingNeedsRequest Screenshot 2024-08-09 at 2 00 29 PM
2. Citrine responds with a NotifyEVChargingNeedsResponse with status Accepted 
3. Citrine sends a SetChargingProfileRequest Screenshot 2024-08-09 at 2 00 45 PM
5. The Charging Station responds with a SetChargingProfileResponse Screenshot 2024-08-09 at 2 01 06 PM
6. The Charging Station sends a NotifyEVChargingScheduleRequest
7. Citrine responds with a NotifyEVChargingScheduleResponse. With status Accepted Screenshot 2024-08-09 at 2 03 20 PM

**Case 2: Schedule Limit in the NotifyEVChargingScheduleRequest is bigger than existing one**
In this case, Citrine still accepts the request but send another setChargingProfileRequest. Screenshot 2024-08-09 at 2 05 02 PM

**Case 3: Calculated schedule limit based on the NotifyEVChargingNeedsRequest is bigger than existing one**
Citrine should reject the request. Screenshot 2024-08-09 at 1 27 24 PM Screenshot 2024-08-09 at 1 27 46 PM


02_Util/src/smartCharging/LocalSmartCharging.ts
```diff
@@ -0,0 +1,265 @@
...
+} from '@citrineos/base';
+import {
+  IChargingProfileRepository,
+  Transaction,
+  ChargingSchedule,
+  ChargingProfile,
+} from '@citrineos/data';
+import { ILogObj, Logger } from 'tslog';
+
+export class LocalSmartCharging implements ISmartCharging {
```
ChrisWeissmann, line 21: Should this be in 02_Util or rather in the smart charging module? I would think rather in the module as I don't think it would be use anywhere else?

I might be missing something, but why is this bound to local smartcharging? K03 defines local smart charging as:
> Local Smart Charging describes a use case in which smart charging enabled Charging Stations have charging limits controlled locally by a Local Controller, not directly by the CSMS.

Are we going to add an implementation for Central smart charging?

lydiazcheng: I moved these files into the module itself.

The file name is indeed confusing. Thanks for pointing this out! It is not related to local smart charging in OCPP protocol. 
In the future, we plan to integration with external smart charging services which calculate charging profiles for us. But before integration, we create this script to achieve some basic calculation or validation. I rename it to `InternalSmartCharging`.  Please feel free to let me know if there is any better fitting name.

ChrisWeissmann (approved):

PR #195, opened 2024-07-29
(Release 1.4) Update NPM scripts to use cross-env where possible

yuvenus:
OCPI PR: https://github.com/citrineos/citrineos-ocpi/pull/123
PR to update citrineos.github.io: https://github.com/citrineos/citrineos.github.io/pull/31

This PR adds the `cross-env` dependency (https://www.npmjs.com/package/cross-env) to allow cross-OS usability of the same NPM scripts that need to set environment variables. The two targeted commands were `start` and `start-everest`, but if there are more commands I should update, let me know!


ChrisWeissmann (approved):
Nice! Probably need an update to the citrineos.github.io site.

yuvenus:
> Nice! Probably need an update to the citrineos.github.io site.

oooh good point! i can open a PR for that and link it to this one so we know when that change should be published 👍

langiole:
So sweet 😮‍💨

yuvenus:
note: not sure what happened but github seems to think changes that are already in branch are still changes 😭

PR #199, opened 2024-07-30
fix: ensuring authorizer is passed into the Transactions module as well

elliot-sabitov:


yuvenus (approved):
approved! eventually though if we separate out the ocpi stuff this will have to be "undone" right?

elliot-sabitov:
> approved! eventually though if we separate out the ocpi stuff this will have to be "undone" right?

yep, we will adjust that in OCPI /Server to include this, but Core /Server will not include anything OCPI related

PR #201, opened 2024-07-31
feat: add github action to run npm test

ChrisWeissmann:
Automate running the npm tests when a pull request is opened.

### DIFF
--- a/.github/workflows/jest-tests.yml
+++ b/.github/workflows/jest-tests.yml
@@ -4,7 +4,7 @@ on:
   pull_request:
 
 jobs:
-  test:
+  jest-tests:
     runs-on: ubuntu-latest
 
     steps:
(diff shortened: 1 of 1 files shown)


elliot-sabitov (approved):

.github/workflows/jest-tests.yml
```diff
@@ -0,0 +1,31 @@
...
+          path: ~/.npm
+          key: ${{ runner.os }}-node-${{ hashFiles('**/package-lock.json') }}
+          restore-keys: |
+            ${{ runner.os }}-node-
+
+      - name: Install dependencies
+        run: npm install && npm run build
+
+      - name: Run tests
+        run: npm test
```
elliot-sabitov, line 31: @ChrisWeissmann dont we need like `npm run test`? It may be good to have everything reference the single script/command in package.json that way if you ever need to make a change, you only need to adjust that one place... what do you think?

PR #204, opened 2024-07-31
feat: adjusted setup such that OCPI is not referenced anywhere in cor…

elliot-sabitov:
- feat: adjusted setup such that OCPI is not referenced anywhere in core. Reverting workflows to not clone ocpi, config to no longer have ocpi component, docker compose to trigger docker file without ocpi, nodemon to not refresh when there are file changes in ocpi, /Server dependencies, tsconfig, etc. and running server will now run without OCPI,
- fix: issues with mismatching dependencies between core and OCPI that I think may have only became visible recently when we adjusted `npm run fresh` to clear npm cache to make sure we are really starting fresh.
- chore: removing --workspaces to ensure that all dev dependencies are installed along with the workspace dependencies otherwise some dependencies are missed.


yuvenus (approved):
nice! just curious about those directory changes, is it just to make sure the way we resolve paths is consistent throughout the project? :O

elliot-sabitov:
> nice! just curious about those directory changes, is it just to make sure the way we resolve paths is consistent throughout the project? :O

this is primarily for ensuring that the build that uses those .ts files simply references the assets that are in that relative folder and we make sure it is present as it is included in the build. These can always be overridden, but we are essentially doing this right now anyway, where these assets are being copied somewhere onto the docker container for example and we write out a static patch to the asset file in the container, like so:
'/usr/local/apps/citrineos-core/Server/src/assets/certificates/rootCertificate.pem',
but it anything changes, like the name of the folders, etc. then this needs to be updated. I think with
path.resolve(
              path.dirname(__filename),
              '../../assets/certificates/rootCertificate.pem',
            ),
alternative, it still should reference the same file, but make it easier to reference since it is relative to the .ts file rather then the full path to the file on the file system. It is important to note that both can be used though, so we are flexible, but since we are committing this env files and using them with the assets, it made sense for me to make this simplification…

00_Base/src/config/types.ts
```diff
@@ -226,10 +226,6 @@ export const systemConfigInputSchema = z.object({
   logLevel: z.number().min(0).max(6).default(0).optional(),
   maxCallLengthSeconds: z.number().int().positive().default(5).optional(),
   maxCachingSeconds: z.number().int().positive().default(10).optional(),
-  ocpiServer: z.object({
```
thanaParis, line 229: Please keep this here.

01_Data/src/layers/sequelize/repository/Authorization.ts
```diff
@@ -44,14 +40,16 @@ export class SequelizeAuthorizationRepository extends SequelizeRepository<Author
     }
 
     const savedAuthorizationModel = await Authorization.findOne({
-      include: [{
```
thanaParis, line 47: you merged a lint-fix branch into 1.3.2. Why is this formatting change still happening? We need to be more strict with this, we can't have every pr contain irrelevant formatting changes.

Server/copy-assets.ts
```diff
@@ -0,0 +1,37 @@
+import * as path from 'path';
```
thanaParis, line 1: Please put this change into its own pr.

Server/src/config/envs/docker.ts
```diff
@@ -106,12 +110,18 @@ export function createDockerConfig() {
...
             port: 8443,
             protocol: 'ocpp2.0.1',
-            tlsKeyFilePath:
-              '/usr/local/apps/citrineos-core/Server/src/assets/certificates/leafKey.pem',
-            tlsCertificateChainFilePath:
-              '/usr/local/apps/citrineos-core/Server/src/assets/certificates/certChain.pem',
-            rootCACertificateFilePath:
-              '/usr/local/apps/citrineos-core/Server/src/assets/certificates/rootCertificate.pem',
+            tlsKeyFilePath: path.resolve(
+              path.dirname(__filename),
```
thanaParis, line 114: These changes belong in their own PR.

thanaParis (changes requested):

00_Base/src/ocpp/model/enums/index.ts
```diff
@@ -808,7 +808,7 @@ export const enum UpdateEnumType {
  * within the CSMS.
  *
  */
-export const enum RegistrationStatusEnumType {
+export enum RegistrationStatusEnumType {
```
ChrisWeissmann, line 811: Out of curiosity, what triggered this change. From why I understand, the const does an inline replacement. But I am not quite aware of the other tradeoffs here.

01_Data/src/layers/sequelize/repository/Authorization.ts
elliot-sabitov: 💯 yeah let me try running lint again, I am not sure what happened here...

elliot-sabitov: I created this PR here - https://github.com/citrineos/citrineos-core/pull/208 to lint-fix... I am not sure how changes got in without the lint catching these, but I do see this change in the lint PR so I think once that one is merged, then this PR should no longer show this 👍

00_Base/src/ocpp/model/enums/index.ts
elliot-sabitov: const enums are not available at run time. Essentially, those enums get transpiled as explicit values and the enum itself ends up not being available at run time. So when you do code like:
```
if (someVal === SeomEnum.SomeVal) {
  ...
}
```
for example, in your code, it will work, but the JS code will be something along the lines of
```
if (someVal === 'SomeVal') {
  ...
}
```
which can be fine.. but I was facing an issue where if this particular enum is used in an annotation, for example like so: 
```
@IsNotEmpty()
@IsEnum(RegistrationStatusEnumType)
@Enum(RegistrationStatusEnumType, 'RegistrationStatusEnumType')
unknownChargerStatus: RegistrationStatusEnumType;
```
then it does not accept a `const enum`...

Server/src/config/envs/docker.ts
elliot-sabitov: not a problem 👍 I can quickly split this out into a separate PR, will adjust

Server/copy-assets.ts
elliot-sabitov: will split into separate PR

Server/src/config/envs/docker.ts
elliot-sabitov: created a separate PR here - https://github.com/citrineos/citrineos-core/pull/210

00_Base/src/config/types.ts
elliot-sabitov: adjusted

Server/copy-assets.ts
elliot-sabitov: separate PR https://github.com/citrineos/citrineos-core/pull/209

elliot-sabitov:
Broke this original PR into 4 by creating 3 additional PRs:
- https://github.com/citrineos/citrineos-core/pull/211
- https://github.com/citrineos/citrineos-core/pull/210
- https://github.com/citrineos/citrineos-core/pull/209

00_Base/src/ocpp/model/enums/index.ts
ChrisWeissmann: Ah, got it. For you it was around the annotations. I ran into this when creating tests and fakers. 
So probably the inlining (your second code snippet) is more efficient. But maybe not worth the efficiency gain for a node backend application against keeping it as an object where you can to more operations against.

elliot-sabitov: I think in cases where the enum is quite large and used in a lot of places, perhaps the performance gain, of not having the enum object in run-time but instead having static strings everywhere in the running JS can have some impact, especially if it is compounded with other enums as the project scales and in partifular for enums that have high frequency of lookups since the `const enums` will be returning values directly while `enums` need to look it up in the object. On the other hand, if you have a small enum that is only used in a small hand full of places, rarely running, and you need to access the actual enum mapping, for example, if you ever need to iterate over the enum, or if you want to have a dynamic `valueFrom(str)` then with the `const enum` approach that can only be achieved via a switch or if statements. I guess what I am trying to say is that there is most likely a balance that needs to be struck identifying enums that need to exist during run-time and their use-cases, vs the enums that should just be transpiled into const strings.

thanaParis:
@elliot-sabitov I think we still need this PR, to remove the references to OCPI. Can you fix the lint & merge conflict issues?

thanaParis (changes requested):
fix lint & merge conflicts

ChrisWeissmann (approved):

thanaParis (approved):
Approved!

PR #205, opened 2024-08-01
Subscriptions robustness

falanadamian:
Implemented a feature to refresh subscriptions at regular intervals for connected stations, ensuring that data remains current and synchronized. I also extracted and refactored the code responsible for dispatching messages.


03_Modules/OcppRouter/src/module/webhook.dispatcher.ts
```diff
@@ -0,0 +1,288 @@
...
+  addOnCloseCallback(
+    identifier: string,
+    onCloseCallback: OnCloseCallback,
+  ): void {
+    this._onCloseCallbacks.has(identifier)
+      ? this._onCloseCallbacks.get(identifier)!.push(onCloseCallback)
+      : this._onCloseCallbacks.set(identifier, [onCloseCallback]);
+  }
+
+  addOnMessageCallback(
```
ChrisWeissmann, line 56: I am a bit confused about the flow here. So when we add a subscription it's done via the  API which directly creates the data in the database. Then every once in a while you refresh the subscriptions by getting them from the database. But doesn't that introduce a long delay until the subscriptions get picked up? Shouldn't the `add...` be called when the subscription is added via API?

03_Modules/OcppRouter/src/module/webhook.dispatcher.ts
```diff
@@ -0,0 +1,288 @@
...
+    const sentMessageCallbacks: OnSentMessageCallback[] = [];
+
+    const subscriptions =
+      await this._subscriptionRepository.readAllByStationId(
+        connectionIdentifier,
+      );
+
+    for (const subscription of subscriptions) {
+      if (subscription.onConnect) {
+        onConnectionCallbacks.push(this._onConnectionCallback(subscription));
```
ChrisWeissmann, line 142: I wonder if this makes sense, shouldn't all instances hold all subscriptions for `onConnection`? Since onConnect fires when the charger connects, shouldn't all instances hold all onConnection subscriptions, else you only pick them up when the charger is already connected, since you filter by the connection that the instance is connected to.

03_Modules/OcppRouter/src/module/webhook.dispatcher.ts
ChrisWeissmann: Ah, apologies, I might have been over optimizing here. We had discussed that 3 min delay was acceptable against introducing a bunch of new complexity of getting instant consistency between the instances.

03_Modules/OcppRouter/src/module/webhook.dispatcher.ts
ChrisWeissmann: Since the subscriptions get loaded when the charger connects, and before the subscription actions are triggered, this should be fine. I misinterpreted when the subscriptions get loaded.

ChrisWeissmann (approved):
Looks good to me

03_Modules/OcppRouter/src/module/webhook.dispatcher.ts
```diff
@@ -0,0 +1,301 @@
...
+      event: string;
+      origin?: MessageOrigin;
+      message?: string;
+      error?: any;
+      info?: Map<string, string>;
+    },
+    url: string,
+  ): Promise<boolean> {
+    try {
+      const response = await fetch(url, {
```
elliot-sabitov, line 260: Can we utilize `typed-rest-client` here. We use this for performing HTTP requests in OCPI for PUSH messages, perhaps it can also be helpful here. We do have a BaseClientApi that may not 100% fit into this because we set OCPI headers and stuff, so perhaps we can incorporate a new, more generic base... It can be good to reuse the same rest client when performing external HTTP requests to keep things organized..

03_Modules/OcppRouter/src/module/webhook.dispatcher.ts
thanaParis: @ChrisWeissmann just for posterity--after we move away from Directus, we will revisit this. the need for polling the database may go away after adopting a new UI stack, since it's data entry via Directus that this polling is needed to catch.

thanaParis (approved):
Great work!

PR #209, opened 2024-08-05
chore: created npm run copy-assets

elliot-sabitov:
chore: created npm run copy-assets, in a simple helper script to ensure that it can run properly across platforms. Triggering copy-assets in postbuild to ensure that assets folder is copied over otherwise we face errors


Server/copy-assets.ts
```diff
@@ -0,0 +1,37 @@
+import * as path from 'path';
+import * as fs from 'fs';
+import * as util from 'util';
+
+const mkdir = util.promisify(fs.mkdir);
```
ChrisWeissmann, line 5: I think you could use fs/promises here instead of `promisify` everything by hand. 
https://nodejs.org/api/fs.html#promises-api

Server/copy-assets.ts
```diff
@@ -0,0 +1,37 @@
...
+const readdir = util.promisify(fs.readdir);
+const stat = util.promisify(fs.stat);
+const copyFile = util.promisify(fs.copyFile);
```
ChrisWeissmann, line 8: If you use the fs/promises you don't have to create these costs but can use them right away https://nodejs.org/api/fs.html#fspromisescpsrc-dest-options

Server/copy-assets.ts
```diff
@@ -0,0 +1,37 @@
...
+    }
+  } catch (err) {
+    console.error('Error copying directory:', err);
+  }
+}
+
+async function main() {
+  const srcDir = path.resolve(__dirname, './src/assets');
+  const destDir = path.resolve(__dirname, './dist/assets');
+  await copyDirectory(srcDir, destDir);
```
ChrisWeissmann, line 33: If you use `fs/promise.cp` I think you can just pass in do `recursive: true`  and get rid of this recursive for loop
`fs.cp(src, dest, { recursive: true });`
Then you can get rid of the manual `copyDirectory` and maybe just wrap with a try catch here.

elliot-sabitov: adjusted

Server/copy-assets.ts
elliot-sabitov: adjusted

PR #210, opened 2024-08-05
fix: asset path resolution issues

elliot-sabitov:
- chore: adjusting config values that reference a file in the assets directory to reference relatively to the asset files to ensure that they can be loaded and used easily otherwise there seem to be issues with loading assets and had to adjust to use path.resolve(path.dirname(__filename)…) to ensure that the file where this code was written is used instead of the runtime file that is running the code, which is what happens with path.resolve(__dirname…)


lydiazcheng (approved):

ChrisWeissmann (approved):
Looks good to me. 🚀

PR #211, opened 2024-08-05
chore: created npm run clear cache and enhancing npm run fresh to als…

elliot-sabitov:
- chore: removing —workspaces when running npm i as it does not seem to be used
- chore: created npm run clear cache and enhancing npm run fresh to also clear cache after noticing issues after clearing local cache, clearing cache with fresh should make sure that any future issues are captured and resolved early.
- chore: enhancing npm run clean to properly delete dist and tsbuildinfo after discovering issue where dist was being removed from node_modules resulting in errors finding external modules after running clean / rebuild, then requiring re-install


ChrisWeissmann (approved):
Do you want to add any docs on the new commands?

PR #212, opened 2024-08-06
Moved TransactionEvent handler logic into seperate file

langiole:


03_Modules/Transactions/src/module/module.ts
```diff
@@ -189,10 +186,20 @@ export class TransactionsModule extends AbstractModule {
 
     this._authorizers = authorizers || [];
 
-    this._sendCostUpdatedOnMeterValue =
-      config.modules.transactions.sendCostUpdatedOnMeterValue;
+    this._sendCostUpdatedOnMeterValue = config.modules.transactions.sendCostUpdatedOnMeterValue;
     this._costUpdatedInterval = config.modules.transactions.costUpdatedInterval;
 
+    this._handlers.set(CallAction.TransactionEvent, new TransactionEventHandler(
```
elliot-sabitov, line 192: Instead of this - we can consider going a step further and annotate the `TransactionEventHandler` like so:
```
@NewAsHandlerAnnoation(CallAction.TransactionEvent)
export class TransactionEvent...
```
and then, in your handler definition, you would set some `Reflect.defineMetadata` .. then, in a single other place in the code (or perhaps multiple if really needed), and probably right here, you should be able to get all of the annotated classes using `Reflect.getMetadata`, loop through them, and do this `this.handlers.set`... or a mix in between.

PR #214, opened 2024-08-07
(Release 1.4) Signed Meter Values support

yuvenus:
### Description
This PR adds logic to process signed meter values signed with `RSASSA-PKCS1-v1_5`, which is one of the allowed signing algorithms provided by OCPP 2.0.1 spec. 

Currently, every charging station will share the same public key config as configured in the system config. A future enhancement will be to make this more flexible for different use cases (such as signing with ECDSA and having different keys for different sets of charging stations).

Note that the "built-in" `node:crypto` library was used over `jsrasign` because `jsrasign` no longer supports `RSASSA-PKCS1-v1_5`.

In general, a signed meter value is invalid if:
1. `SignedMeterValuesConfig` is not configured (no public key, no signing method, etc.)
2. The incoming signed meter value's signing method doesn't match the configured signing method
3. The incoming signed meter value's public key is empty and there isn't one already stored for the charging station in a `ChargingStationSecurityInfo`
4. The incoming signed meter value's public key isn't empty, but it doesn't match what's stored in the configured public key

### Changes:

1. Added new util `SignedMeterValuesUtil` to process and validate meter values
6. Added new class `ChargingStationSecurityInfo` to store key information for charging stations + TODO to expand upon this in a future release
7. Added repository for `ChargingStationSecurityInfo`
8. Added `SignedMeterValuesConfig` + updated SystemConfig with optional config for signed meter values
9. Updated MeterValues handler to store incoming meter values
10. Updated MeterValues handler to validate incoming signed meter values
11. Updated TransactionEvent handler to validate incoming signed meter values


01_Data/src/layers/sequelize/model/ChargingStationSecurityInfo.ts
```diff
@@ -0,0 +1,23 @@
...
+export class ChargingStationSecurityInfo extends Model {
+  static readonly MODEL_NAME: string = Namespace.ChargingStationSecurityInfo;
+
+  @Column({
+    type: DataType.STRING,
+    unique: true,
+  })
+  stationId!: string;
+
+  // TODO: store public key information into the database
```
elliot-sabitov, line 17: does it only work with the one charger until this todo is completed?

02_Util/src/security/SignedMeterValuesUtil.ts
```diff
@@ -0,0 +1,209 @@
...
+  public async validateMeterValues(
+    stationId: string,
+    meterValues: [MeterValueType, ...MeterValueType[]],
+  ): Promise<boolean> {
+    let validMeterValues = true;
+
+    const expectPublicKey =
+      await this._deviceModelRepository.readAllByQuerystring({
+        stationId,
+        component_name: 'OCPPCommCtrlr',
```
elliot-sabitov, line 67: since these are part of the spec, would it make sense to create some enums around the component name and the variable names? I see that these are the possible values (below), so maybe it may be helpful to have as enum since they are not truly dynamic strings... 
```
ControllerEnum {
  AlignedDataCtrlr
  AuthCacheCtrlr
  AuthCtrlr
  CHAdeMOCtrlr
  ClockCtrlr
  CustomizationCtrlr
  DeviceDataCtrlr
  DisplayMessageCtrlr
  ISO15118Ctrlr
  LocalAuthListCtrlr
  MonitoringCtrlr
  OCPPCommCtrlr
  ReservationCtrlr
  SampledDataCtrlr
  SecurityCtrlr
  SmartChargingCtrlr
  TariffCostCtrlr
  TxCtrlr
}

VariableName list
//-2.1-General
2.1.1.-ActiveNetworkProfile
2.1.2.-AllowNewSessionsPendingFirmwareUpdate
2.1.3.-DefaultMessageTimeout
2.1.4.-FileTransferProtocols
2.1.5.-HeartbeatInterval
2.1.6.-NetworkConfigurationPriority
2.1.7.-NetworkProfileConnectionAttempts
2.1.8.-OfflineThreshold
2.1.9.-QueueAllMessages
2.1.10.-MessageAttemptsTransactionEvent
2.1.11.-MessageAttemptIntervalTransactionEvent
2.1.12.-UnlockOnEVSideDisconnect
2.1.13.-WebSocketPingInterval
2.1.14.-ResetRetries
2.1.15.-MessageFieldLength
2.1.16.-ItemsPerMessageGetReport
2.1.17.-ItemsPerMessageGetVariables
2.1.18.-BytesPerMessageGetReport
2.1.19.-BytesPerMessageGetVariables
2.1.20.-ConfigurationValueSize
2.1.21.-ReportingValueSize
2.1.22.-ItemsPerMessageSetVariables
2.1.23.-BytesPerMessageSetVariables
2.1.24.-DateTime
2.1.25.-NtpSource
2.1.26.-NtpServerUri
2.1.27.-TimeOffset
2.1.28.-NextTimeOffsetTransitionDateTime
2.1.29.-TimeOffsetNextTransition
2.1.30.-TimeSource
2.1.31.-TimeZone
2.1.32.-TimeAdjustmentReportingThreshold
2.1.33.-CustomImplementationEnabled
//-2.2-Security-related
2.2.1.-BasicAuthPassword
2.2.2.-Identity
2.2.3.-OrganizationName
2.2.4.-CertificateEntries
2.2.5.-SecurityProfile
2.2.6.-AdditionalRootCertificateCheck
2.2.7.-MaxCertificateChainSize
2.2.8.-CertSigningWaitMinimum
2.2.9.-CertSigningRepeatTimes
//-2.3-Authorization-related
2.3.1.-AuthEnabled
2.3.2.-AdditionalInfoItemsPerMessage
2.3.3.-OfflineTxForUnknownIdEnabled
2.3.4.-AuthorizeRemoteStart
2.3.5.-LocalAuthorizeOffline
2.3.6.-LocalPreAuthorize
2.3.7.-MasterPassGroupId
2.3.8.-DisableRemoteAuthorization
//-2.4-Authorization-Cache-related
2.4.1.-AuthCacheEnabled
2.4.2.-AuthCacheAvailable
2.4.3.-AuthCacheLifeTime
2.4.4.-AuthCacheStorage
2.4.5.-AuthCachePolicy
2.4.6.-AuthCacheDisablePostAuthorize
//-2.5-Local-Authorization-List-Management-related
2.5.1.-LocalAuthListEnabled
2.5.2.-LocalAuthListEntries
2.5.3.-LocalAuthListAvailable
2.5.4.-ItemsPerMessageSendLocalList
2.5.5.-BytesPerMessageSendLocalList
2.5.6.-LocalAuthListStorage
2.5.7.-LocalAuthListDisablePostAuthorize
2.5.8.-LocalAuthListSupportsExpiryDateTime
//-2.6-Transaction-related
2.6.1.-EVConnectionTimeOut
2.6.2.-StopTxOnEVSideDisconnect
2.6.3.-TxBeforeAcceptedEnabled
2.6.4.-TxStartPoint
2.6.5.-TxStopPoint
2.6.6.-TxStartStopPoint-values
2.6.6.1.-TxStartPoint-values
2.6.6.2.-TxStopPoint-values
2.6.7.-MaxEnergyOnInvalidId
2.6.8.-StopTxOnInvalidId
//-2.7-Metering-related
2.7.1.-SampledDataEnabled
2.7.2.-SampledDataAvailable
2.7.3.-SampledDataSignReadings
2.7.4.-SampledDataTxEndedMeasurands
2.7.5.-SampledDataTxEndedInterval
2.7.6.-SampledDataTxStartedMeasurands
2.7.7.-SampledDataTxUpdatedMeasurands
2.7.8.-SampledDataTxUpdatedInterval
2.7.9.-AlignedDataEnabled
2.7.10.-AlignedDataAvailable
2.7.11.-AlignedDataMeasurands
2.7.12.-AlignedDataInterval
2.7.13.-AlignedDataSendDuringIdle
2.7.14.-AlignedDataSignReadings
2.7.15.-AlignedDataTxEndedMeasurands
2.7.16.-AlignedDataTxEndedInterval
2.7.17.-PublicKeyWithSignedMeterValue
2.7.18.-SampledDataRegisterValuesWithoutPhases
//-2.8-Reservation-related
2.8.1.-ReservationEnabled
2.8.2.-ReservationAvailable
2.8.3.-ReservationNonEvseSpecific
//-2.9-Smart-Charging-related
2.9.1.-SmartChargingEnabled
2.9.2.-SmartChargingAvailable
2.9.3.-ACPhaseSwitchingSupported
2.9.4.-ChargingProfileMaxStackLevel
2.9.5.-ChargingScheduleChargingRateUnit
2.9.6.-PeriodsPerSchedule
2.9.7.-ExternalControlSignalsEnabled
2.9.8.-NotifyChargingLimitWithSchedules
2.9.9.-Phases3to1
2.9.10.-ChargingProfileEntries
2.9.11.-LimitChangeSignificance
//-2.10-Tariff-&-Cost-related
2.10.1.-TariffEnabled
2.10.2.-TariffAvailable
2.10.3.-TariffFallbackMessage
2.10.4.-CostEnabled
2.10.5.-CostAvailable
2.10.6.-TotalCostFallbackMessage
2.10.7.-Currency
//-2.11-Diagnostics-related
2.11.1.-MonitoringEnabled
2.11.2.-MonitoringAvailable
2.11.3.-ItemsPerMessageClearVariableMonitoring
2.11.4.-ItemsPerMessageSetVariableMonitoring
2.11.5.-BytesPerMessageClearVariableMonitoring
2.11.6.-BytesPerMessageSetVariableMonitoring
2.11.7.-OfflineMonitoringEventQueuingSeverity
2.11.8.-ActiveMonitoringBase
2.11.9.-ActiveMonitoringLevel
//-2.12-Display-Message-related
2.12.1.-DisplayMessageEnabled
2.12.2.-DisplayMessageAvailable
2.12.3.-NumberOfDisplayMessages
2.12.4.-DisplayMessageSupportedFormats
2.12.5.-DisplayMessageSupportedPriorities
//-2.13-Charging-Infrastructure-related
2.13.1.-Available
2.13.2.-AvailabilityState
2.13.3.-AllowReset
2.13.4.-ConnectorType
2.13.5.-PhaseRotation
2.13.6.-SupplyPhases
2.13.7.-Power
2.13.8.-Example-Reporting-of-EVSEs-and-Connectors-via-device-model
//-2.14-ISO-15118-Related
2.14.1.-CentralContractValidationAllowed
2.14.2.-ContractValidationOffline
2.14.3.-ProtocolSupportedByEV
2.14.4.-ProtocolAgreed
2.14.5.-ISO15118PnCEnabled
2.14.6.-ISO15118V2GCertificateInstallationEnabled
2.14.7.-ISO15118ContractCertificateInstallationEnabled
2.14.8.-ISO15118RequestMeteringReceipt
2.14.9.-ISO15118SeccId
2.14.10.-ISO15118CountryName
2.14.11.-ISO15118OrganizationName
2.14.12.-ISO15118EvseId
```

01_Data/src/layers/sequelize/model/ChargingStationSecurityInfo.ts
yuvenus: for now, every charger gets the same security config (as needed). so if you have three charging stations and all three needed to send signed meter values, there would be three entries into this table but their public key would be the same 👍

02_Util/src/security/SignedMeterValuesUtil.ts
yuvenus: oooh yes i wondered this as well! i'll look into making a task and creating a separate PR for this... 🤔

02_Util/src/security/SignedMeterValuesUtil.ts
```diff
@@ -0,0 +1,209 @@
...
+
+  // https://developer.chrome.com/blog/how-to-convert-arraybuffer-to-and-from-string
+  private str2ab(str: string): ArrayBuffer {
+    const buf = new ArrayBuffer(str.length);
+    const bufView = new Uint8Array(buf);
+    for (let i = 0, strLen = str.length; i < strLen; i++) {
+      bufView[i] = str.charCodeAt(i);
+    }
+    return buf;
+  }
```
lydiazcheng, line 208: Just a tiny idea, `stringToArrayBuffer` from 'pvutils' can also be an option. The source code is the same as this piece of code. This `pvutils` lib also provides some other useful methods, like from/toBase64.
`import { stringToArrayBuffer } from 'pvutils';`

lydiazcheng (commented):
Just some small points. The code looks good!

langiole:
> This PR adds logic to process signed meter values signed with RSASSA-PKCS1-v1_5, which is one of the allowed signing algorithms provided by OCPP 2.0.1 spec.

Where in the spec is this stated?

Also, is there anywhere in the spec I can read about signed meter values to get an understanding of how it works and its purpose? Didn't really see anything in depth.

03_Modules/Transactions/src/module/module.ts
```diff
@@ -337,11 +372,33 @@ export class TransactionsModule extends AbstractModule {
...
     // TODO: Meter values can be triggered. Ideally, it should be sent to the callbackUrl from the message api that sent the trigger message
     // TODO: If sendCostUpdatedOnMeterValue is true, meterValues handler triggers cost update
     //  when it is added into a transaction
 
+    const meterValues = message.payload.meterValue;
+    const stationId = message.context.stationId;
+
+    await Promise.all(
+      meterValues.map((meterValue) =>
+        this.transactionEventRepository.createMeterValue(meterValue),
```
langiole, line 384: Does it make sense to store the meter values if we might potentially reject the request a few lines below?

02_Util/src/security/SignedMeterValuesUtil.ts
```diff
@@ -0,0 +1,209 @@
...
+      new sequelize.SequelizeChargingStationSecurityInfoRepository(
+        config,
+        logger,
+      );
+
+    this._signedMeterValuesConfiguration =
+      config.modules.transactions.signedMeterValuesConfiguration;
+  }
+
+  public async validateMeterValues(
```
langiole, line 58: A little blurb explaining what constitutes valid meter values might be nice if it is straightforward.

02_Util/src/security/SignedMeterValuesUtil.ts
```diff
@@ -0,0 +1,209 @@
...
+    stationId: string,
+    meterValues: [MeterValueType, ...MeterValueType[]],
+  ): Promise<boolean> {
+    let validMeterValues = true;
+
+    const expectPublicKey =
```
langiole, line 64:
```suggestion
    const expectedPublicKey =
```

02_Util/src/security/SignedMeterValuesUtil.ts
```diff
@@ -0,0 +1,209 @@
...
+          const signedMeterValue = sampledValue.signedMeterValue;
+
+          if (!signedMeterValue) {
+            continue;
+          }
+
+          if (
+            signedMeterValue.publicKey &&
+            signedMeterValue.publicKey.length > 0
+          ) {
```
langiole, line 83: Isn't `publicKey` required if `signedMeterValue` is present?

02_Util/src/security/SignedMeterValuesUtil.ts
```diff
@@ -0,0 +1,209 @@
...
+
+    const expectPublicKey =
+      await this._deviceModelRepository.readAllByQuerystring({
+        stationId,
+        component_name: 'OCPPCommCtrlr',
+        variable_name: 'PublicKeyWithSignedMeterValue',
+      });
+
+    for (const meterValue of meterValues) {
+      if (expectPublicKey.length > 0 && expectPublicKey[0].value !== 'Never') {
```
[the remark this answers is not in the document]
langiole: Going off of this, can also put the stuff that the for loop executes into its own method. Then, this method can look like:
```
  public async validateMeterValues(
    stationId: string,
    meterValues: [MeterValueType, ...MeterValueType[]],
  ): Promise<boolean> {
    let validMeterValues = true;

    const expectPublicKey =
      await this._deviceModelRepository.readAllByQuerystring({
        stationId,
        component_name: 'OCPPCommCtrlr',
        variable_name: 'PublicKeyWithSignedMeterValue',
      });

    if (expectPublicKey.length < 0 || expectPublicKey[0].value === 'Never') {
      return true;
    }

    for (const meterValue of meterValues) {
      for (const sampledValue of meterValue.sampledValue) {
        const signedMeterValue = sampledValue.signedMeterValue;
        if (signedMeterValue) {
          if (!this.validateSignedMeterValue(stationId, signedMeterValue)) {
            return false;
          }
        }
      }
    }

    return true;
  }
```

Although I think you'd want to update such that when

02_Util/src/security/SignedMeterValuesUtil.ts
```diff
@@ -0,0 +1,209 @@
...
+            signedMeterValue.publicKey.length > 0
+          ) {
+            const incomingPublicKeyIsValid =
+              await this.isMeterValueSignatureValid(signedMeterValue);
+
+            if (
+              this._signedMeterValuesConfiguration &&
+              incomingPublicKeyIsValid
+            ) {
+              await this._chargingStationSecurityInfoRepository.readOrCreateChargingStationInfo(
```
langiole, line 91: This is purely to make sure charging station info is created if it doesn't exist?

Asking cause we don't do anything with return value.

yuvenus:
> > This PR adds logic to process signed meter values signed with RSASSA-PKCS1-v1_5, which is one of the allowed signing algorithms provided by OCPP 2.0.1 spec.
> 
> Where in the spec is this stated?
> 
> Also, is there anywhere in the spec I can read about signed meter values to get an understanding of how it works and its purpose? Didn't really see anything in depth.

indeed it's quite a trek to find this information 😩 for specifically signed meter values, they're mentioned in the `OCPP-2.0.1_edition3_part2_specification.pdf` section `2.4. Signed Meter Values`, but the blurb leaves much to be desired 😔 in the same pdf, the section `2.7.17. PublicKeyWithSignedMeterValue` explains which variable attribute should be used to actually tell the signed meter value when to have a non-empty public key.

as for the allowed algorithms for signed messages in general, christian pointed me in the direction of the `OCPP-2.0.1_edition3_part4_ocpp-j-specification.pdf`, section `7.3. Allowed Algorithms`:

```
ES256: ECDSA using P-256 and SHA-256
RS256: RSASSA-PKCS1-v1_5 using SHA-256
RS384: RSASSA-PKCS1-v1_5 using SHA-384
```

02_Util/src/security/SignedMeterValuesUtil.ts
yuvenus: unfortunately no :( based on the section `2.7.17. PublicKeyWithSignedMeterValue` in the `OCPP-2.0.1_edition3_part2_specification.pdf`:

```
This Configuration Variable can be used to configure whether a public key needs to be sent with a signed meter
value. Note, that the field is required, so it needs to be present as an empty string when the public key is not sent.
```

langiole:
> > > This PR adds logic to process signed meter values signed with RSASSA-PKCS1-v1_5, which is one of the allowed signing algorithms provided by OCPP 2.0.1 spec.
> > 
> > 
> > Where in the spec is this stated?
> > Also, is there anywhere in the spec I can read about signed meter values to get an understanding of how it works and its purpose? Didn't really see anything in depth.
> 
> indeed it's quite a trek to find this information 😩 for specifically signed meter values, they're mentioned in the `OCPP-2.0.1_edition3_part2_specification.pdf` section `2.4. Signed Meter Values`, but the blurb leaves much to be desired 😔 in the same pdf, the section `2.7.17. PublicKeyWithSignedMeterValue` explains which variable attribute should be used to actually tell the signed meter value when to have a non-empty public key.
> 
> as for the allowed algorithms for signed messages in general, christian pointed me in the direction of the `OCPP-2.0.1_edition3_part4_ocpp-j-specification.pdf`, section `7.3. Allowed Algorithms`:
> 
> ```
> ES256: ECDSA using P-256 and SHA-256
> RS256: RSASSA-PKCS1-v1_5 using SHA-256
> RS384: RSASSA-PKCS1-v1_5 using SHA-384
> ```

Oh wow. That seems quite crazy to me but maybe it's because it's really not that complicated?

Presumably it is as simple as charger encrypts meter values with public key and CSMS decrypts with private key.

My question is how/when does this key get sent to the charger. And is this key only used for signing meter values.

Also, what the heck do the contents of a signed meter value look like? Is it just a regular meter value after being decrypted? 

And if we have a `signedMeterValue` present, presumably the other inline fields are present since they're still required?

02_Util/src/security/SignedMeterValuesUtil.ts
yuvenus: yup exactly, we want it to be created if it doesn't already exist

02_Util/src/security/SignedMeterValuesUtil.ts
langiole: Gotcha. So if public key is not sent, use the public key stored on the CSMS? Which is presumably the same public key the charger used to encrypt the message?

02_Util/src/security/SignedMeterValuesUtil.ts
```diff
@@ -0,0 +1,209 @@
...
+
+          if (
+            signedMeterValue.publicKey &&
+            signedMeterValue.publicKey.length > 0
+          ) {
+            const incomingPublicKeyIsValid =
+              await this.isMeterValueSignatureValid(signedMeterValue);
+
+            if (
+              this._signedMeterValuesConfiguration &&
```
langiole, line 88: Why would this CSMS config need to be present in order for the signed meter values to be deemed valid?

At this point, we already validated that the signed meter value sent by the charger is valid so why does it matter if we have this config present?

Also, if this config is truly required for any sort of signed meter values validation, maybe we should have it be one of the first checks we perform in the `validateMeterValues` method?

01_Data/src/layers/sequelize/model/ChargingStationSecurityInfo.ts
```diff
@@ -0,0 +1,23 @@
+import { Column, Model, Table, DataType } from 'sequelize-typescript';
+import { Namespace } from '@citrineos/base';
+
+/**
+ * Represents the security information found on a particular charging station.
+ */
+@Table
+export class ChargingStationSecurityInfo extends Model {
```
langiole, line 8: Why exactly is it required that we store this information?

00_Base/src/config/types.ts
```diff
@@ -92,6 +92,13 @@ export const systemConfigInputSchema = z.object({
       port: z.number().int().positive().default(8081).optional(),
       costUpdatedInterval: z.number().int().positive().default(60).optional(),
       sendCostUpdatedOnMeterValue: z.boolean().default(false).optional(),
+      signedMeterValuesConfiguration: z
+        .object({
+          publicKeyFileId: z.string(),
+          privateKeyFileId: z.string(),
+          signingMethod: z.string(),
```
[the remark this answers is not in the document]
yuvenus: good point, done!

02_Util/src/security/SignedMeterValuesUtil.ts
```diff
@@ -0,0 +1,209 @@
...
+            signatureBuffer,
+            signatureBuffer,
+          );
+          return true;
+        } catch (e) {
+          if (e instanceof DOMException) {
+            this._logger.warn(
+              `Error decrypting private key or verifying signature from Signed Meter Value. Error: ${JSON.stringify(e.message)}`,
+            );
+          }
```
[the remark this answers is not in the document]
yuvenus: makes sense! updated to consider both DOMException and other errors that may arise 👍

02_Util/src/security/SignedMeterValuesUtil.ts
yuvenus: AHA yes thank you! removed the "homemade" method to use this import 🙌

02_Util/src/security/SignedMeterValuesUtil.ts
langiole: @thanaParis do we know how the private key the charger uses for signing is provisioned?

The spec seems to be unclear on this.

Presumably, if it follows the same flow as provisioning new PnC certs, then we would not need to store the public cert here, and would instead store it when provisioned.

02_Util/src/security/SignedMeterValuesUtil.ts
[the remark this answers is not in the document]
yuvenus: i left it in the loop on the crazy chance that someone sent multiple meter values, but not all of them were signed -- i don't think that's possible, right?? either all meter values are signed or they're not 😆 i'll update and refactor 👍

02_Util/src/security/SignedMeterValuesUtil.ts
thanaParis: The private key will not be sent over the ocpp connection--it must end up on the charger some other way, ideally by firmware update or manual entry on the HMI.

On Citrine's side I would expect the private and public keys to be added via secure file server, so that they are encrypted at rest. They would both need to be provisioned ahead of time.

02_Util/src/security/SignedMeterValuesUtil.ts
thanaParis: I'm fine with adding enums as references, please do not type any objects or methods with these enums--the component and variable names should always be strings. This is important to maintain the flexibility that the device model operates on as a core principle. 

The enums will make it a lot easier to access/recall the known components/variables 👍

01_Data/src/layers/sequelize/repository/ChargingStationSecurityInfo.ts
```diff
@@ -0,0 +1,29 @@
...
+    return existingInfo ? existingInfo.publicKeyFileId : '';
+  }
+
+  async readOrCreateChargingStationInfo(stationId: string, publicKeyFileId: string): Promise<void> {
+    await this.readOrCreateByQuery({
+      where: {
+        stationId,
+      },
+      defaults: {
+        stationId,
```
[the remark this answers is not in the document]
yuvenus: oh wow didn't know that! updated to remove unneeded station id assignment in defaults then 👍

01_Data/src/layers/sequelize/repository/TransactionEvent.ts
```diff
@@ -324,4 +324,9 @@ export class SequelizeTransactionEventRepository extends SequelizeRepository<Tra
         return transactions[0];
       });
   }
+
+  async createMeterValue(meterValue: MeterValueType): Promise<void> {
+    const createdMeterValue = await MeterValue.create({ ...meterValue });
+    this.meterValue.emit('created', [createdMeterValue]);
+  }
```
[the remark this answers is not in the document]
yuvenus: good point, updated!

02_Util/src/security/SignedMeterValuesUtil.ts
```diff
@@ -0,0 +1,209 @@
...
+ */
+export class SignedMeterValuesUtil {
+  private readonly _fileAccess: IFileAccess;
+  private readonly _logger: Logger<ILogObj>;
+  private readonly _deviceModelRepository: IDeviceModelRepository;
+  private readonly _chargingStationSecurityInfoRepository: IChargingStationSecurityInfoRepository;
+
+  private readonly _signedMeterValuesConfiguration:
+    | SignedMeterValuesConfig
+    | undefined;
```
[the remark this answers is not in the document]
yuvenus: the linter keeps adding that in 😱 not sure why though...?

02_Util/src/security/SignedMeterValuesUtil.ts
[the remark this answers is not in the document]
yuvenus: refactored so that `expectPublicKey` is renamed to `shouldPublicKeyBeUsedForValidation` for clarity + moving it outside the for loop + refactored the logic into a helper method for readability 👍

03_Modules/Transactions/src/module/module.ts
```diff
@@ -12,16 +12,21 @@ import {
...
   CostUpdatedResponse,
   CrudRepository,
+  ErrorCode,
   EventGroup,
   GetTransactionStatusResponse,
   HandlerProperties,
   ICache,
+  IdTokenInfoType,
+  IdTokenType,
+  IFileAccess,
```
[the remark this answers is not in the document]
yuvenus: oops, this was a merge conflict i merged wrong 😆 fixed!

03_Modules/Transactions/src/module/module.ts
yuvenus: to prevent data loss we're always storing the meter value, and in theory there shouldn't consequences to storing the data even if the signature's off... hopefully 🤔

02_Util/src/security/SignedMeterValuesUtil.ts
yuvenus: a task has been created, will resolve this thread, thanks!

02_Util/src/security/SignedMeterValuesUtil.ts
yuvenus: yup makes sense -- added the proper comment docs 👍

00_Base/src/config/types.ts
```diff
@@ -334,6 +341,13 @@ export const systemConfigSchema = z
           port: z.number().int().positive().optional(),
           costUpdatedInterval: z.number().int().positive().optional(),
           sendCostUpdatedOnMeterValue: z.boolean().optional(),
+          signedMeterValuesConfiguration: z
```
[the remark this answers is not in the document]
yuvenus: 1. eventually we'll move the configs into the database so we can foreign key on a specific public key config, rather than having it stored in the system config 🤞 
2. i'm not too sure myself 😳 it seems the `systemConfigInputSchema` is used to actually instantiate the config, while the `systemConfigSchema` is used just as a type 🤔

01_Data/src/layers/sequelize/model/ChargingStationSecurityInfo.ts
yuvenus: this is stored because the publicKey isn't always sent in the signed meter value -- if it's missing, we'll need some place to refer to a previously sent public key for the charging station

01_Data/src/layers/sequelize/model/ChargingStationSecurityInfo.ts
yuvenus: aha looks like @thanaParis made the ticket, thank you!! 🙏

02_Util/src/security/SignedMeterValuesUtil.ts
yuvenus: good point! updated the name for more clarity on what `expectedPublicKey` should actually represent 👍

02_Util/src/security/SignedMeterValuesUtil.ts
yuvenus: yup exactly, if the public key is not sent in the signed meter values message, find the public key stored in the csms. 

for right now, if the public key is in the signed meter values message, we go ahead and create the key if it doesn't exist, or just confirm it exists, but don't actually do anything with the key at that time. it's mostly just a sanity check 😅

02_Util/src/security/SignedMeterValuesUtil.ts
yuvenus: i see! so basically on both the CSMS and charger side, there's some human intervention needed to make sure those keys end up in the right place 🤔 no way to ask the CSMS to pass the keys along, it seems :(

02_Util/src/security/SignedMeterValuesUtil.ts
yuvenus: we validate mostly as a sanity check to make sure some really messed up data isn't just being processed (although, we save it, so i suppose it'll still end up somewhere in our system.)

checking if the config is present is only necessary to help us create that database entry that stores the link between the charging station + the public key (file id). right now, the database entry stores the public key file id, rather than the actual public key itself, and the only way to derive the public key file id is by checking the config 😅

02_Util/src/security/SignedMeterValuesUtil.ts
langiole: @thanaParis I figured the use case `A02 - Update Charging Station Certificate by request of CSMS` might be relevant here since it details how the charger can provision itself a new pair of keys on demand:
Screenshot 2024-08-08 at 16 42 20

02_Util/src/security/SignedMeterValuesUtil.ts
```diff
@@ -0,0 +1,225 @@
...
+    const shouldPublicKeyBeUsedForValidation =
+      publicKeyFrequencyVariableAttribute.length > 0 &&
+      publicKeyFrequencyVariableAttribute[0].value !== 'Never';
+
+    if (shouldPublicKeyBeUsedForValidation) {
+      for (const meterValue of meterValues) {
+        for (const sampledValue of meterValue.sampledValue) {
+          validMeterValues =
+            validMeterValues &&
+            (await this.isSignedSampledValueValid(stationId, sampledValue));
```
langiole, line 99: Shouldn't we short circuit upon first occurrence of `false`?

I would think if one meter value is invalid, then short circuit and return false.

02_Util/src/security/SignedMeterValuesUtil.ts
langiole, line 1: @thanaParis I believe @yuvenus has the current validation logic to set up such that if public key is provided in MeterValues request, then we check to see if the public key in the request matches what is configured into our CSMS. Otherwise, we consider it as invalid.

I think this makes sense as it deters the usage of unknown public keys but wanted to make sure this wasn't an unnecessary step.

02_Util/src/security/SignedMeterValuesUtil.ts
```diff
@@ -0,0 +1,225 @@
...
+
+      if (configuredPublicKey !== signedMeterValuePublicKey) {
+        return false;
+      }
+    }
+
+    switch (signingMethod) {
+      case 'RSASSA-PKCS1-v1_5':
+        try {
+          const cryptoPublicKey = await crypto.subtle.importKey(
```
langiole, line 147: Could dump the contents of this case into its own `validateRSASignature(...)` method

02_Util/src/security/SignedMeterValuesUtil.ts
```diff
@@ -0,0 +1,225 @@
...
+            validMeterValues &&
+            (await this.isSignedSampledValueValid(stationId, sampledValue));
+        }
+      }
+    }
+
+    return validMeterValues;
+  }
+
+  private async isMeterValueSignatureValid(
```
langiole, line 107: I'm a bit conflicted on the naming since on the one hand I think it would be nice if it was named as `validateSignedMeterValue` to match the other method above which is prefixed with `validate`.

But on the other hand, I see why you went with the `is` prefix since it is a singular object being validated and not multiple.

02_Util/src/security/SignedMeterValuesUtil.ts
```diff
@@ -0,0 +1,225 @@
...
+        stationId,
+        component_name: 'OCPPCommCtrlr',
+        variable_name: 'PublicKeyWithSignedMeterValue',
+      });
+
+    const shouldPublicKeyBeUsedForValidation =
+      publicKeyFrequencyVariableAttribute.length > 0 &&
+      publicKeyFrequencyVariableAttribute[0].value !== 'Never';
+
+    if (shouldPublicKeyBeUsedForValidation) {
```
langiole, line 94: Not sure about this.

If charger is set to never send public key, then I'd think that we should still to validation but should just use the CSMS configured public key.

cc: @thanaParis

langiole: Honestly, I'm not sure we really need to fetch this variable attribute in the first place.

We can just do a check on this `publicKey` field of the SignedMeterValue to check if it is non-empty.

If it is present, then check to see if it matches the public key that the CSMS has configured, and proceed with validation. If it is not present, then just do validation.

01_Data/src/layers/sequelize/model/ChargingStationSecurityInfo.ts
langiole: Is it really necessary to store the key in the first place?

When charger sends key in the meter values, don't we reject the request if it doesn't match the key CSMS has configured?

By this logic, then the key that we store in the DB will always be what is configured in the CSMS. In which case, the entire DB will consist of redundant data.

02_Util/src/security/SignedMeterValuesUtil.ts
elliot-sabitov: In this case, if it is possible for these strings to further extend the possible values defined in the enum, then should we create a type that extends both enum and string like:
```
export type Components = ComponentsEnum | string;
```
using a type like this would allow for the code to be able to support both enums and other string values...

01_Data/src/layers/sequelize/model/ChargingStationSecurityInfo.ts
yuvenus: yup indeed currently, the key should always match what's in the CSMS, rather than needing to refer to this object. however, down the line, chargers may be able to have different configured keys, so this is really a setup for the future 🤞

02_Util/src/security/SignedMeterValuesUtil.ts
yuvenus: moved into its own helper method 👍

02_Util/src/security/SignedMeterValuesUtil.ts
yuvenus: yup makes sense since we currently only care if any of the meter values was invalid... updated to return if validMeterValues is false 👍

02_Util/src/security/SignedMeterValuesUtil.ts
langiole: @yuvenus I don't think that change got through?

02_Util/src/security/SignedMeterValuesUtil.ts
yuvenus: yup i'm always back and forth on this, for consistency though i've updated all the methods in here to be `validate...`, so this is now `validteSignedMeterValueSignature`

02_Util/src/security/SignedMeterValuesUtil.ts
yuvenus: rolling around this logic in my head, i see what you mean. i think i was using this initially to choose whether or not to do the public key processing. but if a signed meter value exists in general, we don't really care if a public key is in the message or not bc we already have logic that can check the DB for the public key instead, and we'd always want to do the validation if we have a signed meter value 🤔 

let me simplify the logic 👍

02_Util/src/security/SignedMeterValuesUtil.ts
```diff
@@ -0,0 +1,222 @@
...
+          validMeterValues = await this.validateSignedSampledValue(
+            stationId,
+            sampledValue.signedMeterValue,
+          );
+          if (!validMeterValues) {
+            return validMeterValues;
+          }
+        }
+      }
+    }
```
langiole, line 94: if u wanted to:
```suggestion
    for (const meterValue of meterValues) {
      for (const sampledValue of meterValue.sampledValue) {
        if (sampledValue.signedMeterValue) {
          if (!(await this.validateSignedSampledValue(
            stationId,
            sampledValue.signedMeterValue,
          )) {
            return false;
          }
        }
      }
    }
    return true;
```

02_Util/src/security/SignedMeterValuesUtil.ts
```diff
@@ -0,0 +1,222 @@
...
+    if (
+      !this._signedMeterValuesConfiguration?.publicKeyFileId ||
+      (publicKeyFileId &&
+        publicKeyFileId !==
+          this._signedMeterValuesConfiguration?.publicKeyFileId) ||
+      (!publicKeyFileId && incomingPublicKeyString.length === 0) ||
+      this._signedMeterValuesConfiguration?.signingMethod !== signingMethod
+    ) {
+      return false;
+    }
```
langiole, line 145: Would maybe suggest breaking this if statement up so that we can log out exactly why the signature is invalid:

```suggestion
    if (!this._signedMeterValuesConfiguration?.publicKeyFileId) {
      console.log("Invalid signature due to public key missing in system config");
      return false;
    }
    if (!publicKeyFileId) {
      console.log("Invalid signature due to public key missing from request");
      return false;
    }
    if (publicKeyFileId !== this._signedMeterValuesConfiguration?.publicKeyFileId)) {
      console.log("Invalid signature due to public key not matching system config public key");
      return false;    
    }
```

02_Util/src/security/SignedMeterValuesUtil.ts
yuvenus: good point! simplified 👍

02_Util/src/security/SignedMeterValuesUtil.ts
yuvenus: very true -- updated so that there are more explicit blocks with logging 👍

00_Base/src/config/types.ts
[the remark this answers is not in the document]
thanaParis: systemConfigInputSchema has different typing around required fields because some required fields have defaults--i.e., for the input, they aren't actually required.
systemConfigSchema is the type that's actually used by the rest of the application, so all required fields are actually required in its typing.

02_Util/src/security/SignedMeterValuesUtil.ts
thanaParis: yes, this is the desired behavior

thanaParis (approved):
Approved

PR #219, opened 2024-08-12
[Cost updated] Improve accuracy of cost calculations

falanadamian:
Previously, cost calculations were performed using standard JavaScript numbers, which introduced floating-point issues and resulted in inaccurate results. 

This PR introduces a Money and a Currency classes to better handle monetary values. The Money class is implemented using the Big.js library to ensure accuracy. All monetary operations are delegated to this class to avoid precision loss.


ChrisWeissmann (approved):

yuvenus (approved):
very cool! just curious, was there a reason to not use a library such as [currency.js](https://currency.js.org/) (also understandable to not use bc we don't want more dependencies 😬)

ChrisWeissmann (approved):

yuvenus:
> > very cool! just curious, was there a reason to not use a library such as [currency.js](https://currency.js.org/) (also understandable to not use bc we don't want more dependencies 😬)
> 
> @yuvenus good point! I looked into using a library for handling monetary values and found two potentially useful options: DineroJS and currency.js.
> 
> * currency.js was almost a perfect fit, but I couldn’t find a way to force it to round down; it always rounds up when getting the final amount, which might not be suitable for our needs (as I understand we always want to round down on final amount). If you think I missed something, feel free to check it out, maybe it's doable?
> 
> * DineroJS seems more mature, but its v1 version operates on integers, whereas our code uses decimals for major units, which could cause issues (and is not very handy). The v2 version looks promising, but it's been in alpha for over three years, and the project appears somewhat inactive. The documentation even has a warning about the alpha status (You're browsing the documentation for v2.0.0-alpha.14. Things might break!). If we were open to using an alpha release, DineroJS v2 might be a good option due to its sophistication and maturity. (I checked the source code and that was my impression)
> 
> 
> After researching, many recommend using libraries like Big.js or Decimal.js for money handling. Implementing our own Money type based on Big.js could be beneficial, as it allows us to design the API to suit our needs. I’ve written some tests, and it seems to work well so far. This approach aligns with practices in other languages, like Java’s Moneta, which is based on BigDecimal (though some prefer using longs).
> 
> What are your thoughts? Should we consider DineroJS v2 alpha, or stick to implementing our own Money based on Big.js?
> 
> Additionally, it might be worth considering a reconciliation process to periodically check for discrepancies over longer periods, such as daily or weekly, what do you think?

regarding the libraries, makes sense! better safe than sorry and implement exactly as we need. it blows my mind a little bit though that no one's really solved all the problems needed to handle currency in javascript. (perhaps it's time to release `citrineos-currency` 😆) 

regarding the reconciliation process, what discrepancies do you anticipate? is it things like the decimal precision being wrong, currencies changing, currency formats changing, etc...? i agree in general if we're maintaining essentially our own currency "library", there should be ways to make sure our currency processing isn't going out of date - but i'm not sure what this would look like in the long term 🤔

00_Base/src/assertion/assertion.ts
```diff
@@ -0,0 +1,27 @@
+export function assert(
+  predicate: boolean | (() => boolean),
+  message?: string,
+): asserts predicate {
+  switch (typeof predicate) {
+    case 'boolean': {
+      if (!predicate) {
+        throw new Error(message);
```
elliot-sabitov, line 8: What do you think about adding a custom AssertionError class for this?

elliot-sabitov (approved):

PR #220, opened 2024-08-13
Create scheduler base class

falanadamian:
This PR is focused on extracting some scheduling logic into a base class to make it more reusable in the future. An alternative approach would be to use composition instead of inheritance, but that comes with its own set of challenges:
- we might run into key collisions between clients. We could either assign each client a unique namespace or have each client use a unique prefix for their keys. Another option is to use different Scheduler (prototype instead of singleton) instance for each client.
- unit testing becomes tricky. For instance, how would we test logic encapsulated in callbacks? We could mock the scheduler to trigger the callback, but it might end up being messy?

Even though I’m not typically a fan of inheritance, it seems like a practical solution in this case. Let me know your thoughts!

### DIFF
--- a/03_Modules/Transactions/src/module/CostNotifier.ts
+++ b/03_Modules/Transactions/src/module/CostNotifier.ts
@@ -2,12 +2,9 @@ import { ITransactionEventRepository } from '@citrineos/data';
 import { AbstractModule, CallAction } from '@citrineos/base';
 import { ILogObj, Logger } from 'tslog';
 import { CostCalculator } from './CostCalculator';
+import { Scheduler } from './Scheduler';
 
-export class CostNotifier {
-  private _registry: Map<string, NodeJS.Timeout> = new Map();
-
-  private readonly _logger: Logger<ILogObj>;
-
+export class CostNotifier extends Scheduler {
   private readonly _transactionEventRepository: ITransactionEventRepository;
   private readonly _module: AbstractModule;
   private readonly _costCalculator: CostCalculator;
@@ -18,12 +15,10 @@ export class CostNotifier {
     costCalculator: CostCalculator,
     logger?: Logger<ILogObj>,
   ) {
+    super(logger);
     this._transactionEventRepository = transactionEventRepository;
     this._module = module;
     this._costCalculator = costCalculator;
-    this._logger = logger
-      ? logger.getSubLogger({ name: this.constructor.name })
-      : new Logger<ILogObj>({ name: this.constructor.name });
   }
 
   /**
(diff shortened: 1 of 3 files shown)


elliot-sabitov (approved):

elliot-sabitov:
It looks like you may need to `npm run lint-fix` to resolve the few linting issues

elliot-sabitov:
> > It looks like you may need to `npm run lint-fix` to resolve the few linting issues
> 
> @elliot-sabitov 💯 you are right, I see there are some warnings, but they don’t seem related to the new code and appear to be from the base branch? Would it be alright if we proceed with this PR now, and I can open a separate PR to address the linting issues on the rc-1.4.0 branch? This way, we can keep the changes for linting separate from the updates in this PR (easier to review + cleaner git history)? If you prefer, I can fix it here? 👍

sounds good 🙇

ChrisWeissmann (approved):
🚀

03_Modules/Transactions/src/module/CostNotifier.ts
```diff
@@ -2,12 +2,9 @@ import { ITransactionEventRepository } from '@citrineos/data';
...
 import { ILogObj, Logger } from 'tslog';
 import { CostCalculator } from './CostCalculator';
+import { Scheduler } from './Scheduler';
 
-export class CostNotifier {
-  private _registry: Map<string, NodeJS.Timeout> = new Map();
-
-  private readonly _logger: Logger<ILogObj>;
-
+export class CostNotifier extends Scheduler {
```
langiole, line 7: Questioning if we really need to extend from Scheduler.

langiole: With the current set up, if you try to unit test CostNotifier, won't you inadvertently be testing `Scheduler` due to the inheritance?

langiole: I think my preference would be that 1) there is no inheritance, and 2) `Scheduler` and `CostNotifier` each have there own unit tests.

My argument against inheritance is that we will be inadvertently testing inherited methods when we set up unit tests for any class that extends `Scheduler`. So if something in `Scheduler` breaks, then it will break unit tests for all classes that inherit from it.

ChrisWeissmann: @langiole why no inheritance? Seems like a nice generic base functionality to have, and then extend with explicit implementations. Would you rather have a Scheduler class instantiated at start up and then passed down to the CostNotifier?

langiole: That is what I was thinking. Like Damian said, composition over inheritance.

My reason against inheritance is that you will inadvertently be testing the `Scheduler` when you write unit test for any classes that inherit it (e.g. `CostNotifier`). In other words, you cannot unit test `CostNotifier` without inadvertently testing `Scheduler`.

That being said, I'm fine with keeping it since it makes sense conceptually.

elliot-sabitov (approved):

PR #221, opened 2024-08-13
Used npm ci to restore file in a reproducible way

Prev-I:
Fixes citrineos/citrineos#31

Added npm restore as shorthand to npm ci.
Changed npm i to ci in dockerbuild
Added package-lock.json to be used by npm ci and removed it from gitignore

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -196,7 +196,7 @@ dist
 
 # Files produced by build
 lib/
-package-lock.json
+#package-lock.json
 
 # Sqlite
 .sqlite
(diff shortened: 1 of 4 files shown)


elliot-sabitov:
There is an issue with committing package-lock.json into git. We currently use two dependencies in this project `bcrypt`, and `deasync` that unfortunately have platform specific implementations. This means that if you run npm install on a mac to generate the package-lock.json and commit that into git, this same package-lock.json will then fail to install on a Windows machine. Similar situation within the Docker container if the package-lock is copied from local mac for example into the linux docker container. 

Because of this, our current solution has been to avoid committing the package-lock and instead rely on a new one being generated every time on the machine that is running the npm install. An alternative would have been to create OS specific package-lock files. This way, when running on windows, npm would pull in the windows pacakge-lock, but in the container, the linux one can be used...

elliot-sabitov:
There have been discussions regarding creating multiple package-lock.json files for common OS's. There would most likely be 2-3 files that would dynamically be copied and used as part of the install process.. We have not made a decision on our approach, however, and how we handle the above issues is still yet to be determined!

PR #222, opened 2024-08-13
OCPI Reservation

lydiazcheng:
## Purpose
This PR is to support the changes in the OCPI PR (https://github.com/citrineos/citrineos-ocpi/pull/136). More details can be found in another PR.


01_Data/src/layers/sequelize/repository/Reservation.ts
```diff
@@ -62,11 +62,16 @@ export class SequelizeReservationRepository extends SequelizeRepository<Reservat
           evseId: evseDBId,
           idToken: reserveNowRequest.idToken,
           groupIdToken: reserveNowRequest.groupIdToken ?? null,
+          isActive: isActive === true ? isActive : false,
```
ChrisWeissmann, line 65: are you sure about this change? Seems a bit off to me that if I leave out an optional, it will still filter by that optional being false. I would expect to have the optional ignored in the query.

ChrisWeissmann (approved):

PR #224, opened 2024-08-15
(Release 1.4) Clear Variable Monitoring handler unit tests

yuvenus:
This PR adds unit tests for the `ClearVariableMonitoring` handler in Monitoring module by moving out the functionality into a dedicated `MonitoringService`. The intention is to setup this service for PRs down the line that can expand upon the tests and add more of the Monitoring module's handler's functionality into the service.

Note: We're at the point where we're repeating enough test-related utils (i.e. `UpdateUtils`) that we'd want to refactor and pull it out into the a folder like `02_Util` to be reused for all the modules, but that will be a separate PR 😄 

Additionally, the repository imports were updated in the module itself since we now export the repositories individually rather than under the `sequelize` export.

### DIFF
--- a/03_Modules/Monitoring/test/module/MonitoringService.test.ts
+++ b/03_Modules/Monitoring/test/module/MonitoringService.test.ts
@@ -1,4 +1,4 @@
-import { IVariableMonitoringRepository } from '@citrineos/data/dist/interfaces/repositories';
+import { IVariableMonitoringRepository } from '@citrineos/data';
 import { MonitoringService } from '../../src/module/MonitoringService';
 import {
   ClearMonitoringResultType,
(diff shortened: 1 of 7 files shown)


03_Modules/Monitoring/test/module/MonitoringService.test.ts
```diff
@@ -0,0 +1,63 @@
...
+describe('MonitoringService', () => {
+  let mockVariableMonitoringRepository: jest.Mocked<IVariableMonitoringRepository>;
+  let monitoringService: MonitoringService;
+
+  beforeEach(() => {
+    mockVariableMonitoringRepository = {
+      rejectVariableMonitoringByIdAndStationId: jest.fn(),
+    } as unknown as jest.Mocked<IVariableMonitoringRepository>;
+
+    monitoringService = new MonitoringService(mockVariableMonitoringRepository);
```
elliot-sabitov, line 18: since we are just initializing the class here, I think you can just define these outside of `beforeEach` right?

elliot-sabitov (approved):

03_Modules/Monitoring/test/module/MonitoringService.test.ts
yuvenus: indeed i think it is like this right now only because the dependencies themselves are also declared in the beforeEach 🤔 perhaps something we can look into simplifying down the line so we don't have to remake the service every time... unless jest doesn't let us do that easily 😅

PR #237, opened 2024-08-19
Feature: Unit Tests for Methods used in CertificateAuthorityService

lydiazcheng:
This PR is to add unit tests for methods which are used in CertificateAuthorityService but belongs to other scripts, including `Acme`, `Hubject` and `CertificateUtil`. The tests for CertificateAuthorityService itself will be in the next PR.

It also fix/improve few places in the source code and add test coverage command in the project as below.
Screenshot 2024-08-19 at 10 17 41 AM

There is an 404 error when using `signCertificateByExternalCA` in `Acme` so unit test is not added for this method. We use self-signed certificate currently. I created a ticket for this issue and we can plan and fix it later.


package.json
```diff
@@ -32,7 +32,8 @@
     "fresh-and-install-all": "npm run fresh && npm run install-all",
     "sync-db": "ts-node ./db.sync.ts",
     "start-everest": "npm run start-everest --prefix ./Server",
-    "test": "jest --config jest.config.ts"
+    "test": "jest --config jest.config.ts",
+    "coverage": "jest --config jest.config.ts --coverage"
```
ChrisWeissmann, line 36: nice! love it!

Server/src/config/envs/local.ts
```diff
@@ -81,6 +81,7 @@ export function createLocalConfig() {
         host: '0.0.0.0',
         port: 8055,
         generateFlows: false,
+        token: '-ssaT85n4S-wVD21LKOCDwvXN5PtnJc0',
```
ChrisWeissmann, line 84: Is this token needed?

02_Util/test/certificate/client/Acme.test.ts
```diff
@@ -0,0 +1,121 @@
...
+            name: 'acme',
+            acme: {
+              env: 'staging',
+              accountKeyFilePath: faker.lorem.word(),
+            },
+          },
+        },
+      },
+    } as any;
+    logger = new Logger<ILogObj>({ name: expect.getState().currentTestName });
```
ChrisWeissmann, line 56: I think you could leave out the logger even. the Acme class instantiates it as it's optional.

02_Util/test/certificate/client/Hubject.test.ts
```diff
@@ -0,0 +1,199 @@
...
+            hubject: {
+              baseUrl: mockBaseURL,
+              tokenUrl: mockTokenURL,
+              isoVersion: mockISOVersion,
+            },
+          },
+        },
+      },
+    } as any;
+    logger = new Logger<ILogObj>({ name: expect.getState().currentTestName });
```
ChrisWeissmann, line 40: In theory you don't need the logger I believe

ChrisWeissmann (approved):
Just some minor nitpicks

lydiazcheng:
> Just some minor nitpicks

Hi Christian, thanks a lot for helping review this PR!

Server/src/config/envs/local.ts
lydiazcheng: Yes. This static token is used when core creats directus client. It is here: https://github.com/citrineos/citrineos-core/blob/823849c04e2a58212e9facff946496cdde218f5d/02_Util/src/util/directus.ts#L42 Then directus client works when, e.g., upload files. I updated it in docker config before but forgot local. 😞  https://github.com/citrineos/citrineos-core/blob/823849c04e2a58212e9facff946496cdde218f5d/Server/docker-compose.yml#L84

PR #239, opened 2024-08-19
fix: deasync server initialization hanging

elliot-sabitov:
fix: ensuring that deasync is only used in constructors and not within a promise which seems to cause deasync to hang.

Screenshot shows server starting up properly after adjustments:
Screenshot 2024-08-19 at 2 37 33 PM


yuvenus (approved):
pulled down branch on my windows machine, server spins up and websocket connects no problem 👀🤞

lydiazcheng (approved):
Shall we create 1.4.2 for this fix?

PR #242, opened 2024-08-20
Network connection authenticator tests

falanadamian:
In this PR, I have added unit tests for the network connection Authenticator and made one logic update: previously, the system considered variable statuses other than "Rejected" during authentication. I have modified this to only consider "Accepted" statuses, which seems to be in line with OCPP specs:

**(A01.FR.03)**
precondition: The Charging Station responds to this SetVariablesRequest with a SetVariablesResponse with status _Accepted_. 
then: The CSMS SHALL assume that the authorization key change was successful, and no longer accept the credentials previously used by the Charging Station

**(A01.FR.04)**
precondition: The Charging Station responds to this SetVariablesRequest with a SetVariablesResponse with _status other than Accepted_
then: The CSMS SHALL assume that the Charging Station has NOT changed the password. Therefore the CSMS SHALL keep accepting the old credentials.

Additionally, maybe we could consider using the "value" field of the BasicAuthPassword variable directly for password extraction, instead of relying on status updates? I believe this would achieve the same result, but I wouldd appreciate feedback on whether this approach might miss any edge cases?


thanaParis:
> Additionally, maybe we could consider using the "value" field of the BasicAuthPassword variable directly for password extraction, instead of relying on status updates? I believe this would achieve the same result, but I wouldd appreciate feedback on whether this approach might miss any edge cases?

This is a good idea! We will be moving away from the VariableStatus entity in a future release anyways.

00_Base/src/interfaces/router/Authenticator.ts
```diff
@@ -6,7 +6,7 @@ export interface IAuthenticator {
   authenticate(
     allowUnknownChargingStations: boolean,
     identifier: string,
-    username?: string,
```
thanaParis, line 9: these need to stay optional. security profile 0 exists and may still want to exclude unknown chargers. security profile 3 also exists and does not use a username/password.

02_Util/src/networkconnection/authenticator/Authenticator.ts
```diff
@@ -36,40 +37,50 @@ export class Authenticator implements IAuthenticator {
   async authenticate(
     allowUnknownChargingStations: boolean,
     identifier: string,
-    username?: string,
```
thanaParis, line 39: as mentioned above, these need to stay optional

02_Util/src/networkconnection/authenticator/Authenticator.ts
```diff
@@ -36,40 +37,50 @@ export class Authenticator implements IAuthenticator {
...
-    } else if (await this._cache.get(identifier, CacheNamespace.Connections)) {
+    }
+
+    if (await this._isAlreadyConnected(identifier)) {
       this._logger.warn(
         'New connection attempted for already connected identifier',
         identifier,
       );
       return false;
-    } else if (username && password) {
```
thanaParis, line 56: for extra security, we could check the security profile here instead of just assuming (i.e., SP1 and SP2 *must* have these values)? I think that may happen already in the Network connection class though

02_Util/test/networkconnection/authenticator/Authenticator.test.ts
thanaParis, line 1: excellent!

thanaParis (changes requested):
a small change is needed for the username/password optional

02_Util/src/util/RequestOperations.ts
```diff
@@ -0,0 +1,22 @@
+import { IncomingMessage } from 'http';
+import { splitOnce } from './StringOperations';
+
+export function extractBasicCredentials(req: IncomingMessage): {
```
ChrisWeissmann, line 4: We can probably then remove the private method in `WebsocketNetworkConnection:286` since this seems to duplicate.

Add some doc string pls.

02_Util/src/util/StringOperations.ts
```diff
@@ -0,0 +1,6 @@
+export function splitOnce(value: string, separator: string): [string, string?] {
```
ChrisWeissmann, line 1: This seems to duplicate `WebsocketNetworkConnection:306` can we remove the private method then?

ChrisWeissmann (approved):

thanaParis (approved):
approved

PR #252, opened 2024-08-21
[WIP] Feature/local auth list

thanaParis:


03_Modules/Configuration/src/module/api.ts
```diff
@@ -386,6 +386,42 @@ export class ConfigurationModuleApi
     };
   }
 
+  @AsDataEndpoint(
+      Namespace.LocalAuthList,
+      HttpMethod.Get,
+      ChargingStationKeyQuerySchema,
+      UpdateLocalAuthListSchema,
```
ChrisWeissmann, line 393: You are passing him a schema that is being interpreted as a body. But body's aren't allowed on `GET` probably a mix up with the `PUT` below. Also are you sure about the query schema here? Seems we wouldn't want the objects around idToken to be required? Only stationId.

ChrisWeissmann (commented):
I think you need to still add the db models to the sequelize instance in Server.
Just so I understand the releatonships between the models.
You search for all tokens based on idToken and then take the databaseId and create lots of new objects that tie the stationId to the idToken via the databaseId.

What I don't quite get is around the version number. Since we have that and are increasing that when an update happens, should we actually be destroying previous data or just go off latest version number?

thanaParis:
Closing in favor of https://github.com/citrineos/citrineos-core/pull/257
@ChrisWeissmann these concerns are mitigated in the above

PR #255, opened 2024-08-30
Added deepEqual method for object comparison

thanaParis:
Adding deepEqual method for object comparison.

### DIFF
--- a/00_Base/src/assertion/assertion.ts
+++ b/00_Base/src/assertion/assertion.ts
@@ -25,3 +25,25 @@ export function assert(
 export function notNull(object: any): boolean {
   return object !== undefined && object !== null;
 }
+
+export function deepEqual(obj1: any, obj2: any): boolean {
+  if (obj1 === obj2) return true;
+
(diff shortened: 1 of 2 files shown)


thanaParis:
> I noticed there aren't any usages of this function yet; just curious, where do we plan to use it?

It's here:
https://github.com/citrineos/citrineos-core/pull/257/files#diff-97a69a95a94e9566e5d92764e15dd22f7b41999023d83506fee782b4fc4c28b8R61

> Also, I wanted to confirm my understanding: the intention is for directionality to work such that
`deepDirectionalEqual({}, {a: 1, b: 2}) returns true`
but
`deepDirectionalEqual({a: 1, b: 2}, {}) returns false`
correct?

Exactly!

> Also, it looks like there might be a risk of infinite recursion with circular references - should we address that, or is it acceptable in this case?

a good point. I've adjust the method to be able to avoid this.

ChrisWeissmann (approved):

PR #256, opened 2024-09-06
WIP - feat: Operator UI adjustments

elliot-sabitov:
feat: adjusting to use enum instead of const enums so that we can largely reuse the same objects in other components in this case operator UI


lydiazcheng (changes requested):
Hi Elliot, this file is generated by the `json-schema-processor.js`. We will need to update this file otherwise changes will be overwritten when running this script next time.
```
      fs.writeFileSync(
        `./src/ocpp/model/enums/index.ts`,
        licenseComment +
          Object.values(globalEnumDefinitions).sort().join('\n\n') +
          '\n',
      );
```
```
function extractEnums(ts) {
  const pattern =
    /^\/\*\*\s*\n([^\*]|(\*(?!\/)))*\*\/\nexport const enum (\w)* {(\s|\w|-|\.|=|'|,)*}\n*/gm;
```
It seems we intend to do have a const enum. @thanaParis might know more about this.

elliot-sabitov:
> Hi Elliot, this file is generated by the `json-schema-processor.js`. We will need to update this file otherwise changes will be overwritten when running this script next time.
> 
> ```
>       fs.writeFileSync(
>         `./src/ocpp/model/enums/index.ts`,
>         licenseComment +
>           Object.values(globalEnumDefinitions).sort().join('\n\n') +
>           '\n',
>       );
> ```
> 
> ```
> function extractEnums(ts) {
>   const pattern =
>     /^\/\*\*\s*\n([^\*]|(\*(?!\/)))*\*\/\nexport const enum (\w)* {(\s|\w|-|\.|=|'|,)*}\n*/gm;
> ```
> 
> It seems we intend to do have a const enum. @thanaParis might know more about this.

ah this is a great point! Perhaps we can adjust the processor? We had similar instances in OCPI where these enums were giving us trouble if there were `const enum`. If we include them as enums we can potentially reuse them in more places!

lydiazcheng (approved):

thanaParis:
Closed in favor of https://github.com/citrineos/citrineos-core/pull/260

PR #257, opened 2024-09-09
Feature/local auth list certification

thanaParis:
Adding support for Local Auth List OCPP 2.0.1 Certification Profile use & test cases.


01_Data/src/layers/sequelize/repository/LocalAuthList.ts
```diff
@@ -0,0 +1,243 @@
...
+                        },
+                    },
+                    {
+                        model: IdTokenInfo,
+                        include: [IdToken],
+                    },
+                ],
+            });
+            if (!auth) {
+                throw new Error(`Authorization not found for ${JSON.stringify(authData)}, invalid SendLocalListRequest (create necessary Authorizations first)`);
```
elliot-sabitov, line 85: I think this throw may be unhandled in the map promise, worth a double check but I think here you are throwing the error within a dynamic async function that you are passing into .map. The map will end up returning a list of promises which will not be executed in this case until further down when there is an await present on line 111. And so there is no explicit handling of the promises to ensure that any errors can be caught which I think what may cause the throws to result in unhandled rejections, and those errors may go unnoticed

03_Modules/EVDriver/src/module/LocalAuthListService.ts
```diff
@@ -0,0 +1,111 @@
...
+    }
+
+    const itemsPerMessageSendLocalList =
+      await this.getItemsPerMessageSendLocalListByStationId(stationId)
+      || authorizations.length;
+
+    return this.generateSendLocalListRequestsFromMaxLengthAndAuthorizationsAndStationId(itemsPerMessageSendLocalList, authorizations, stationId);
+  }
+
+  async *generateSendLocalListRequestsFromMaxLengthAndAuthorizationsAndStationId(itemsPerMessageSendLocalList: number, authorizations: Authorization[], stationId: string): AsyncGenerator<SendLocalListRequest> {
```
elliot-sabitov, line 69: typo?

03_Modules/EVDriver/src/module/module.ts
```diff
@@ -701,6 +724,43 @@ export class EVDriverModule extends AbstractModule {
...
+    if (!sendLocalListRequest) {
+      this._logger.error(`Unable to process SendLocalListResponse. SendLocalListRequest not found for StationId ${stationId} by CorrelationId ${message.context.correlationId}.`);
+      return;
+    }
+
+    const sendLocalListResponse = message.payload;
+
+    switch (sendLocalListResponse.status) {
+      case SendLocalListStatusEnumType.Accepted:
+        this._localAuthListRepository.createOrUpdateLocalListVersionFromStationIdAndSendLocalList(stationId, sendLocalListRequest);
```
elliot-sabitov, line 741: I am not sure if you intentionally left out await here but without this the handler may complete successfully even if the DB request fails

03_Modules/EVDriver/src/module/LocalAuthListService.ts
thanaParis: removed!

01_Data/src/layers/sequelize/repository/LocalAuthList.ts
thanaParis: good catch, fixed!

03_Modules/EVDriver/src/module/module.ts
thanaParis: fixed!

03_Modules/EVDriver/src/module/LocalAuthListService.ts
```diff
@@ -0,0 +1,124 @@
...
+      throw new Error('Could not get max local auth list entries, required by D01.FR.12');
+    } else if (newLocalAuthListLength > maxLocalAuthListEntries) {
+      throw new Error(`Updated local auth list length (${newLocalAuthListLength}) will exceed max local auth list entries (${maxLocalAuthListEntries})`);
+    }
+
+    const itemsPerMessageSendLocalList =
+      await this.getItemsPerMessageSendLocalListByStationId(stationId)
+      || (sendLocalListRequest.localAuthorizationList ? sendLocalListRequest.localAuthorizationList?.length : 0);
+
+    if (itemsPerMessageSendLocalList && sendLocalListRequest.localAuthorizationList && itemsPerMessageSendLocalList < sendLocalListRequest.localAuthorizationList.length) {
```
ChrisWeissmann, line 52: Should we make a ticket to automate this instead of throwing? Seems we could handle it in an automated fashion

03_Modules/EVDriver/src/module/LocalAuthListService.ts
```diff
@@ -0,0 +1,124 @@
...
+    });
+    const sendLocalList = await this.createSendLocalListFromStationIdAndRequestAndCurrentVersion(stationId, correlationId, sendLocalListRequest, localListVersion);
+
+    const newLocalAuthListLength = await this.countUpdatedAuthListFromRequestAndCurrentVersion(sendLocalList, localListVersion);
+    // TODO If Device Model is updated to allow different variable characteristics for the same variable per station, then we need to update this
+    const maxLocalAuthListEntries = await this.getMaxLocalAuthListEntries();
+    if (!maxLocalAuthListEntries) {
+      throw new Error('Could not get max local auth list entries, required by D01.FR.12');
+    } else if (newLocalAuthListLength > maxLocalAuthListEntries) {
+      throw new Error(`Updated local auth list length (${newLocalAuthListLength}) will exceed max local auth list entries (${maxLocalAuthListEntries})`);
```
ChrisWeissmann, line 45: If this happens, don't we need to rollback? Seems we already persisted the new auth tokens, meaning our database is now lying about what is on the charger.

03_Modules/EVDriver/src/module/LocalAuthListService.ts
```diff
@@ -0,0 +1,124 @@
...
+    const localListVersion = await this._localAuthListRepository.readOnlyOneByQuery({
+      where: {
+        stationId: stationId,
+      },
+      include: [LocalListAuthorization],
+    });
+    const sendLocalList = await this.createSendLocalListFromStationIdAndRequestAndCurrentVersion(stationId, correlationId, sendLocalListRequest, localListVersion);
+
+    const newLocalAuthListLength = await this.countUpdatedAuthListFromRequestAndCurrentVersion(sendLocalList, localListVersion);
+    // TODO If Device Model is updated to allow different variable characteristics for the same variable per station, then we need to update this
```
ChrisWeissmann, line 40: Can we change this away from a TODO?

01_Data/src/layers/sequelize/repository/Authorization.ts
```diff
@@ -131,46 +131,9 @@ export class SequelizeAuthorizationRepository extends SequelizeRepository<Author
...
-    const idTokenInclude: Includeable[] = [];
-    if (value.idToken.additionalInfo) {
-      idTokenInclude.push(AdditionalInfo);
-    }
-    include.push({ model: IdToken, include: idTokenInclude });
-    return { include };
-  }
 
-  private async _updateIdToken(value: IdTokenType, transaction?: Transaction): Promise<IdToken> {
+  async updateIdToken(value: IdTokenType, transaction?: Transaction): Promise<IdToken> {
```
ChrisWeissmann, line 136: Why make it public instead of private? Still used only inside the class

03_Modules/EVDriver/src/module/LocalAuthListService.ts
thanaParis: No, the SendLocalList was created but the LocalListVersion remains unmodified. It only is updated when the SendLocalListResponse arrives.

01_Data/src/layers/sequelize/repository/Authorization.ts
thanaParis: Oops, didn't mean to keep that change. Fixed!

03_Modules/EVDriver/src/module/LocalAuthListService.ts
thanaParis: To clarify further, LocalListAuthorizations are made here, but they are only associated with the SendLocalList entry, which will later be used to update the actual LocalListVersion for the station when the SendLocalListResponse is processed.

There is still a concern for "orphaned" entries--SendLocalList entries that will never again be referenced. I don't see that concern (data consolidation) as being in scope for this module.

03_Modules/EVDriver/src/module/LocalAuthListService.ts
thanaParis: I avoided that here because of the version number. Say some upstream service pushes a SendLocalListRequest with version number 2, they may (reasonably) assume that if it's successful the version number is 2. However, if the message needed to be split up, the version number will be higher in accordance with however many messages ended up being sent.

"Hidden" functionality in the MessageAPI, like splitting up SetVariables messages in the Monitoring module, is acceptable to me when the only potential downside is partial success--here, the version number ending up out of sync could cause bigger issues and returning it in the response doesn't align with the idea that all Message API endpoints are supposed to work the same.

03_Modules/EVDriver/src/module/LocalAuthListService.ts
thanaParis: What did you have in mind? This being a TODO will make it easier to spot cases that need to be updated when we finally overhaul the DeviceModel.

01_Data/src/layers/sequelize/model/Authorization/LocalListVersion.ts
```diff
@@ -0,0 +1,28 @@
...
+@Table
+export class LocalListVersion extends Model {
+    static readonly MODEL_NAME: string = Namespace.LocalListVersion;
+
+    @Column({
+      unique: true,
+    })
+    declare stationId: string;
+    
+    @Column(DataType.DECIMAL)
```
lydiazcheng, line 21: Should `versionNumber` be integer based on the protocol in part2, 1.23.2? Same question for the versionNumber in SendLocalList table.

03_Modules/EVDriver/src/module/api.ts
```diff
@@ -137,6 +139,19 @@ export class EVDriverModuleApi
       );
   }
 
+  @AsDataEndpoint(
+    Namespace.LocalListVersion,
+    HttpMethod.Get,
+    ChargingStationKeyQuerySchema,
+  )
+  async getLocalAuthorizationList(
```
lydiazcheng, line 147: How about renaming this to `getLocalAuthorizationListVersion`?

lydiazcheng (approved):
The PR looks good. Only minor points from my side. And it would be perfect to solve the lint issues.

01_Data/src/layers/sequelize/model/Authorization/LocalListVersion.ts
thanaParis: Good point, fixed!

03_Modules/EVDriver/src/module/api.ts
thanaParis: That makes more sense. I couldn't name it 'getLocalListVersion' since that's already a message api endpoint, but this works. Fixed!

03_Modules/EVDriver/src/module/LocalAuthListService.ts
ChrisWeissmann: My understanding was to get away from TODOs and find a different method of pointing out things that will be improved. It's also a bit misleading right now. If I want to resolve the TODO I can't really yet. If we make it a different comment that we can make searchable that would already be fine by me. E.g. `DEVICE MODEL UPDATE: required action...`

PR #260, opened 2024-09-11
Feat/normal enums over const enums

elliot-sabitov:
- feat: adjusting our json-schema-processor to create normal TS enums instead of const enums to allow us to export these enums in other places that may want to use the enum in a decorator which is not available with const enums.
- chore: ran schema processor to make the adjustments


lydiazcheng (approved):

00_Base/json-schema-processor.js
```diff
@@ -60,7 +60,10 @@ fs.readdir(path, (error, files) => {
               `',\n  passwordString = 'passwordString` +
               enumDefinition.substring(lastLineIndex);
           }
-          globalEnumDefinitions[enumName] = enumDefinition;
+          globalEnumDefinitions[enumName] = enumDefinition.replaceAll(
+            'export const enum',
```
thanaParis, line 64: There's a better way to do this. 
See: https://github.com/bcherny/json-schema-to-typescript?tab=readme-ov-file#options
We can set enableConstEnums to false in the jsts.compile call around line 140

thanaParis:
@elliot-sabitov PR has been updated, please review

elliot-sabitov:
> @elliot-sabitov PR has been updated, please review

Looks good!

PR #261, opened 2024-09-11
Feature: Enable ESM support in 00_Base

elliot-sabitov:
feat: enabling 00_Base to build for ESM which is used by many common tools such as Vite to ensure that users that need ESM support can more easily import contents from 00_Base.

** Screenshot after running `npm run build` in root:
Screenshot 2024-09-11 at 12 34 09 PM


thanaParis (approved):
Approved! To be released with 1.4.3

PR #271, opened 2024-09-19
feat: enable cors

elliot-sabitov:
feat: enabling cors to ensure requests to data endpoints can be made successfully from browsers (particularly applicable to Operator UI)


Server/src/index.ts
```diff
@@ -127,6 +128,12 @@ export class CitrineOSServer {
     this._server =
       server || fastify().withTypeProvider<JsonSchemaToTsProvider>();
 
+    // enable cors
+    this._server.register(cors, {
+      origin: true, // This can be customized to specify allowed origins
+      methods: ['GET', 'POST', 'PUT', 'DELETE'], // Specify allowed HTTP methods
```
yuvenus, line 134: no PATCH? (if that's even a valid method in this case)

thanaParis: I don't think we have any PATCHes right now

thanaParis (approved):
approved

PR #272, opened 2024-09-19
chore: lint-fix

elliot-sabitov:


01_Data/src/layers/sequelize/repository/LocalAuthList.ts
```diff
@@ -3,196 +3,209 @@
...
+          `Authorization group idToken in SendLocalListRequest ${JSON.stringify(authData.idTokenInfo.groupIdToken)} does not match group idToken in database ${JSON.stringify(auth.idTokenInfo?.groupIdToken)} (update the group idToken first)`,
+        );
+      }
+      // While new IdTokens will NOT be created or newly associated via message api, idTokenInfo can be allowed to be unique for the local auth list
+      const localListAuthIdTokenInfo = await IdTokenInfo.create({
+        ...authData.idTokenInfo,
+        groupIdTokenId: auth.idTokenInfo?.groupIdTokenId,
+      });
+
+      const { id, _idTokenInfoId, _idTokenInfo, ...authorizationFields } = auth;
```
ChrisWeissmann, line 75: @elliot-sabitov why the `_` prefix? This breaks the build. I have a commit I can push to fix this. But wanted to ask if this was intentional.

elliot-sabitov: not intentional, I will adjust, this is an error, looks like those two fields arent being used and lint fails so I incorrectly prefixed with underscore by mistake, great catch thank you so much! 😮‍💨 Not sure how I missed that

ChrisWeissmann (approved):

PR #275, opened 2024-09-27
Charging Station Sequences

falanadamian:
In this PR, I have implemented a serial sequence generation per Charging Station and type.

This implementation is concurrency-safe and should maintain good performance. If necessary, we can consider introducing a caching layer that generates multiple sequence values in a single SQL query. This would involve incrementing the sequence value by a specified amount rather than just 1, allowing us to store these IDs in application memory for faster access.

I couldn't find any specific guidelines in the documentation regarding the patterns for generating IDs so I believe this should suffice.

I would appreciate any advice on how to verify if the generated IDs meet the standards set by OCTT?


thanaParis (approved):
Looks good!

PR #276, opened 2024-09-27
Request ID generation

falanadamian:
In this PR, I have replaced the old random generation method for requestId with a new implementation that relies on database table.


thanaParis (approved):
Approved

PR #279, opened 2024-10-01
Extend charging station sequence types

falanadamian:
As per Thana's comment (https://github.com/citrineos/citrineos-core/pull/276#discussion_r1781703155) I have extended charging station sequence types so that we have one sequence type per message/ action type (instead of using one sequence for all messages). @thanaParis I hope this is what you had in mind! 😄


thanaParis (approved):
Approved

PR #284, opened 2024-10-19
Feature/persist installed certificate hashes

elliot-sabitov:
- feat: adjusting implementation such that a GetInstalledCertificateIds message is triggered after a successful InstallCertificate response is received from the charger. The GetInstalledCertificateIds body that is received from the charger will now also be persisted in the new InstalledCertificate table to keep track of the certificates that are installed on the charger allowing easy access to issuerName and issuerKey hash values. Ensuring that a successful DeleteCertificates response from the charger also triggers the GetInstalledCertificateIds to ensure that our Installed Certificates table data is refreshed.
- feat: ensuring that we persist issuerKey of the certs that are generated in certificateChain request.


03_Modules/Certificates/src/module/module.ts
```diff
@@ -312,27 +323,131 @@ export class CertificatesModule extends AbstractModule {
...
-  ): void {
+  ): Promise<void> {
     this._logger.debug('GetInstalledCertificateIds received:', message, props);
+    const certificateHashDataList: CertificateHashDataChainType[] =
+      message.payload.certificateHashDataChain!;
+    // persist installed certificate information
+    if (certificateHashDataList && certificateHashDataList.length > 0) {
+      // delete previous hashes for station
+      try {
+        await this._installedCertificateRepository.deleteAllByQuery({
```
lydiazcheng, line 361: We cannot always delete all certificates in a station because user can send the GetInstalledCertificateIdsRequest to get only one or few GetCertificateIdUseEnumTypes, e.g., V2GRootCertificate. How about finding all certificate types inside the `certificateHashDataList` at first and only delete these types of certificates?

03_Modules/Certificates/src/module/module.ts
```diff
@@ -312,27 +323,131 @@ export class CertificatesModule extends AbstractModule {
...
+    } else if (
+      message.payload.status === GetInstalledCertificateStatusEnumType.NotFound
+    ) {
+      // charger has no certs
+      this._logger.info(
+        'No certs found on charger, will delete for station',
+        message.context.stationId,
+      );
+      try {
+        await this._installedCertificateRepository.deleteAllByQuery(
```
lydiazcheng, line 408: Here is the same reason as in the previous comment, so we cannot simply delete all installed certificates in db. 

Another idea I have is that we can add a new optional field `certificateType` in this `CallMessage` table. Whenever `GetInstalledCertificateIdsRequest` is called, we stores the correlation id and certificateTypes in this `CallMessage` table. Then when getting response, we can retrieve the CallMessage entity to know what types of certificate are requested. An example is here: https://github.com/citrineos/citrineos-core/blob/f0bda14d4aed4b9f6ec800e4407119ca072a31c3/03_Modules/EVDriver/src/module/api.ts#L289

01_Data/src/layers/sequelize/model/Certificate/InstalledCertificate.ts
```diff
@@ -0,0 +1,45 @@
+import { GetCertificateIdUseEnumType, Namespace } from '@citrineos/base';
+import { Column, DataType, ForeignKey, Model, Table } from 'sequelize-typescript';
+import { ChargingStation } from '../Location';
+
+@Table
+export class InstalledCertificate extends Model {
```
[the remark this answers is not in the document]
elliot-sabitov: added

01_Data/src/layers/sequelize/model/Certificate/InstalledCertificate.ts
```diff
@@ -0,0 +1,45 @@
...
+    type: DataType.STRING(36),
+    allowNull: false,
+  })
+  declare stationId: string;
+
+  @Column({
+    type: DataType.STRING,
+    allowNull: false,
+  })
+  declare hashAlgorithm: string;
```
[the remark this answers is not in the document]
elliot-sabitov: adjusted

03_Modules/Certificates/src/module/api.ts
```diff
@@ -605,6 +605,7 @@ export class CertificatesModuleApi
     const certObj = new jsrsasign.X509();
     certObj.readCertPEM(certPem);
     certificateEntity.issuerName = certObj.getIssuerString();
+    certificateEntity.issuerKey = KEYUTIL.getPEM(certObj.getPublicKey());
```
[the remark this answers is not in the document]
elliot-sabitov: removed

03_Modules/Certificates/src/module/module.ts
elliot-sabitov: for now adjusting module to not automatically trigger GetInstalledCertificateIds, and when GetInstalledCertificateIds payload comes in from the charger, deleting only installed certificate records with matching certificate types.

03_Modules/Certificates/src/module/module.ts
elliot-sabitov: for now adjusting module to not automatically trigger GetInstalledCertificateIds, and when GetInstalledCertificateIds payload comes in from the charger, deleting only installed certificate records with matching certificate types.

lydiazcheng (approved):
Hi Elliot, why we decide not to send get certificate ids request in the delete cert response? I think that one is still helpful.

elliot-sabitov:
> Hi Elliot, why we decide not to send get certificate ids request in the delete cert response? I think that one is still helpful.

For the OCA Certification, it looks like the get cert ids message will be triggered manually. In order to simplify the setup for now, we decided to take this out

PR #285, opened 2024-10-23
Feature: LatestStatusNotifications Table

elliot-sabitov:
- feat: created LatestStatusNotification model and adjusting logic such that when a new status notification is added for a station we also update the LatestStatusNotifications table ensuring that the latest record for the evse/connector is saved in the DB


thanaParis (approved):
Approved

PR #287, opened 2024-10-28
fix: TransactionEvent failing authorization

elliot-sabitov:
- fix: adjusting concurrent check to only happen after authorizations have been applied to ensure that eventType = Updated does not result in unknown status
- 
Screenshot 2024-10-28 at 4 57 29 PM


thanaParis (approved):
approved

PR #289, opened 2024-11-04
Feature - Adding type definitions to openapi spec

elliot-sabitov:
Intended to resolve https://github.com/citrineos/citrineos/issues/49

- feat: adjusted implementation such that the Swagger UI and the OpenAPI spec include type definitions and $refs are used
- feat: adding $ids to schemas where it was missing and adjusting QuerySchema to take in the ID / name as the first parameter and apply it to the schema that is created
- feat: removing seemingly no longer necessary OcppTransformObject and instead creating buildLocalReference refResolved to be used to ensure that the generated schemas use the title/id as the name rather then generated "def-n" that is used by default

Screenshot 2024-11-04 at 12 54 12 PM
Screenshot 2024-11-04 at 12 54 05 PM
Screenshot 2024-11-04 at 12 53 57 PM
Screenshot 2024-11-04 at 12 53 52 PM


00_Base/src/interfaces/api/AbstractModuleApi.ts
```diff
@@ -263,13 +264,118 @@ export abstract class AbstractModuleApi<T extends IModule>
...
+  };
+
+  protected registerSchema = (
+    fastifyInstance: FastifyInstance,
+    schema: any,
+  ): object | null => {
+    try {
+      let id = schema['$id'];
+      if (!id) {
+        console.error('Could not register schema because no ID', schema);
```
lydiazcheng, line 320: How about using `this._logger.error` instead?

lydiazcheng (approved):
The code changes look great!

thanaParis (approved):
Approved!

PR #290, opened 2024-11-05
Feature/server network profiles

thanaParis:
Adding management of 'Server Network Profiles' to Citrine.
These represent the Websocket Server Config objects used to manage Citrine's websocket servers.
They can be linked to SetNetworkProfile messages via ChargerNetworkProfile objects.
Server Network Profiles updated from config on every startup using `Server/`

'SetNetworkProfile' objects added to track SetNetworkProfileRequests.

@AsMessageApi decorator extended in order to allow custom query params in addition to the 3 defaults. SetNetworkProfile message api extended with 'websocketServerConfigId' in order to associate message with Citrine's websocket servers. Since SetNetworkProfileRequest details usually refer to cloud infrastructure, mapping them to Citrine's internal representation of its websocket servers must be done explicitly like this and cannot be inferred (as hosts & ports & configuration slots will not match).

New AuthenticationFilter added for test case TC_A_19_CSMS. This will reject connection attempts from charging stations whose OCPPCommCtrlr.NetworkConfigurationPriority value only includes references to higher security profiles than the one the charging station is trying to connect with.
Notes: This filter does not attempt to work if any of the configured NetworkProfiles are not matched with Citrine's internal websocket server configurations, in order to avoid charger lockout. Although this feature is required by the specification, in the field this will only occur when a charger either breaches the protocol by attempting to connect with a Network Profile it doesn't have configured or when a charger has its configuration adjusted outside of OCPP communication. Due to the former being unlikely and the latter requiring human intervention, it is unwise to be overly prescriptive and risk unnecessary charger lockout.


01_Data/src/layers/sequelize/model/Location/SetNetworkProfile.ts
```diff
@@ -0,0 +1,84 @@
...
+import { ChargingStation } from './ChargingStation';
+import { ServerNetworkProfile } from './ServerNetworkProfile';
+
+/**
+ * The CallMessage model can be extended with new optional fields,
+ * e.g. chargingProfileId, for other correlationId related lookups.
+ */
+@Table
+export class SetNetworkProfile extends Model {
+    static readonly MODEL_NAME: string = 'CallMessage';
```
lydiazcheng, line 17: How about implements `NetworkConnectionProfileType` for this class and add it in `namespace.ts`? Then use it as the model name here? Or we just replace with 'SetNetworkProfile' for model name.

01_Data/src/layers/sequelize/util.ts
```diff
@@ -140,6 +143,7 @@ export class DefaultSequelizeInstance {
         Reservation,
         SalesTariff,
         SecurityEvent,
+        ServerNetworkProfile,
```
lydiazcheng, line 146: `SetNetworkProfile` seems missing?

elliot-sabitov (approved):

elliot-sabitov:
Can we fix lint 🙏

03_Modules/Configuration/src/module/api.ts
```diff
@@ -88,26 +101,43 @@ export class ConfigurationModuleApi
...
+      const websocketServerConfigId = extraQueries[SetNetworkProfileExtraQuerystrings.websocketServerConfigId];
+      await SetNetworkProfile.build({
+        stationId: identifier,
+        correlationId,
+        configurationSlot: request.configurationSlot,
+        websocketServerConfigId,
+        apn: JSON.stringify(request.connectionData.apn),
+        vpn: JSON.stringify(request.connectionData.vpn),
+        ...request.connectionData
+      }).save();
```
lydiazcheng, line 132: Just a question, in this way the create CrudEvent will not be emitted but it is also not necessary to emit SetNetworkProfile and related db entries, right? 🤔

01_Data/src/layers/sequelize/model/CallMessage.ts
```diff
@@ -22,4 +22,10 @@ export class CallMessage extends Model {
 
   @Column(DataType.INTEGER)
   declare reservationId?: number;
+
+  @Column(DataType.STRING)
+  declare websocketServerConfigId?: string
+
+  @Column(DataType.INTEGER)
+  declare configurationSlot?: number
```
lydiazcheng, line 30: I did not find any usages of these 2 cols. Maybe we can remove them?

03_Modules/Configuration/src/module/module.ts
```diff
@@ -529,11 +533,25 @@ export class ConfigurationModule extends AbstractModule {
...
   @AsHandler(CallAction.SetNetworkProfile)
-  protected _handleSetNetworkProfile(
+  protected async _handleSetNetworkProfile(
     message: IMessage<SetNetworkProfileResponse>,
     props?: HandlerProperties,
-  ): void {
+  ): Promise<void> {
     this._logger.debug('SetNetworkProfile response received:', message, props);
+
+    const setNetworkProfile = await SetNetworkProfile.findOne({ where: { correlationId: message.context.correlationId } });
```
lydiazcheng, line 542: I think we need to check the response status and update if it is `Accepted`

lydiazcheng (commented):
The changes look great! Just some small points. Also it would be great to fix the failed test.

02_Util/src/networkconnection/authenticator/NetworkProfileFilter.ts
```diff
@@ -0,0 +1,91 @@
...
+                if (configurationSlotsArray.length == 0) {
+                    this._logger.debug("No valid configuration slots to check: " + configurationSlotsString);
+                    return true;
+                } else {
+                    let securityProfileAllowed = false;
+                    for (const configurationSlot of configurationSlotsArray) {
+                        const chargingStationNetworkProfile = await ChargingStationNetworkProfile.findOne({ where: { stationId: identifier, configurationSlot: configurationSlot } });
+                        if (chargingStationNetworkProfile) {
+                            const serverNetworkProfile = await ServerNetworkProfile.findByPk(chargingStationNetworkProfile.websocketServerConfigId);
+                            if (serverNetworkProfile && securityProfile >= serverNetworkProfile.securityProfile) {
```
yuvenus, line 72: in the scenario where a security profile is found to be greater than a configuration slot (and possibly all existing configuration slots), is there a need to store that security profile as well as a configuration? or is that handled otherwise in the SetNetworkProfile handler in the configuration api?

02_Util/src/networkconnection/authenticator/NetworkProfileFilter.ts
```diff
@@ -0,0 +1,91 @@
...
+
+    constructor(
+        deviceModelRepository: IDeviceModelRepository,
+        logger?: Logger<ILogObj>,
+    ) {
+        super(logger);
+        this._deviceModelRepository = deviceModelRepository;
+    }
+
+    protected shouldFilter(options: AuthenticationOptions): boolean {
```
yuvenus, line 24: minor, linter complaint that `options` should be `_options`

03_Modules/Configuration/src/module/api.ts
thanaParis: Yes!

00_Base/src/interfaces/api/AbstractModuleApi.ts
```diff
@@ -111,42 +110,57 @@ export abstract class AbstractModuleApi<T extends IModule>
...
+      return method.call(
         this,
-        request.query.identifier,
-        request.query.tenantId,
+        identifier,
+        tenantId,
         request.body,
-        request.query.callbackUrl,
+        callbackUrl,
+        Object.keys(extraQueries).length > 0 ? extraQueries : undefined,
```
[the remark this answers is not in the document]
thanaParis: that function cannot return undefined--it can only return an empty array, which is why there's a check for the length

PR #292, opened 2024-11-11
Feature/set is online

thanaParis:


03_Modules/OcppRouter/src/module/router.ts
```diff
@@ -136,6 +149,7 @@ export class MessageRouterImpl
 
   async deregisterConnection(connectionIdentifier: string): Promise<boolean> {
     this._webhookDispatcher.deregister(connectionIdentifier);
+    this._locationRepository.setChargingStationIsOnline(connectionIdentifier, false);
```
lydiazcheng, line 152: Do we need `await` here?

lydiazcheng (approved):
The change looks good!

03_Modules/OcppRouter/src/module/router.ts
thanaParis: we don't right now

PR #293, opened 2024-11-12
Bugfix: Store Total Cost in DB

lydiazcheng:
This PR is to fix this issue: https://github.com/citrineos/citrineos/issues/50

**Tests:**
When sending total cost in `CostUpdatedRequest`, the cost is stored in db.Screenshot 2024-11-12 at 10 12 28 AM Screenshot 2024-11-11 at 5 49 52 PM
When sending total cost in `TransactionEventResponse`, the cost is stored in db.Screenshot 2024-11-12 at 10 13 22 AM Screenshot 2024-11-11 at 6 04 22 PM

### DIFF
--- a/03_Modules/Transactions/test/module/CostNotifier.test.ts
+++ b/03_Modules/Transactions/test/module/CostNotifier.test.ts
@@ -19,6 +19,7 @@ describe('CostNotifier', () => {
 
     transactionEventRepository = {
       readTransactionByStationIdAndTransactionId: jest.fn(),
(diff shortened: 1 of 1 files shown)


yuvenus (approved):
looks good to me!

just curious, will there be a need to also store just the "cost" within a transaction event (so that you can see a history of the cost throughout the session) or is it enough just to have the total cost on the transaction level 🤔 (likely another PR though if the answer is yes)

ChrisWeissmann (approved):

PR #294, opened 2024-11-14
Fix: generate directus flows

elliot-sabitov:
- fix: generate directus flows failing because full route object schemas were no longer available after registering schemas and using $refs
- fix: adjusting to register schemas globally so that they can be available when generating directus flows
- feat: added MessageConfirmationSchema and setting default message endpoint responses to be MesageConfirmationSchema
- fix: bug where if shema was already registered it would not properly return the $ref for the schema resulting in missing definition in OpenAPI spec
- fix: TlsCertificates schema to properly set certificateChain as string array
- fix: AsDataEndpoint to properly ignore `undefined` tags instead of setting `[undefined] as the tag list which resulted in incorrect null tags
- fix: registerSchema in AbstractModuleApi to remove empty required lists, recursively register nested schema definitions (which fastify does not do by default), fix property.$refs and property.items.$refs to not use '#/definitions/' since fastify simply expects the unique id
- fix: bringing back a smaller version of OcppTransformObject which now only sets the default tags (if not passed in) based on path key

### DIFF
--- a/00_Base/src/interfaces/api/AbstractModuleApi.ts
+++ b/00_Base/src/interfaces/api/AbstractModuleApi.ts
@@ -338,6 +338,36 @@ export abstract class AbstractModuleApi<T extends IModule>
     }
     try {
       const schemaCopy = this.removeUnknownKeys(schema);
+      if (
+        schemaCopy.required &&
+        Array.isArray(schemaCopy.required) &&
+        schemaCopy.required.length === 0
+      ) {
+        delete schemaCopy.required;
+      }
+      if (schema.definitions) {
+        Object.keys(schema.definitions).forEach((key) => {
+          const definition = schema.definitions[key];
+          if (!definition['$id']) {
+            definition['$id'] = key;
+          }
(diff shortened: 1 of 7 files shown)


ChrisWeissmann (commented):
I still get some ref errors e.g.
```Resolver error at paths./ocpp/certificates/certificateSigned.post.requestBody.content.application/json.schema.properties.customData.$ref
Could not resolve reference: Could not resolve pointer: /components/schemas/CustomDataType does not exist in document
Resolver error at paths./ocpp/certificates/certificateSigned.post.requestBody.content.application/json.schema.properties.certificateType.$ref
Could not resolve reference: Could not resolve pointer: /components/schemas/CertificateSigningUseEnumType does not exist in document```

Also could we maybe straighten out the tags? It seems the `@AsDataEndpoint` have no tags (as we haven't set them when annotating the endpoint) while the other endpoints have `Default` as their tag.

thanaParis (approved):
Approved!

PR #297, opened 2024-11-20
Feature/unknown 404

thanaParis:
From the OCPP 2.0.1 specification:

> • If the CSMS does not recognize the Charging Station identifier in the URL path, it SHOULD send an HTTP response with status 404 and abort the WebSocket connection as described in [RFC6455].

Citrine previously always gave a 401, this pr gives a 404 when the charger is unknown and 500 when the server itself has an error.

### DIFF
--- a/02_Util/src/networkconnection/WebsocketNetworkConnection.ts
+++ b/02_Util/src/networkconnection/WebsocketNetworkConnection.ts
@@ -18,6 +18,7 @@ import fs from 'fs';
 import { ErrorEvent, MessageEvent, WebSocket, WebSocketServer } from 'ws';
 import { ILogObj, Logger } from 'tslog';
 import { SecureContextOptions } from 'tls';
+import { IUpgradeError } from './authenticator/errors/IUpgradeError';
 
 export class WebsocketNetworkConnection {
   protected _cache: ICache;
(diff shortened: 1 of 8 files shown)


02_Util/src/networkconnection/WebsocketNetworkConnection.ts
```diff
@@ -268,21 +269,21 @@ export class WebsocketNetworkConnection {
...
         wss.emit('connection', ws, req);
       });
-    } catch (error) {
+    } catch (error: any) {
       this._logger.warn(error);
-      this._rejectUpgradeUnauthorized(socket);
+      /**
+       * See {@link IUpgradeError.terminateConnection}
+       **/
+      error?.terminateConnection?.(socket) || this._terminateConnectionInternalError(socket);
```
elliot-sabitov, line 277: isnt the left and the right always going to fire? I am seeing that `error?.terminateConnection` returns void, doesnt that mean that the right side will fire unintentionally?

elliot-sabitov: would it be cleaner to move:

```
terminateConnection(socket: Duplex): void {
        socket.write('HTTP/1.1 401 Unauthorized\r\n');
        socket.write(
            'WWW-Authenticate: Basic realm="Access to the WebSocket", charset="UTF-8"\r\n',
        );
        socket.write('\r\n');
        socket.end();
        socket.destroy();
    }
```
as a helper method along with `this._terminateConnectionInternalError` and instead maybe do something like:
```
if (error instanceOf UpgradeAuthenticationError) {
  this.terminateAuthenticationError();
} else {
  this.terminateConnectionError();
}
```
the two functions are more or less the same and it feels a bit odd that the socket handling is in the error

elliot-sabitov: ah I see, there is also UpgradeUnknownError, but maybe there can be a handle helper function here like: 
```
terminateConnectionHelper(socket: Duplex, str1 = 'DEFAULT', str2 = 'DEFAULT'): void {
        socket.write('${str1}\r\n');
        socket.write(
            '${str2}\r\n',
        );
        socket.write('\r\n');
        socket.end();
        socket.destroy();
    }
```
and then just:
```
if (error instanceOf UpgradeAuthenticationError) {
  this.terminateConnectionHelper(socket, 'HTTP/1.1 401 Unauthorized', ''WWW-Authenticate: Basic realm="Access to the WebSocket", charset="UTF-8"');
} else if (error instanceOf UpgradeUnknownError) {
  this.terminateConnectionHelper(socket, 'HTTP/1.1 404 Not Found');
} else {
 this.terminateConnectionHelper(socket);
}
```

ChrisWeissmann: I am not sure what we are gaining. Aren't we pulling complexity up instead of pushing it down here? The Error's thrown here are very specific. So including the operation in them makes sense to me.

elliot-sabitov: Cool, sounds good, but please double check on this line:
```
error?.terminateConnection?.(socket) || this._terminateConnectionInternalError(socket);
```
I do believe that it may misfire, see attached screenshot
Screenshot 2024-11-27 at 9 37 28 AM

thanaParis: Updated so that the method returns a boolean, preventing the right hand side from being executed if the left is successful.

lydiazcheng (approved):

PR #298, opened 2024-11-20
fix: name_instance unique constraint in Components and Variables model not working properly

elliot-sabitov:
- fix: adding await before request to repository to resolve unhandled promise rejection which terminates the server
- feat: adding unique index on `name` field when `instance` field IS NULL because NULL by default is not treated as DISTINCT by Postgres resulting in name_instance unique index not being enforced where instance is null


lydiazcheng (approved):

ChrisWeissmann (approved):

thanaParis (approved):
approved

PR #301, opened 2024-11-25
Feature/Handle Meter Value Cost Update

lydiazcheng:
## Context
If `sendCostUpdatedOnMeterValue` is true, meterValues handler should trigger cost update

## Changes
1. Add checks for `sendCostUpdatedOnMeterValue` config and `meterValue` context in the meterValues handler. If they are `true` and `ReadingContextEnumType.Sample_Periodic`, handler stores the `transactionDbId` in `meterValue` record.  And then it triggers a `CostUpdated` request.
2. Update the `totalCost` in database.
3. Fix missing last `meterValue` problem when creating a new `meterValue` in db and then reading all `meterValues`.

## Test
1. In config, remove `costUpdatedInterval` and set `sendCostUpdatedOnMeterValue` to true.
2. Create a tariff for charger.
3. Charger sends `TransactionEvent` with `"eventType":"Started"` to start a new transaction
4. Charger sends `MeterValues`
5. Citrine responds the request and sends a `CostUpdated` req containing the `totalCost`Screenshot 2024-11-25 at 11 49 54 AM
6. The totalCost is updated in db. Screenshot 2024-11-25 at 11 53 53 AM


thanaParis (approved):
Approved

PR #303, opened 2024-11-27
Feature/url handling streaming data

eddygrebt:
Modifying IFileAccess, adding new method (GetFileURL). Adjusting all classes that uses IFileAccess. Creating an Endpoint to gather the File URL.


thanaParis (approved):
Approved

PR #304, opened 2024-11-27
Feature/ocpp 1.6 types

thanaParis:
Adding support for OCPP 1.6.

- json schema generator created for OCPP 1.6 Json v4 schemas
- OCPP types separated into namespaces, package structure changed to accommodate this
- AsHandler decorator updated to include protocol
- websocket handling fixed to reject incorrect protocols
- Support for Google Pubsub is being suspended as part of this PR. It may return as part of a future PR which accommodates the changes made to how CallActions are organized.


00_Base/src/util/request.ts
```diff
@@ -24,6 +24,7 @@ export class RequestBuilder {
     eventGroup: EventGroup,
     origin: MessageOrigin,
     timestamp: Date = new Date(),
+    protocol?: string,
```
lydiazcheng, line 27: How about using `OCPPVersion` instead of `string`?

lydiazcheng (approved):
Great change. Just few minor points.

00_Base/src/config/types.ts
```diff
@@ -155,7 +148,7 @@ export const systemConfigInputSchema = z.object({
           })
           .optional(),
       })
-      .refine((obj) => obj.pubsub || obj.kafka || obj.amqp, {
+      .refine((obj) => obj.kafka || obj.amqp, {
```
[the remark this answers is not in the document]
thanaParis: Support for pubsub is being removed. It clashed with the changes in this pr in a way that would take some effort to fix, so I have removed it for now. If we hear back that it's wanted, we can fix those issues in a separate PR and return support.

00_Base/src/util/request.ts
thanaParis: Adjusted!

PR #305, opened 2024-12-02
Fix: logs in the cloud broken into multiple lines

elliot-sabitov:
- fix: using json log type when initializing logger and process.env.DEPLOYMENT_TARGET is 'cloud' to help make sure that logs in the cloud are not split into multiple lines
- feat: setting transportJSON override to remove _meta when systemConfig.logLevel > 2 to ensure that it does not include unnecessary information


thanaParis (approved):
approved

PR #307, opened 2024-12-03
Feature: Moving Fastify dependencies to Base and enforcing NodeJS version

elliot-sabitov:
- feat: moved fastify dependencies to base so that they can be used transitively through base's dependencies
- feat: moved NodeJS version requirement from /Server to root package.json, increased the required version to 18.18.0 as required by eslint dependency and created .npmrc to enable engine-strict check otherwise an error is not thrown when installing with a lower version of node


thanaParis (approved):
Approved

PR #309, opened 2024-12-03
Bugfix/Issue61 Fix BootNotification Crash

lydiazcheng:
# Context
This PR is to fix the issue:  https://github.com/citrineos/citrineos/issues/61 
With this fix citrine does not crash when handling the invalid BootNotification message from charger.
# Test
Send the invalid BootNotification and receive response. Connection doesn't crash. Screenshot 2024-12-03 at 6 11 24 PM

### DIFF
--- a/03_Modules/Configuration/src/module/module.ts
+++ b/03_Modules/Configuration/src/module/module.ts
@@ -250,7 +250,7 @@ export class ConfigurationModule extends AbstractModule {
       await this.sendCallResultWithMessage(message, bootNotificationResponse);
 
     // Update device model from boot
-    this._deviceModelService.updateDeviceModel(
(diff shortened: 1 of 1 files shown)


03_Modules/Configuration/src/module/module.ts
```diff
@@ -250,7 +250,7 @@ export class ConfigurationModule extends AbstractModule {
       await this.sendCallResultWithMessage(message, bootNotificationResponse);
 
     // Update device model from boot
-    this._deviceModelService.updateDeviceModel(
+    await this._deviceModelService.updateDeviceModel(
```
lydiazcheng, line 253: Here is the only change. Others are just reformat.

elliot-sabitov (approved):

PR #311, opened 2024-12-05
Feature/59 remove deasync

elliot-sabitov:
- feat: removing deasync dependency by instead awaiting promises outside of constructors
- refactor: adjusted logic such that the requests/responses are defined in the AbstractModule and creating an initHandlers helper method that can be triggered on the module to await the existing initHandler promise that was previously wrapped in deasyncPromise
- refactor: adjusted logging behavior such that the AbstractModule can log the initialization and initialization duration messages
- refactor: adjusted the CitrineOSServer to trigger initSystem() during the async initialize() that is triggered post constructor initialization via run(), made initSystem() async and made, created helper methods to initialize the individual modules, and fixed initModule helper to invoke one of the individual module initialization helpers. Individual module initialization methods are now also async and will await initHandlersAndAddModule helper method which will trigger and await the new initHandlers method in the AbstractModule, thereby allowing the promises to be awaited outside of the constructors
- chore: removing Timer class entirely and simplifying how we computed the initialization duration captured solely within AbstractModule class via single startTime variable
- chore: removing getSync and setSync along with todo as these methods are no longer used and are only other place where deasyncPromise was used
- chore: removing deasync from dockerfiles


03_Modules/Configuration/src/module/module.ts
```diff
@@ -61,29 +61,23 @@ import {
...
-import { SequelizeChargingStationSequenceRepository } from '@citrineos/data';
 
 /**
  * Component that handles Configuration related messages.
  */
 export class ConfigurationModule extends AbstractModule {
   public _deviceModelService: DeviceModelService;
 
-  protected _requests: CallAction[] = [
+  _requests: CallAction[] = [
```
lydiazcheng, line 80: change for `_responses` is missing?

03_Modules/SmartCharging/src/module/module.ts
```diff
@@ -68,7 +62,7 @@ export class SmartChargingModule extends AbstractModule {
    * Fields
    */
 
-  protected _requests: CallAction[] = [
+  _requests: CallAction[] = [
```
lydiazcheng, line 65: Do we need change for` _responses` as well?

lydiazcheng (approved):
Hi Elliot, the change looks good! Only few tiny points from my side.

03_Modules/Configuration/src/module/module.ts
elliot-sabitov: 🙊 not sure how I missed this, actually probably would work anyway, but it's good to stay consistent and not have unnecessary code / access modifier on the derived class as it will be inherited from the abstract base class - thank you catching this! 🙇

03_Modules/SmartCharging/src/module/module.ts
elliot-sabitov: removed thank you!

03_Modules/Configuration/src/module/module.ts
lydiazcheng: > 🙊 not sure how I missed this, actually probably would work anyway, but it's good to stay consistent and not have unnecessary code / access modifier on the derived class as it will be inherited from the abstract base class - thank you catching this! 🙇

I see. 👍 Thank you for updating, Elliot!

PR #312, opened 2024-12-06
Feature/ocpp broadcast api

eddygrebt:
Refactoring all endpoints that focus on sending calls to charging points. Making the endpoints able to call more than a charging point per function.


eddygrebt:
This might need a deep look into every endpoint, I've tested a bunch of them, but there's a lot more to test.

thanaParis (approved):
approved

PR #313, opened 2024-12-06
fix string interpolation in exception message

exterm:
While trying to set up CitrineOS locally with EVerest, I got a very unhelpful error message. It looks like the string interpolation in this error message is broken, so I'd like to submit this fix.


thanaParis (approved):
Approved! Thank you for this fix.

PR #314, opened 2024-12-06
Feature/everest updates

elliot-sabitov:
- feat: updating everest-demo to newest version 0.0.23 including adjustments to account for shift from alphine to debian and adjusted entrypoint to be start.sh to trigger sil script properly


thanaParis (approved):
Approved

PR #315, opened 2024-12-06
Feature - Replacing Bcrypt with NodeJS built in PBKDF2

elliot-sabitov:
- feat: replacing bcrypt with nodes built in pbkdf2Sync to persist and  check for matching hashes of passwordString VariableAttribute dataType during BasicAuthentication handling


yuvenus:
FYI working so far on windows 11 as well 👀🙏

ChrisWeissmann:
We are holding this back, as it's a breaking change and it will go in when we have a significant version bump that allows breaking changes.

thanaParis (approved):
approved

PR #299, opened 2024-11-20
chore: disallow floating promises

elliot-sabitov:
- chore: adjusted eslint to disallow floating promises which can lead to unhandled promise rejections and result in server crashes

As seen in this PR https://github.com/citrineos/citrineos-core/pull/298 - unhandled promise exceptions can lead to server crashes, so it is a good idea to use the linter to catch and resolve any such cases!

Attached below is a list of lint issues that arise with the introduction of this PR showing around ~80 cases where Promises are not awaited.
[lint-issues.txt](https://github.com/user-attachments/files/17836021/lint-issues.txt)

Created https://github.com/citrineos/citrineos-core/pull/316 PR to resolve the issues

PR #316, opened 2024-12-10
fix: disallowed floating promises

elliot-sabitov:
This PR addresses the list of lint issues that is created as a result of disallowing floating promises in the lint config which was introduced here - https://github.com/citrineos/citrineos-core/pull/299


thanaParis (approved):
Approved

PR #317, opened 2024-12-11
feat: adding conditional unique indexes on nullable fields in Evse and VariableAttributes

elliot-sabitov:
- feat: adding conditional unique indexes on nullable fields in Evse and VariableAttributes models because otherwise postgres does not treat NULLs as distinct and unique index is not enforced


thanaParis (approved):
Approved

PR #318, opened 2024-12-11
Fix - docker compose error because of failing hasura engine 

elliot-sabitov:
- fix: healthchecks for hasura services
- chore: removing unnecessary depends_on
- chore: increasing retries


yuvenus (approved):
tried on my machine, no docker compose errors so far!

ChrisWeissmann (approved):

PR #319, opened 2024-12-11
fix: hasConcurrentTransactions check

elliot-sabitov:
- fix: hasConcurrentTransactions check to properly only get matching transactions for given idToken by adjusting readAllActiveTransactionsByIdToken to use `required: true` when joining TransactionEvent and IdToken models so that INNER JOIN is used, otherwise OUTER JOIN is used resulting in any active transaction being included with null TransactionEvent and IdToken values
- chore: adjusting to async syntax


thanaParis (approved):
Approved

PR #322, opened 2025-01-07
Issue 70: Generalize Boot (No.1)

lydiazcheng:
## Context
This PR is to generalize the Boot db entity to make it work with multiple protocols.
## Changes
1. Remove OCPP 2.0.1 references from `Boot`. Add null type for optional fields.
2. Create a intermediate model `BootMapper` to map between ocpp 2.0.1 types and general types the db entity.
3. Use the new mapper and provides the correct types in `determineBootStatus` method.
## Test
1. create a boot in db.Screenshot 2025-01-09 at 1 23 11 PM
2. send a boot notification req Screenshot 2025-01-09 at 1 56 20 PM
3. receive the response with the status, which means the usage of the BootMapper in `determineBootStatus` works. Screenshot 2025-01-09 at 1 56 52 PM


01_Data/src/layers/sequelize/model/Boot.ts
```diff
@@ -49,10 +49,10 @@ export class Boot extends Model implements BootConfig {
...
   @Column(DataType.JSON)
-  declare variablesRejectedOnLastBoot: OCPP2_0_1.SetVariableResultType[];
+  declare variablesRejectedOnLastBoot?: object[] | null;
 
   @Column(DataType.BOOLEAN)
-  declare bootWithRejectedVariables?: boolean;
+  declare bootWithRejectedVariables?: boolean | null;
 
-  declare customData?: OCPP2_0_1.CustomDataType;
+  declare customData?: object;
```
langiole, line 57: need `| null here`?

01_Data/src/layers/sequelize/model/Boot.ts
```diff
@@ -25,22 +25,22 @@ export class Boot extends Model implements BootConfig {
...
   @Column(DataType.INTEGER)
-  declare heartbeatInterval?: number;
+  declare heartbeatInterval?: number | null;
 
   @Column(DataType.INTEGER)
-  declare bootRetryInterval?: number;
+  declare bootRetryInterval?: number | null;
 
   @Column(DataType.STRING)
-  declare status: OCPP2_0_1.RegistrationStatusEnumType;
```
langiole, line 37: Should we defining the `OCPP1_6.RegistrationStatusEnumType` somewhere?

lydiazcheng: > Should we defining the `OCPP1_6.RegistrationStatusEnumType` somewhere?

Yes. We decided to introduce intermediate objects, e.g., BootMapper. The protocol specific types are defined in these mappers. So the `OCPP1_6.RegistrationStatusEnumType` is defined [here](https://github.com/citrineos/citrineos-core/pull/325/files#diff-c4dcafdb6122f702f987d46523802644cfc5b246ab448ff7cb717ddba3f05545).

lydiazcheng:
See new PR(https://github.com/citrineos/citrineos-core/pull/348) for this.

PR #325, opened 2025-01-15
Issue 92: Implement OCPP 1.6 BootNotification (No.2)

lydiazcheng:
## Info
1. ⚠️  Please take care of the merge order or update the destination branch before merging.
2. Most changed files are due to renaming.

## Context
This change is to implement the BootNotification in OCPP 1.6. When booting, it handles it as follow,
1. For new boot:
    - When default status in config is accepting
       - Send accept response
       - Store boot and changing station in db
   - When default status in config is pending
      - Send pending response
      - Blacklist actions
      - Change configuration and then get all configurations
      - Trigger another boot
   - When default status in config is rejected
      - Send rejected response
      - Blacklist actions
      - Store boot and changing station in db
2. Existing boot
    - When its status is accepting
       - Send accept response
       - Update charging station and boot
    - When its status is pending
       - Send accept response if no need for configuration related actions
       - Update charging station and boot
       - If need, change and get configs again and trigger another boot

## Changes
1. Change OCPP namespace into 3 enums. Update data endpoint to work with any of the namespaces.
2. Extend system config
    - Extend configuration module to have `ocpp1_6` and `ocpp2_0_1` definitions
    - Add network connection config for ocpp1.6 local and docker config file
3. Add OCPP 1.6 `BootMapper` object and `ChangeConfiguration` table. Extend the `ChargingStation` table.
4. Add OCPP 1.6 Handlers for
    - BootNotificationRequest
    - ChangeConfigurationResponse
    - GetConfigurationResponse

## Tests
**1. New boot when default config is pending.**
Prepare: Set `unknownChargerStatus` to Pending. Store 2 configurations in db.
Below is the message flow, Screenshot 2025-01-15 at 11 20 05 AM
All db entities are created/updated.
Screenshot 2025-01-15 at 11 22 07 AM
Screenshot 2025-01-14 at 7 39 43 PM
Screenshot 2025-01-15 at 11 25 10 AM
**2. New boot when default config is Accepted.**
Below is the message flow, Screenshot 2025-01-15 at 11 27 07 AM
**3. New boot when default config is Rejected.**
Below is the message flow, 
Screenshot 2025-01-15 at 11 28 38 AM
**4. Data endpoint still works after update the namespace.**
Screenshot 2025-01-15 at 10 09 37 AM


01_Data/src/layers/sequelize/model/Authorization/Authorization.ts
```diff
@@ -3,14 +3,14 @@
...
-import { Namespace, OCPP2_0_1 } from '@citrineos/base';
+import { OCPP2_0_1_Namespace, OCPP2_0_1 } from '@citrineos/base';
 import { BelongsTo, Column, DataType, ForeignKey, Model, Table } from 'sequelize-typescript';
 import { type AuthorizationRestrictions } from '../../../../interfaces';
 import { IdToken, IdTokenInfo } from '.';
 
 @Table
 export class Authorization extends Model implements OCPP2_0_1.AuthorizationData, AuthorizationRestrictions {
-  static readonly MODEL_NAME: string = Namespace.AuthorizationData;
+  static readonly MODEL_NAME: string = OCPP2_0_1_Namespace.AuthorizationData;
```
langiole, line 13: Little bit confused by this change because in the other generalize PRs we were reverting this back to `Namespace.etc`.

lydiazcheng: > Little bit confused by this change because in the other generalize PRs we were reverting this back to `Namespace.etc`. Shouldn't these models remain version agnostic?

Not all data entities are generalized. Some entities are still related to the specific version, e.g., variableAttribute. So I split the namespace into 3. There might be other better ideas. Pls feel free to share.
- `Namespace` for generalized entities
- `OCPP2_0_1_Namespace` for 2.0.1 related entities
- `OCPP1_6_Namespace` for 1.6 related entities.

In this PR, I only renamed the old `Namespace` like above but did not regroup all db entities. That's should be the reason `Authorization` is moved to `Namespace` again in another PR since it is a generalized one.

lydiazcheng:
New PR for this is here: https://github.com/citrineos/citrineos-core/pull/347

PR #334, opened 2025-01-27
fix: Check transaction status and set 'active' to false if status is SuspendedEVSE or SuspendedEV

binc0:
# Fix Description

This update resolves an issue where transactions stuck in **`SuspendedEVSE`** or **`SuspendedEV`** states remain marked as `active`, causing persistent `concurrenttx` errors and blocking new transactions.

---

## Problem  
- Transactions failing to start (e.g., due to concurrency limits, hardware conflicts, or initialization errors) enter `SuspendedEVSE` or `SuspendedEV` states.  
- The `active` flag is not automatically set to `false` in these states, causing the system to treat them as "still running."  
- This triggers infinite loops of `concurrenttx` errors, preventing new transactions from initiating.

---

## Solution  
A conditional check is added to the transaction status handler:  
1. **Status Detection**: Actively monitor transactions for `SuspendedEVSE` or `SuspendedEV` states.  
2. **Flag Update**: Automatically set `active` to `false` when either suspended state is detected.  
3. **Concurrency Handling**: Free up the system to process new transactions by marking stuck ones as inactive.  

**Key Benefits**:  
- 🛑 Prevents infinite `concurrenttx` error loops.  
- 🔄 Enables seamless retries after failures.  
- 🛠 Graceful handling of edge cases where transactions fail to initialize.  

---

## Impact  
- ✅ Transactions in suspended states no longer block subsequent operations.  
- ✅ System stability improves by avoiding zombie transactions.  
- ✅ Users/chargers can retry transactions without manual intervention.

### DIFF
--- a/03_Modules/Transactions/src/module/TransactionService.ts
+++ b/03_Modules/Transactions/src/module/TransactionService.ts
@@ -137,6 +137,19 @@ export class TransactionService {
         }
       }
     }
+    
+    // Check transaction status and set 'active' to false if status is SuspendedEVSE or SuspendedEV
+    if (transactionEvent.transactionInfo?.chargingState === OCPP2_0_1.ChargingStateEnumType.SuspendedEVSE  ||
+      transactionEvent.transactionInfo?.chargingState === OCPP2_0_1.ChargingStateEnumType.SuspendedEV) {
+      this._logger.debug(
+        `${transactionEvent.transactionInfo?.chargingState} for ${transactionEvent.transactionInfo.transactionId} transaction`,
+      );
+      await Transaction.update(
+        { isActive: false },
+        { where: { transactionId: transactionEvent.transactionInfo.transactionId } }
+      );
+    }
+
     this._logger.debug(
       'idToken Authorization final status:',
       response.idTokenInfo.status,


thanaParis:
Hi @binc0 ! Thank you for opening this issue

This is a tricky one... looking at OCPP 2.0.1 Part 2 - specification, there's descriptions for the SuspendedEV and SuspendedEVSE states:


> SuspendedEV:
> When the EV is connected to the EVSE and the EVSE is offering energy but the EV is not taking any energy.
> 
> SuspendedEVSE:
> When the EV is connected to the EVSE but the EVSE is not offering energy to the EV, e.g. due to a smart charging restriction, local supply power constraints, or when charging has stopped because of the authorization status in the response to a transactionEventRequest indicating that charging is not allowed etc.

While SuspendedEV leaves some room for interpretation, SuspendedEVSE is fairly explicit about not being a terminal state. It is possible for a transaction to enter and then leave SuspendedEVSE during a transaction due to smart charging. Therefore, marking a transaction as inactive due to having that state is going to cause problems.

I have a point of confusion about your scenario. Why did the charger not send the TransactionEvent with eventType 'Ended'? And if it tried to, but couldn't, why was that message not queued? The scenario you bring up, where a charging station ends a session with 'SuspendedEV' or 'SuspendedEVSE', but CitrineOS does not mark the transaction as inactive, could only happen if the charging station is not communicating correctly. If the eventType was set appropriately, and the message eventually received, this won't happen.

There is another problem hinted at here. What happens if one charging station loses connection during a transaction, then the user from that transaction attempts to charge at a different station before the first one has come back online and delivered queued messages? The 'concurrentTx' response would prevent them from charging, causing a poor user experience. This may be the scenario you are trying to describe. We should change when concurrentTx triggers. It shouldn't happen when the new transaction is being started on the same charger, for one. It also shouldn't happen when the other 'active' transaction is stale (no updates for longer than the SampledDataCtrlr.TxUpdatedInterval). See if you can't fix your issue with those changes instead!

thanaParis:
> My initial confusion stemmed from the fact that, in the simulator I am using ([chargestation.one](https://chargestation.one/)), only a SuspendedEVSE event is sent during a concurrent transaction.

> 
> Regarding the concurrentTx event, I am unsure whether it should automatically trigger a termination event. However, based on my tests, the simulator does not appear to send a termination event in this case, which could be contributing to the issue.

If I'm understanding you correctly, you're seeing the station send 2 events, one 'Started', one 'Updated', where the 'Started' event receives a concurrentTx rejection for the authorization and the 'Updated' event has 'SuspendedEVSE' as the cause, and then no more Transaction events are sent. Could you provide some logs? If this is what you're trying to fix, it may not be possible as that would be a serious breach of the protocol that can't be discerned from normal communication by CitrineOS.

Arbitrarily ending the current transaction whenever a concurrentTx is detected is not a good strategy. A station that is following the protocol will send additional TransactionEvents after a 'Started' or 'Updated' receives a rejection. There are only 2 cases where a transaction should be declared inactive: an 'Ended' event has been received, or we have reason to believe an 'Ended' event was supposed to have been sent but hasn't been received yet/will never be received.

binc0:
Many thanks for your advice! I initially misunderstood and thought that the termination of a transaction (Ended) would need to be managed by the Central System. After further investigation, I fixed the issue in the simulator and will now close this PR.

PR #341, opened 2025-01-31
add note to README about overriding config with env vars

evenlis:
Adds a note about using the hierarchical overriding of config values with underscore-separated environment variables (see [defineConfit.ts#mergeConfigFromEnvVars](https://github.com/citrineos/citrineos-core/blob/main/00_Base/src/config/defineConfig.ts#L70-L120))


langiole:
Confirmed working after running `export CITRINEOS_centralSystem_port=8081`:
Screenshot 2025-02-17 at 22 34 35

PR #348, opened 2025-02-07
Issue 70: Generalize Boot

langiole:
Generalized Boot related entities and added mapping for 2.0.1.


lydiazcheng (approved):
The PR looks good!

03_Modules/Configuration/src/module/BootNotificationService.ts
```diff
@@ -66,7 +66,7 @@ export class BootNotificationService {
       }
     }
 
-    return bootStatus;
+    return OCPP2_0_1_Mapper.BootMapper.toRegistrationStatusEnumType(bootStatus);
```
ChrisWeissmann, line 69: Very nitpick, and me just thinking out loud. Would making sure on line 39 that the bootStatus is a `RegistrationStatusEnumType` and from there on out operate on that assumption. Then we have a nicer return type. This just came about because I was looking at `bootStatus = OCPP2_0_1.RegistrationStatusEnumType.Accepted;` why are we putting it through the mapper again...

Server/src/config/envs/docker.ts
```diff
@@ -152,15 +152,6 @@ export function createDockerConfig() {
...
             ),
           },
-          {
-            id: '4',
-            securityProfile: 0,
-            allowUnknownChargingStations: true,
-            pingInterval: 60,
-            host: '0.0.0.0',
-            port: 8092,
-            protocol: 'ocpp1.6',
```
ChrisWeissmann, line 162: Why does this go away?

ChrisWeissmann: Gets added in the next PR correctly

ChrisWeissmann (approved):

PR #355, opened 2025-02-11
Issue 76: Generalize ChargingProfile

langiole:
Generalized ChargingProfile related entities.

### DIFF
--- a/00_Base/src/ocpp/persistence/namespace.ts
+++ b/00_Base/src/ocpp/persistence/namespace.ts
@@ -18,15 +18,15 @@ export enum Namespace {
   IdTokenInfoType = 'IdTokenInfo',
   ChargingProfile = 'ChargingProfile',
   ChargingSchedule = 'ChargingSchedule',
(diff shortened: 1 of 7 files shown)


00_Base/src/ocpp/persistence/namespace.ts
```diff
@@ -9,6 +9,11 @@
 
 export enum Namespace {
   BootConfig = 'Boot',
+  ChargingProfile = 'ChargingProfile',
+  ChargingSchedule = 'ChargingSchedule',
+  ChargingNeeds = 'ChargingNeeds',
+  CompositeSchedule = 'CompositeSchedule',
+  SalesTariff = 'SalesTariff',
```
lydiazcheng, line 16: `ChargingNeeds` and `SalesTariff` are both OCPP 2.0.1 specific object. I think we don't need to generalize them.

lydiazcheng (commented):
The PR looks good. Would be great to fix lint issues.

ChrisWeissmann (approved):

PR #346, opened 2025-02-06
Feature/init-generalize

langiole:
Preparing rc-1.6.0 branch for OCPP 1.6 related changes. All 1.6 related tickets will need these changes so will be branching off of this PR. 

Changes primarily include: 
1. The introduction of `OCPP2_0_1_Namespace` as we slowly migrate from 201 specific namespaces to generic namespaces.
2. Configuration needed to start up a OCPP 1.6 WebSocket server running on port `8092`.

PR #357, opened 2025-02-11
Issue 80: Restructure API and Implement TriggerMessage

lydiazcheng:
⚠️ This branch is based on feature/init-generalize: https://github.com/citrineos/citrineos-core/pull/346

## Changes
1. Split `api.ts` into 3 new files, 
    - `DataApi.ts`: for data endpoints
    - `2.0.1/MessageApi.ts`: for ocpp 2.0.1 message endpoints
    - `1.6/MessageApi.ts`: for ocpp 1.6 message endpoints
⚠️ ⚠️ ⚠️ Other open PRs might change sth in any old `api.ts` files. Please make sure the new `DataApi.ts`, `Ocpp201Api.ts` contains those changes before merging. If not, pls update `DataApi.ts`, `Ocpp201Api.ts` to contains them.
2. Add ocpp version as prefix to schema id in `AbstractModuleApi` and `AbstractMessageRouter` so that they can handle schemas when 2 versions have the same schema id.
3. Implement OCPP 1.6 TriggerMessage message endpoint and handler.

#### Case 1: Test new 1.6 trigger message endpoint
1. Send the request
Screenshot 2025-02-11 at 12 25 05 PM

2. Charger receives the request and then send back a response
Screenshot 2025-02-11 at 12 25 41 PM

4. Citrine receives the response
Screenshot 2025-02-11 at 12 22 46 PM

#### Case 2: Regression test for existing 2.0.1 trigger message endpoint
1. Don't stop the running citrine but disconnect the 1.6 connection and create a new 2.0.1 connection.
2. Send the request 
Screenshot 2025-02-11 at 12 27 45 PM

3. Charger receives the request and then send back a response 
Screenshot 2025-02-11 at 12 32 29 PM

5. Citrine receives the response without errors. (the response schema of 2.0.1 and 1.6 has the same schema id, we need to confirm citrine works in this case. )  
Screenshot 2025-02-11 at 12 23 55 PM

#### Case 3: Regression test for sending a 2.0.1 call from charger to citrine  
Screenshot 2025-02-11 at 12 33 23 PM


03_Modules/Certificates/src/module/Ocpp201Api.ts
langiole, line 1: Alternative file structure idea?
```
.
└── src/
    └── module/
        └── api.ts/
            ├── 2.0.1/
            │   └── api.ts
            └── 1.6/
                └── api.ts
```

langiole: I would say that's better than what we have now.

My preference would be to split the 1.6 and 2.0.1 handlers into seperate files then we can do something like this:

```
.
└── src/
    └── module/
        └── api.ts/
        └── module.ts/
            ├── 2.0.1/
            │   └── api.ts
            │   └── handlers.ts
            └── 1.6/
                └── api.ts
                └── handlers.ts
```

lydiazcheng: Got it. 👍  I will restructure the api files. For handlers, we can have a further discussion in the team.

lydiazcheng: Updated the file structure. 🙌

ChrisWeissmann (approved):
Lets also add this new structure to the docs as a core concept. @langiole

PR #365, opened 2025-02-13
Issue 103: Publish Swagger

langiole:
Added ability to build and publish swagger doc to [citrine](https://github.com/citrineos/citrineos.github.io).

### DIFF
--- a/.github/workflows/publish-swagger.yml
+++ b/.github/workflows/publish-swagger.yml
@@ -5,6 +5,7 @@ on:
 
 permissions:
   contents: write
+  pull-requests: write
 
 jobs:
   build:
--- a/.github/workflows/publish-swagger.yml
+++ b/.github/workflows/publish-swagger.yml
@@ -57,6 +57,8 @@ jobs:
         env:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
         run: |
+          echo "Pull Requests: ${{ github.token_permissions.pull-requests }}"
+
           PR_EXISTS=$(gh pr list --repo citrineos/citrineos.github.io --head swagger-updates --state open --json number --jq '.[0].number' || echo "")
 
           if [ -z "$PR_EXISTS" ]; then
--- a/.github/workflows/publish-swagger.yml
+++ b/.github/workflows/publish-swagger.yml
@@ -1,7 +1,9 @@
 name: Publish Swagger
 
 on:
-  pull_request:
+  push:
+    branches:
+      - main
 
 jobs:
   build:
@@ -30,7 +32,11 @@ jobs:
       - name: Validate Swagger JSON
         id: validate-swagger
         run: |
-          curl -X POST -d @swagger.json -H 'Content-Type:application/json' https://validator.swagger.io/validator/debug
+          if jq empty swagger.json; then
+            echo "swagger_valid=true" >> $GITHUB_OUTPUT
+          else
+            echo "swagger_valid=false" >> $GITHUB_OUTPUT && exit 1
+          fi
 
       - name: Upload Swagger JSON as Artifact
         uses: actions/upload-artifact@v4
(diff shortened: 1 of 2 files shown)


.github/workflows/publish-swagger.yml
```diff
@@ -0,0 +1,77 @@
+name: Publish Swagger
+
+on:
+  push:
+    branches:
+      - main
```
ChrisWeissmann, line 6: Should we make this dependent on release tags instead?

.github/workflows/publish-swagger.yml
```diff
@@ -0,0 +1,77 @@
...
+      - name: Push Swagger JSON
+        run: |
+          cd citrineos.github.io
+          git checkout -b swagger-updates || git checkout swagger-updates
+          cp ../swagger.json ./docs/assets
+          git config user.name "GitHub Actions"
+          git config user.email "actions@github.com"
+          git add ./docs/assets/swagger.json
+          git commit -m "Update Swagger JSON from PR"
+          git push -f origin swagger-updates --verbose
```
ChrisWeissmann, line 77: I think you could automate a PR here with the github CLI (`gh`) 
```shell
gh pr create \
          --repo citrineos/citrineos.github.io \
          --base main \
          --head "$branch_name" \
          --title "Update Swagger JSON - $(date +%Y-%m-%d)" \
          --body "This PR updates the Swagger JSON automatically.
```

Also it might be nice to pass in a date to the branch name.

.github/workflows/publish-swagger.yml
```diff
@@ -0,0 +1,77 @@
...
+
+      - name: Set up Docker Buildx
+        uses: docker/setup-buildx-action@v3
+
+      - name: Set up QEMU
+        uses: docker/setup-qemu-action@v3
+
+      - name: Build and Run Docker Compose
+        run: |
+          docker compose -f ${{ github.workspace }}/Server/docker-compose.yml up -d
```
ChrisWeissmann, line 25: Can we maybe just bring up the services we need? Else we are introducing risk that one of the services fails and this pipeline fails. (looking at you Directus! 👀 )

.github/workflows/publish-swagger.yml
```diff
@@ -0,0 +1,77 @@
...
+
+      - name: Generate Swagger JSON
+        run: |
+          echo "Fetching Swagger JSON..."
+          curl --retry 3 --retry-delay 3 --retry-all-errors -o swagger.json http://localhost:8080/docs/json
+
+      - name: Validate Swagger JSON
+        id: validate-swagger
```
ChrisWeissmann, line 33: This is just a thought, and it might be a bit overboard here. But if you actually want to validate it, Swagger has thing to send the file to and it tells you if its all good:
check out: https://github.com/swagger-api/validator-badge

```shell
curl --data-binary @swagger.yaml -H 'Content-Type:application/yaml' https://validator.swagger.io/validator/
```

.github/workflows/publish-swagger.yml
elliot-sabitov: do we want to test this by adding a feature branch or a tag to confirm that it is doing what we expect?

langiole: @elliot-sabitov do you mean test the workflow itself? If so, it has been tested! I had this set so it would run the workflow whenever I updated this PR.

.github/workflows/publish-swagger.yml
langiole: I was experiencing authorization issues when I had this.

I believe this could be resolved if we update the repository settings to give the workflow write permissions but we are unable to do so. This is why we had to use an SSH key.

.github/workflows/publish-swagger.yml
langiole: I'll wait for others to chime in before making this change.

Only reason is because this can introduce additional overhead for something that is autogenerated so will likely be valid most of the time.

Here is example of response from this cURL when swagger is valid:
```MACUS019:~ leeangioletti$ curl -X POST -d @swagger.json -H 'Content-Type:application/json' https://validator.swagger.io/validator/debug
{"schemaValidationMessages":[]}
```

So script would need to be capable of parsing JSON (unless we just do a string comparison) which requires an install of [jq](https://jqlang.org/).

thanaParis (approved):
Approved!

PR #356, opened 2025-02-11
Feature/generelize transactions

eddygrebt:
⚠️ This branch is based on feature/init-generalize: https://github.com/citrineos/citrineos-core/pull/346

⚠️ This branch is based on Issue 72: Generalize MeterValue: https://github.com/citrineos/citrineos-core/pull/352/

This PR is to generalize the transaction db entity.

PR #366, opened 2025-02-14
Issue 89: Implement StartTransaction

lydiazcheng:
⚠️ This branch is based on implement transaction: https://github.com/citrineos/citrineos-core/pull/356 . 

## Changes
This PR is to implement the OCPP 1.6 StartTransaction handler.
1. Create new `StartTransaction` table and associate with `Transaction`
2. Implement new `StartTransaction` handler
3. Refactor and fix the error (below) in `getNextSequenceValue` for `ChargingStationSequence`
```
SequelizeDatabaseError  null value in column "createdAt" of relation "ChargingStationSequences" violates not-null constraint, SequelizeDatabaseError, error: null value in column "createdAt" of relation "ChargingStationSequences" violates not-null constraint, error: null value in column "createdAt" of relation "ChargingStationSequences" violates not-null constraint, INSERT INTO "ChargingStationSequences" ("stationId", "type", "value")
                        VALUES ('cp001', 'transactionId', 0) RETURNING "value";, [object Object]
```

### Case 1: Start transaction successfully.
Prepare: Make sure the related charging station, connector and authorization db entities exist
1. Send the StartTransaction request to Citrine and receive the response 
Screenshot 2025-02-14 at 3 49 23 PM

2. Transaction and StartTransaction are stored in db 
Screenshot 2025-02-14 at 3 49 35 PM
Screenshot 2025-02-14 at 3 50 03 PM

### Case 2: Concurrent transaction
Send another StartTransaction request when there is an active transaction with the same idToken
Screenshot 2025-02-14 at 3 52 34 PM

### Case 3: StartTransaction req has invalid idToken
Screenshot 2025-02-14 at 3 53 59 PM

### Case 4: Start transaction on a non-existing connector
Screenshot 2025-02-14 at 4 00 17 PM


01_Data/src/layers/sequelize/model/Authorization/IdToken.ts
```diff
@@ -3,13 +3,23 @@
...
 
-@Table
+@Table({
+  indexes: [
+    {
+      unique: true,
+      fields: ['idToken'],
+      where: {
+        type: null,
+      },
```
langiole, line 19: What exactly is this index for? We are only indexing the idToken column for entries where type == null?

01_Data/src/layers/sequelize/model/TransactionEvent/Transaction.ts
```diff
@@ -50,6 +51,9 @@ export class Transaction extends Model {
   @HasMany(() => MeterValue)
   declare meterValues?: MeterValue[];
 
+  @HasOne(() => StartTransaction)
+  declare startTransaction?: StartTransaction;
```
langiole, line 55: I think my only gripe with this StartTransaction entity is that this will result in only 1.6 DB entries having this field populated while the other entries do not. Can see this being confusing.

Is it unreasonable to instead store this in a Started `TransactionEvent`?

01_Data/src/layers/sequelize/model/Authorization/IdToken.ts
lydiazcheng: Hi Lee, thanks for reviewing! This part is to make sure that `type` is `null` can’t have duplicate idToken values.  This [issue](https://github.com/citrineos/citrineos/issues/66) might also provide more info.

01_Data/src/layers/sequelize/model/TransactionEvent/Transaction.ts
lydiazcheng: I would prefer not to mix this 2 concepts and add our own customized translation. They both are protocol specific and also have some different fields. If we extend `TransactionEvent` with 1.6 fields and then translate `StartTransaction` to `TransactionEvent`, both 1.6 and 2.0.1 users might feel confused.

03_Modules/Transactions/src/module/TransactionService.ts
```diff
@@ -142,6 +148,99 @@ export class TransactionService {
...
+        new Date() > new Date(idTokenInfo.cacheExpiryDateTime)
+      ) {
+        response.idTagInfo.status =
+          OCPP1_6.StartTransactionResponseStatus.Expired;
+        return response;
+      }
+
+      // Check concurrent transactions
+      const activeTransactions =
+        await this._transactionEventRepository.readAllActiveTransactionsIncludeStartTransactionByIdToken(
```
langiole, line 200: Bit of a performance hit here needing to dig into Transaction, StartTransaction, and IdToken tables.

03_Modules/Transactions/src/module/module.ts
```diff
@@ -586,4 +595,60 @@ export class TransactionsModule extends AbstractModule {
...
+      await this.sendCallResultWithMessage(message, response);
+    } else {
+      const transactionId = await this._idGenerator.generateRequestId(
+        stationId,
+        ChargingStationSequenceType.transactionId,
+      );
+      this._logger.info(`Generated new transactionId: ${transactionId}`);
+      // Create transaction
+      const newTransaction =
+        await this._transactionEventRepository.createTransactionByStartTransaction(
```
langiole, line 627: I almost feel like sequence generation should just be handled within the `createTransactionByStartTransaction`?

Server/src/index.ts
```diff
@@ -646,6 +646,8 @@ export class CitrineOSServer {
       this._repositoryStore.componentRepository,
       this._repositoryStore.locationRepository,
       this._repositoryStore.tariffRepository,
+      this._repositoryStore.reservationRepository,
+      this._idGenerator,
```
langiole, line 650: Little but confused as to why these weren't here before considering these repos already existed.

langiole (approved):
Looks really good, Zihe!

Just some minor feedback.

Server/src/index.ts
lydiazcheng: Might be missed out. 🤔

lydiazcheng:
> Looks really good, Zihe!
> 
> Just some minor feedback.

Thank you, Lee! I updated the code based on the feedback.

PR #367, opened 2025-02-18
Implementing get Configuration Message Endpoint.

eddygrebt:
Implementing get Configuration Message Endpoint Handler.


03_Modules/Configuration/src/module/1.6/MessageApi.ts
```diff
@@ -128,6 +128,78 @@ export class ConfigurationOcpp16Api
...
+        const keys = request.key || [];
+
+        const sendBatches = async (batches: string[][]) => {
+          return Promise.all(
+            batches.map(async (batch) => {
+              const correlationId = uuidv4();
+              await this._module.callMessageRepository.create(
+                CallMessage.build({ correlationId }),
+              );
+              return await this._module.sendCall(
```
lydiazcheng, line 175: Can we find a way to add the batch index and the identifier into the confirmation payload? If anything is wrong, we can know which part exactly.

eddygrebt:
On my commit I said GetConfirmations Handler, but it is actually GetConfigurations, sorry...

PR #368, opened 2025-02-18
Feature/issue 85 authorize

ChrisWeissmann:
Added Authorize support for 1.6 messages. 
The issue we run into, is that 1.6 idTags don't have the concept of a type, which breaks with our existing look up.
I have added that if the type is null, we expect that there could be multiple authorizations fitting the query. 
Current implementation is conservative in that we check if any of the idTokens is not accepted we reject the authorize.
This approach may need some more discussion.

Screenshot 2025-02-19 at 12 23 52 AM

### DIFF
--- a/01_Data/src/layers/sequelize/mapper/1.6/AuthorizationMapper.ts
+++ b/01_Data/src/layers/sequelize/mapper/1.6/AuthorizationMapper.ts
@@ -1,0 +1,29 @@
+/*
+ * // Copyright Contributors to the CitrineOS Project
+ * //
+ * // SPDX-License-Identifier: Apache 2.0
+ *
+ */
+
+import {
+  OCPP1_6,
+} from '@citrineos/base';
+
+export class AuthorizationMapper {
+
+
+  static toIdTagInfoStatus(status: string): OCPP1_6.AuthorizeResponseStatus {
+    switch (status) {
+      case "Accepted":
+        return OCPP1_6.AuthorizeResponseStatus.Accepted;
+      case "Blocked":
+        return OCPP1_6.AuthorizeResponseStatus.Blocked;
+      case "Expired":
+        return OCPP1_6.AuthorizeResponseStatus.Expired;
+      case "Invalid":
+        return OCPP1_6.AuthorizeResponseStatus.Invalid;
+      default:
+        throw new Error('Unknown IdTagInfoStatus status');
+    }
+  }
+}
--- a/01_Data/src/layers/sequelize/mapper/1.6/AuthorizationMapper.ts
+++ b/01_Data/src/layers/sequelize/mapper/1.6/AuthorizationMapper.ts
@@ -5,22 +5,18 @@
  *
  */
 
-import {
-  OCPP1_6,
-} from '@citrineos/base';
+import { OCPP1_6 } from '@citrineos/base';
 
 export class AuthorizationMapper {
-
-
   static toIdTagInfoStatus(status: string): OCPP1_6.AuthorizeResponseStatus {
     switch (status) {
-      case "Accepted":
+      case 'Accepted':
         return OCPP1_6.AuthorizeResponseStatus.Accepted;
-      case "Blocked":
+      case 'Blocked':
         return OCPP1_6.AuthorizeResponseStatus.Blocked;
-      case "Expired":
+      case 'Expired':
         return OCPP1_6.AuthorizeResponseStatus.Expired;
-      case "Invalid":
+      case 'Invalid':
         return OCPP1_6.AuthorizeResponseStatus.Invalid;
       default:
         throw new Error('Unknown IdTagInfoStatus status');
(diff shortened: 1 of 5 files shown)


01_Data/src/layers/sequelize/repository/Authorization.ts
```diff
@@ -179,12 +179,28 @@ export class SequelizeAuthorizationRepository extends SequelizeRepository<Author
...
+    const idTokenWhere: any = {};
+    if (queryParams.idToken) {
+      // exact match
+      idTokenWhere.idToken = queryParams.idToken;
+      // or partial match:
+      // idTokenWhere.idToken = { [Op.like]: `%${queryParams.idToken}%` };
+    }
+
+    // only include type if it's provided
+    if (queryParams.type != null) {
```
langiole, line 193: Minor: does doing `queryParams.type` not achieve same thing?

langiole (approved):
Overall, looks good! Very minor feedback.

01_Data/src/layers/sequelize/mapper/1.6/AuthorizationMapper.ts
```diff
@@ -0,0 +1,29 @@
...
+ *
+ */
+
+import {
+  OCPP1_6,
+} from '@citrineos/base';
+
+export class AuthorizationMapper {
+
+
```
lydiazcheng, line 14: Super minor: there are 2 empty lines

01_Data/src/layers/sequelize/repository/Authorization.ts
```diff
@@ -179,12 +179,28 @@ export class SequelizeAuthorizationRepository extends SequelizeRepository<Author
   }
 
   private _constructQuery(queryParams: AuthorizationQuerystring): object {
+    //1.6 doens't have the concept of token type. But we need to support token type for 2.0.1 messages.
```
lydiazcheng, line 182:
```suggestion
    // 1.6 doesn't have the concept of token type. But we need to support token type for 2.0.1 messages.
```

03_Modules/EVDriver/src/module/module.ts
```diff
@@ -771,6 +772,59 @@ export class EVDriverModule extends AbstractModule {
...
+    const response: OCPP1_6.AuthorizeResponse = {
+      idTagInfo: {
+        status: OCPP1_6.AuthorizeResponseStatus.Invalid,
+      },
+    };
+    try {
+      const authorizations =
+        await this._authorizeRepository.readAllByQuerystring({
+          idToken: request.idTag,
+          type: null, //explicitly ignore type
```
lydiazcheng, line 790: 👍

01_Data/src/layers/sequelize/repository/Authorization.ts
```diff
@@ -179,12 +179,28 @@ export class SequelizeAuthorizationRepository extends SequelizeRepository<Author
...
+      // or partial match:
+      // idTokenWhere.idToken = { [Op.like]: `%${queryParams.idToken}%` };
+    }
+
+    // only include type if it's provided
+    if (queryParams.type != null) {
+      idTokenWhere.type = queryParams.type;
+    }
+
+
```
lydiazcheng, line 197: Super minor: 1 empty line would be perfect

lydiazcheng (commented):
The PR looks good! Only super minor points and the authorize logic need be updated.

01_Data/src/layers/sequelize/repository/Authorization.ts
lydiazcheng: I think Lee means `if (queryParams.type)`?

ChrisWeissmann: Yea, I think we can also use what @lydiazcheng mentions.

03_Modules/EVDriver/src/module/module.ts
```diff
@@ -264,8 +265,11 @@ export class EVDriverModule extends AbstractModule {
       },
     };
 
-    if (message.payload.idToken.type === OCPP2_0_1.IdTokenEnumType.NoAuthorization) {
-      response.idTokenInfo.status = OCPP2_0_1.AuthorizationStatusEnumType.Accepted;
+    if (
+      message.payload.idToken.type === OCPP2_0_1.IdTokenEnumType.NoAuthorization
+    ) {
+      response.idTokenInfo.status =
+        OCPP2_0_1.AuthorizationStatusEnumType.Accepted;
```
lydiazcheng, line 272: Are these format changes due to the pre-commit git action for running prettier?

lydiazcheng (approved):

03_Modules/EVDriver/src/module/module.ts
ChrisWeissmann: No, those aren't merged yet. I let it format on my local machine using prettier in my IDE. This is currently the correct format as a line not allowed to be longer than 80 char. Take a look at my PR #372 . I suggest moving to 100 chars to have a little less wrapping.

PR #371, opened 2025-02-20
Chore: enforce prettier rules

ChrisWeissmann:
Add prettier to be part of linting rules that the github action runs when linting. This is done using the prettier eslint plugin. 
Running `lint-fix` now also fixes prettier issues.

To avoid formatting noise in future PRs, we also introduce husky to run prettier in a pre-commit hook. 

Also removed duplicate .prettierrc file causing discrepancies.


elliot-sabitov (approved):

thanaParis (approved):
approved

PR #375, opened 2025-02-21
Issue 113: Persist Config

langiole:
Added ability to fetch and persist configuration from S3 bucket.

Removed localstack in favor of minio due to support of persistence via data mounting.

When `APP_ENV=s3`, configuration will be fetched from the specified S3 bucket. The bucket will automatically be created if it does not exist. The `local.ts` is used as the default configuration loaded into the S3 bucket.


00_Base/src/config/S3Service.ts
```diff
@@ -0,0 +1,93 @@
...
+
+    private static getClient(): S3Client {
+        if (!this.instance) {
+            const isLocal = process.env.AWS_LOCAL === "true";
+
+            const s3Config = {
+                region: process.env.AWS_REGION || "us-east-1",
+                endpoint: isLocal ? process.env.AWS_S3_ENDPOINT : undefined,
+                forcePathStyle: isLocal,
+                credentials: {
```
ChrisWeissmann, line 16: Take a look at `@aws-sdk/credential-providers` seems there should be a way to easily pick up AWS credentials not only via env variables. This way when deployed in AWS it just picks them up.

00_Base/src/config/S3Service.ts
```diff
@@ -0,0 +1,93 @@
...
+export class S3Service {
+    private static instance: S3Client;
+
+    private static getClient(): S3Client {
+        if (!this.instance) {
+            const isLocal = process.env.AWS_LOCAL === "true";
+
+            const s3Config = {
+                region: process.env.AWS_REGION || "us-east-1",
+                endpoint: isLocal ? process.env.AWS_S3_ENDPOINT : undefined,
```
ChrisWeissmann, line 14: This locks us into AWS. We should have an interface with multiple implementations for different providers.

ChrisWeissmann (commented):
The formatter should be run. I also have some concerns about being locked in here.

package-lock.json
langiole, line 1: Should this get reverted?

00_Base/src/config/S3Service.ts
langiole: Seems as though this is not even needed.

Was able to load creds set in ~/.aws without this depdency.

00_Base/src/config/S3Service.ts
langiole: Abstracted this away with most recent changes 👍

Server/src/config/envs/local.ts
```diff
@@ -144,6 +144,18 @@ export function createLocalConfig() {
...
       },
+      configStorage: {
+        type: 's3',
+        s3: {
+          endpoint: 'http://localhost:9000',
+          bucketName: 'citrineos-s3-bucket',
+          keyName: 'local-config.json',
+        },
+        local: {
+          fileName: 'local-config.json',
```
langiole, line 155: Prefixed `config.json` with env to avoid conflict if user tries loading both locally and via docker since they both use the same bucket.

E.g. without this prefix, if user starts server locally, `config.json` will be created and saved in /data. If user then starts server with docker, this `config.json` will get loaded in which would be no good as this `config.json` was generated from `local.ts`.

Server/src/config/envs/local.ts
```diff
@@ -144,6 +144,18 @@ export function createLocalConfig() {
           },
         },
       },
+      configStorage: {
```
langiole, line 147: Decided to put the config for storage in the config ts files.

Doing this means we must make call to createLocalConfig/createDockerConfig before attempting to load config from storage since we need to retrieve the config for the storage in the first place.

Server/src/config/envs/local.ts
```diff
@@ -144,6 +144,18 @@ export function createLocalConfig() {
...
+        type: 's3',
```
langiole, line 148: This is where you specify whether to use s3 or local storage.

Server/package.json
```diff
@@ -46,7 +46,8 @@
     "@fastify/cors": "10.0.1",
     "cross-env": "7.0.3",
     "sqlite3": "5.1.6",
-    "ws": "8.17.1"
+    "ws": "8.17.1",
+    "@aws-sdk/client-s3" : "3.750.0"
```
ChrisWeissmann, line 50: I think you can kick this one out again, since you are doing all the AWS S3 stuff in Base.

Server/docker-compose.yml
```diff
@@ -78,19 +78,16 @@ services:
       - /usr/local/apps/citrineos/03_Modules/Reporting/dist/
       - /usr/local/apps/citrineos/03_Modules/SmartCharging/dist/
       - /usr/local/apps/citrineos/03_Modules/Transactions/dist/
+      - ~/.aws:/root/.aws
```
ChrisWeissmann, line 81: Not sure if this works for windows users. It also only works if you have the AWS CLI installed.

ChrisWeissmann (approved):

Server/docker-compose.yml
langiole: I believe the alternative would be to add the AWS env vars onto the citrine docker container definition if that's what you're suggesting we do instead.

Server/docker-compose.yml
```diff
@@ -102,26 +107,43 @@ services:
...
-      interval: 5s
-      timeout: 5s
+      test: ["CMD", "curl", "-f", "http://localhost:9000/minio/health/live"]
+      interval: 30s
+      timeout: 10s
       retries: 5
-      start_period: 30s
+      start_period: 20s
+
+  # Initialize MinIO with bucket and user - with improved logging
```
langiole, line 133: Don't think this comment is still relvant

thanaParis (approved):
approved

PR #376, opened 2025-02-23
adding OCPPMessage

thanaParis:
New OCPPMessage collection! StationId and correlationId columns indexed but not unique for quick lookups. Message JSONB column ensures advanced query support if needed, and easy access to request message fields for response handling.

changes to support:

- Explicit mapping from string to CallAction enums now done through mapToCallAction method
- Protocol now required for all message broker messages rather than just incoming messages from station
- Network hook method signature changed return type from boolean to void in order to better reflect actual usage--if unsuccessful, the method should always throw an error


03_Modules/OcppRouter/src/module/webhook.dispatcher.ts
```diff
@@ -67,17 +78,100 @@ export class WebhookDispatcher {
...
+        ['origin', origin],
+        ['timestamp', timestamp],
+        ['protocol', protocol],
+        ['action', callAction],
+      ]);
+
+      const promises: Promise<any>[] =
+        this._onMessageCallbacks
+          .get(identifier)
+          ?.map((callback) => callback(message, info)) ?? [];
```
lydiazcheng, line 104: What does this part do?

lydiazcheng (approved):
The improvement looks great. Just have a question listed below.

03_Modules/OcppRouter/src/module/webhook.dispatcher.ts
thanaParis: This is where webhook subscription callbacks are triggered. If an upstream service subscribes to sent and/or received messages with a url, a promise that POSTs to that url is added to the list of callbacks and then triggered here whenever a message is sent or received.

lydiazcheng: Got it. Thank you, Thana!

PR #377, opened 2025-02-23
Feature/issue 114 replacing request collections

thanaParis:
Removing CallMessage collection. Altering ChangeConfiguration to use OCPPMessage. Altering Reservation handling to use OCPPMessage.


01_Data/src/interfaces/repositories.ts
```diff
@@ -186,7 +185,7 @@ export interface IReservationRepository extends CrudRepository<Reservation> {
   createOrUpdateReservation(reserveNowRequest: OCPP2_0_1.ReserveNowRequest, stationId: string, isActive?: boolean): Promise<Reservation | undefined>;
 }
 
-export interface ICallMessageRepository extends CrudRepository<CallMessage> {}
+export interface IOCPPMessageRepository extends CrudRepository<OCPPMessage> {}
```
ChrisWeissmann, line 188: Seems this can be left out since it's already present in line 190

ChrisWeissmann (commented):
Just to be sure I am on the right track here, as it seems it's a fairly significant change.
The core idea is to move away from a dedicated callMessage repository (removing it all together) where we were tracking correlation Id to ReservationId as well as ChangeConfigurations. And utilize our new table of all OCPP messages and find the correlationId from there. 

Left a small comment and maybe we can resolve the linter error popping up surrounding unused imports.

thanaParis:
> Just to be sure I am on the right track here, as it seems it's a fairly significant change. The core idea is to move away from a dedicated callMessage repository (removing it all together) where we were tracking correlation Id to ReservationId as well as ChangeConfigurations. And utilize our new table of all OCPP messages and find the correlationId from there.

Yes, that's the idea!

> Left a small comment and maybe we can resolve the linter error popping up surrounding unused imports.

Which unused imports are you talking about? If you mean the ones in ChargingProfile, ChargingSchedule, etc, listed at the bottom of the file diff, none of those were added in this PR so they won't be fixed here either. We could open up a dedicated PR to fix that.

lydiazcheng (approved):
The change looks good!

PR #380, opened 2025-03-07
Feature/system config user preferences

thanaParis:
User preferences for telemetry added to system config now that system config persists across startup.


lydiazcheng (approved):
The change looks good!

PR #381, opened 2025-03-18
FIX: Update charge session password API

brianedwards00:
If set on charger, do not update password on charger (this logic is backwards).
It should be: If set on charger, update password on charger.


brianedwards00:
Hm, according to [the documentation](https://citrineos.github.io/latest/core-concepts/connecting-a-charger/#using-security-profile-1):

"Set setOnCharger to true if the password has already been updated on the charger and you only need to update the device model associated with the station id. If you need to update the password on the charger (OCPP A01 use case), set setOnCharger to false."

That's a little confusing because I thought it meant "set password on charger" and not the actual "is password set on charger".

PR #382, opened 2025-03-19
Feature/unify file storage

thanaParis:
- moving config store implementations from 00_Base into 02_Util
- combine config store implementations with IFileAcess implementations
- splitting IFileAccess into IFileAccess (for generating upload urls) and IFileStorage (for saving and retrieving files)
- altering config store to extend IFileStorage and adjusting implementations to match
- adding local and s3 file access configuration to system config, expanding s3 config for cloud
- moving directus configuration to file access in system config
- adding system config file for setup with directus file access instead of minio
- Server initialization altered to create file access according to system config


lydiazcheng (approved):
Looks good!

PR #386, opened 2025-03-27
Use better healthcheck to ensure rabbitmq is up before citrine

SohamG:
Atleast for me the previous healthcheck of `ping` would cause `citrine` to start before rabbitmq was ready to accept connections, resulting in a crash.

License same as project.


thanaParis (approved):
Approved! Thank you for the PR

PR #387, opened 2025-03-27
Feature/expose ocpp verison

ChrisWeissmann:
Introduces the protocol version to the charging station entity.
This allows outside parties to know which endpoints to call depending on what the charger has registered as their protocol.


thanaParis (approved):
approved

PR #393, opened 2025-04-11
fix: error with 1.6 StopTransaction with IdToken

elliot-sabitov:
- fix: error with 1.6 StopTransaction incorrectly fetching authorization resulting in `More than one value found for query: {\"idToken\":\"IDTOKENVALUE\"}` error

I was able to reproduce the issue locally and with this adjustment, the error goes away
Screenshot 2025-04-11 at 11 30 24 AM
Screenshot 2025-04-11 at 11 30 15 AM

After fix
Screenshot 2025-04-11 at 11 32 44 AM


thanaParis (approved):
approved!

PR #405, opened 2025-05-09
Adjusting System Config, requests and responses are now inserted on systemConfig

eddygrebt:
Modifying all modules to receive the responses and requests CallActions with the constructor method.
Adjusting System Config to provide values for CallActions.
Providing responses and requests CallActions via env configs.

Making sure that all the system config is being used properly:

image


thanaParis (approved):
approved

PR #407, opened 2025-05-21
Adding new boolean field to Authorization, validating before hasConcurrent check in authorizeIdToken function

eddygrebt:
- Adding concurrentTx bool field to Authorization Model
- Validating Authorization.concurrentTx before the whole segment for hasConcurrent.


thanaParis (approved):
approved!

PR #408, opened 2025-05-23
FEATURE: Websocket Configurations CRD

yuvenus:
# Description
This PR adds create, read, and delete endpoints to the AdminApi in OcppRouter to handle websocket create, read, and delete. Update was left out on purpose because it can be a bit complicated and we want to handle that separately. The create/delete endpoints will not only add to/remove from the websocket configs in the module but also the one in file storage, so that the next time the server is restarted, you won't lose the configs.

## Create websocket config
* `createWebsocketConfiguration` will create a websocket configuration as long as a previous one of the same ID doesn't exist. It will update both the module and file storage configs.
  * If there is a conflict, it will throw a `BadRequestError` explaining that a websocket config with the same ID already exists

## Read websocket config
* `getWebsocketConfigurations` will return a list of websocket configurations UNLESS you pass a specific websocket ID in the query params, in which case it'll return just one websocket configuration
  * There's a TODO to eventually filter the websocket configurations list by tenantId once tenantId is implemented in websocket configurations in a separate feature.
* There's a bit of redundancy with the endpoint that returns the system config, but I figured it wouldn't hurt -- if it does hurt, let me know and I can simplify this endpoint down to just returning a websocket config by id only 😄  

## Delete websocket config
* `deleteWebsocketConfiguration` will delete a websocket configuration by ID. 
  * If it finds the websocket config in the list, it'll delete it from the module and file storage configs.
  * If it doesn't find the websocket config in the list, it does nothing and returns OK as normal.

# Screenshots

## New endpoints in Swagger UI
image
image
image
image


thanaParis (approved):
approved

PR #412, opened 2025-06-04
Reconnect logic implementation for message broker

Adex9ja:
## Description

create reconnect logic for message broker. This is a fix for issue: citrine server cannot resume to normal after amqbroker restart [citrine-core] #[139](https://github.com/citrineos/citrineos/issues/139)

## What has been changed?

- Added automatic RabbitMQ reconnection logic
- Removed duplicate event listeners on reconnect
- Ensured only one reconnect attempt at a time
- Added support for an abort signal to break the retry loop cleanly on shutdown.

## How was this tested?

1. Start all services and connect a charging station
2. Stop and restart the RabbitMQ broker
3. Verify the charging station reconnects and resumes operation


thanaParis (changes requested):
Looks good, one small change recommended in a comment. Also, please merge in 1.7.0 to make sure everything works okay and revert the changes to package.json/package-lock.json.

03_Modules/OcppRouter/src/module/router.ts
```diff
@@ -112,6 +112,7 @@ export class MessageRouterImpl extends AbstractMessageRouter implements IMessage
   ): Promise<boolean> {
     const dispatcherRegistration = this._webhookDispatcher.register(tenantId, stationId);
 
+    await this._handler.initConnection();
```
thanaParis, line 115: this design will open a connection to rabbitMQ per websocket connection, which will break rather quickly. please move this into the constructor of this class.

thanaParis (changes requested):
small change requested!

PR #416, opened 2025-06-16
1.7.1

thanaParis:


ChrisWeissmann:
With this release we are:
- adding a native `arm64` Docker release image
- adding specific `arm64` docker compose 
- adding and npm script to spin up an Everest instance running OCPP 1.6
- adding that when local by pass is `true` for authenticating on the API, no `authorization` header is required
- adding support for the `Update firmware` message with OCPP 1.6

PR #418, opened 2025-06-18
Message broker circuit breaker Implementation

deye9:
Implemententation for message broker circuit breaker


00_Base/src/util/CircuitBreaker.ts
```diff
@@ -0,0 +1,75 @@
...
+
+  reset() {
+    this._state = 'OPEN';
+    this.reconnectDelay = 1000;
+    if (this.timer) clearTimeout(this.timer);
+    this.timer = null;
+    this.emitStateChange('OPEN');
+  }
+
+  private scheduleReconnect() {
```
thanaParis, line 56: I think this method doesn't belong here. the circuit breaker itself should only be concerned with state management. I'd prefer the reconnect logic remain isolated in the receiver/sender implementations in 02_Util.

00_Base/src/interfaces/modules/CircuitBreaker.ts
```diff
@@ -0,0 +1,14 @@
+export type CircuitBreakerState = 'OPEN' | 'FAILING' | 'CLOSED';
+
+export interface CircuitBreakerOptions {
+  maxReconnectDelayMs?: number; // default: 30000
+  onStateChange?: (state: CircuitBreakerState, reason?: string) => void;
+}
+
+export interface ICircuitBreaker {
```
thanaParis, line 8: I don't think this interface is necessary. the circuit breaker should be technology agnostic, a single concrete implementation should be acceptable and simpler to maintain.

00_Base/src/interfaces/router/AbstractRouter.ts
```diff
@@ -74,10 +89,63 @@ export abstract class AbstractMessageRouter implements IMessageRouter {
       ? logger.getSubLogger({ name: this.constructor.name })
       : new Logger<ILogObj>({ name: this.constructor.name });
 
+    this._circuitBreaker = new CircuitBreaker({
```
thanaParis, line 92: I'd prefer to allow for dependency injection. We can instantiate the CircuitBreaker if one isn't passed into the constructor.

00_Base/src/interfaces/router/AbstractRouter.ts
```diff
@@ -74,10 +89,63 @@ export abstract class AbstractMessageRouter implements IMessageRouter {
...
+        this._logger.warn('Circuit breaker failing', reason);
+        break;
+      default:
+        this._logger.error('Unknown circuit breaker state', state, reason);
+        return;
+    }
+  }
+
+  // Subclasses can override for custom actions
+  protected onCircuitBreakerClosed(reason?: string) {
```
thanaParis, line 121: I think it would be a good idea to add some unit tests for this logic. I'm wondering if this is perhaps the wrong place to make these changes. I know I asked for an implementation that would include the AbstractModule, but since there's no default behavior for that, and this router logic requires references to the websocket server, I'd like for you to move this to 03_Modules/OcppRouter/src/module/router.ts and don't worry about touching any module classes for now. That should simplify the PR and let you add tests in the OcppRouter package which should be easier than doing them here.

thanaParis (changes requested):

00_Base/src/interfaces/router/AbstractRouter.ts
thanaParis, line 1: I don't think this should go here. It can go directly into the implementation in OcppRouter.

00_Base/src/interfaces/modules/AbstractModule.ts
thanaParis, line 1: I'd prefer not to make any changes here until we have a use case. you can skip this file.

00_Base/src/interfaces/modules/CircuitBreaker.ts
thanaParis, line 1: The CircuitBreaker logic should go in one file. Either here or in /util.

thanaParis (changes requested):

00_Base/src/interfaces/modules/CircuitBreaker.ts
deye9: @thanaParis resolved

00_Base/src/interfaces/modules/AbstractModule.ts
```diff
@@ -53,9 +53,6 @@ export abstract class AbstractModule implements IModule {
     this._sender = sender;
     this._eventGroup = eventGroup;
     this._cache = cache;
-
-    // Set module for proper message flow.
-    this.handler.module = this;
```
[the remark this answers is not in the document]
deye9: Opps. Resolved

PR #421, opened 2025-06-20
Simplify Authorization Object

deye9:
Authorization Object simplification


01_Data/src/interfaces/repositories.ts
```diff
@@ -289,7 +289,7 @@ export interface ITransactionEventRepository extends CrudRepository<TransactionE
   ): Promise<Transaction[]>;
   readAllActiveTransactionsIncludeTransactionEventByIdToken(
     tenantId: number,
-    idToken: OCPP2_0_1.IdTokenType,
+    idToken: string, // was OCPP2_0_1.IdTokenType
```
thanaParis, line 292: This should be a new non-protocol-specific enum type, not just a string. you can use the values from the OCPP2_0_1 enum as a starting point.

01_Data/src/layers/sequelize/model/Authorization/Authorization.ts
```diff
@@ -19,26 +18,58 @@ export class Authorization extends BaseModelWithTenant implements AuthorizationR
...
+
+  @Column(DataType.JSON)
+  declare personalMessage?: any | null;
 
-  @ForeignKey(() => IdTokenInfo)
+  // Self-reference for groupIdToken
+  @ForeignKey(() => Authorization)
   @Column(DataType.INTEGER)
-  declare idTokenInfoId?: number | null;
+  declare groupIdTokenId?: number | null;
```
thanaParis, line 64: This is the right idea! it no longer makes sense to use this field name though. 'groupAuthorizationId' would be more accurate.

01_Data/src/layers/sequelize/model/Authorization/LocalListAuthorization.ts
```diff
@@ -41,19 +41,45 @@ export class LocalListAuthorization
   @Column(DataType.ARRAY(DataType.STRING))
   declare disallowedEvseIdPrefixes?: string[];
 
-  @ForeignKey(() => IdToken)
+  // Flattened fields from Authorization
```
thanaParis, line 44: this is the right idea, but we'll also need to change the class type so that it implements the new authorization object rather than the OCPP2_0_1 interface

migrations/20250620120000-flatten-authorization-structure.ts
thanaParis, line 1: This migration currently just drops the tables. This is not a migration--all the data in the tables would be lost. You will need to copy the data from the old tables into the new. This may require the creation of a temporary table:

1. Create temp table of old Authorization type
2. Copy data from current table into temp table
3. Alter current table (do not drop IdToken/IdTokenInfo/AdditionalInfo yet)
4. Push data from temp table + IdToken/IdTokenInfo/AdditionalInfo into altered table
5. Drop temp/IdToken/IdTokenInfo/AdditionalInfo tables

01_Data/src/layers/sequelize/model/Authorization/IdToken.ts
thanaParis, line 1: This is looking good--do remember to also delete AdditionalInfo and IdTokenAdditionalInfo

thanaParis (changes requested):

PR #422, opened 2025-06-24
Feature/seperate bootstrap envvars

thanaParis:


langiole:
Pulled and was able to get the server running. That being said, the logs show some strange behavior during startup that I do not fully understand. It attempts to load the config using a file path of `/`, which fails, but then it appears to try again with the `/Server` path and proceeds successfully. Not entirely sure why both attempts are happening.

langiole:
Also, seems like the docker compose needs its environment variables to be updated with this new prefix.

langiole:
Also, to replicate you will need to have you config.json deleted before start up in order to get the double invocation.

thanaParis:
Following up on this, I adjusted the bootstrap config to contain the database connection info in order to fix the above, which was caused by data migrations requiring the system config. Should all be fixed now @langiole

yuvenus (approved):
after pulling branch, env vars are being read in safe and sound.

will defer to @langiole for his particular issue as i was unable to replicate. 🙏

PR #426, opened 2025-06-27
feature: don't log the skipping of auth log for health endpoint to re…

ChrisWeissmann:
The health log is annoying in production and doesn't add much insight.


thanaParis (approved):
approved

PR #432, opened 2025-07-04
Feature/add multiplatform build

ChrisWeissmann:
Add building docker images specifically for arm in Github action workflow. 
Introduced new explicit deploy docker file for Directus to get around some docker context issues that arise when building for multiple architectures. 

Tested this with an explicit workflow that then was removed again.


thanaParis (approved):
approved!

PR #433, opened 2025-07-07
TC-39-cache-bug-rabbitmq

deye9:
Added cache cleanup: The unsubscribe method now properly removes the cache entry using this._cache.remove(cacheKey, CacheNamespace.Other) after successfully deleting all queues.

Improved variable scoping: Extracted the cacheKey variable to be reused for both reading from and removing from the cache.

Maintained error handling: The cache removal only happens when queues are successfully deleted, preserving the existing error handling logic.

Expected Behavior After Fix:
When a charging station disconnects and the unsubscribe method is called, all queues will be deleted from RabbitMQ AND the cache entry will be removed
When the same charging station reconnects, a fresh cache entry will be created
The cache will no longer accumulate stale queue references

You should see each queue deleted only once, not multiple times
This fix ensures that the ICache remains clean and prevents the memory leak that was occurring due to accumulating stale queue references.

### DIFF
--- a/02_Util/src/queue/rabbit-mq/receiver.ts
+++ b/02_Util/src/queue/rabbit-mq/receiver.ts
@@ -157,12 +157,9 @@ export class RabbitMqReceiver extends AbstractMessageHandler {
   }
 
   unsubscribe(identifier: string): Promise<boolean> {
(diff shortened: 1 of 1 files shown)


thanaParis (approved):
approved!

PR #164, opened 2024-07-11
Fix Charging Profile Push

lydiazcheng:
This PR is to support the fix in OCPI https://github.com/citrineos/citrineos-ocpi/pull/75

PR #434, opened 2025-07-09
#164 Allow "no auth header" for local development - default to dummy header

ChrisWeissmann:
feature: move auth header to not being needed for local bypass during…development. Also adds support for accepting customer headers, instead of only `Authorization` allowing for static api keys

This should address issue [#164](https://github.com/citrineos/citrineos/issues/146)


thanaParis (approved):
Approved!

PR #436, opened 2025-07-10
adding dtos to 01_Data

thanaParis:
- 01_Data classes now implement 00_Base dtos


yuvenus (approved):
approved! just one super, super minor thing: i noticed there are a few places left where the IBaseDto is './base.dto', when in others it's '../..'. do they need to be consistent or all good as is?

thanaParis:
> approved! just one super, super minor thing: i noticed there are a few places left where the IBaseDto is './base.dto', when in others it's '../..'. do they need to be consistent or all good as is?

I think this is fine for now, but thank you for point this out, I'll try to get it aligned before the release is finalized.

PR #438, opened 2025-07-10
Feature/authorization ocpi

thanaParis:
features
- Authorization status enum created that is specific to CitrineOS rather than to an OCPP protocol
- IdTokenType and AdditionalInfo also made generic
- IAuthorizer interface moved to 00_Base
- IAuthorizer interface refactored to return new generic Authorization status enum rather than partial authorization object that was specific to OCPP 2.0.1
- IAuthorizer interface now also used in OCPP 1.6 authorization flows
- Real time authorization refactored in order to not need special handling, now a normal implementation of IAuthorizer interface that is used like every other implementation of that in


deye9 (changes requested):
There are merge conflicts

02_Util/src/authorizer/RealTimeAuthorizer.ts
```diff
@@ -0,0 +1,101 @@
...
+    authorization: Authorization,
+    context: IMessageContext,
+  ): Promise<AuthorizationStatusEnumType> {
+    if (!authorization.realTimeAuthUrl) {
+      this._logger.debug(`No Realtime Auth URL from authorization ${authorization.id}`);
+      return authorization.status;
+    } else if (
+      !authorization.realTimeAuth ||
+      authorization.realTimeAuth === RealTimeAuthEnumType.Allowed
+    ) {
```
lydiazcheng, line 36: When realtime is allowed, why we skip the realtime auth call and return? I might misunderstand the meaning this `RealTimeAuthEnumType.Allowed` or it might be a typo of `AllowedOffline`?

03_Modules/EVDriver/src/module/module.ts
```diff
@@ -207,9 +206,9 @@ export class EVDriverModule extends AbstractModule {
       this._deviceModelRepository,
     );
 
-    this._authorizers = authorizers || [];
-    this._realTimeAuthorizer =
+    const _realTimeAuthorizer =
       realTimeAuthorizer || new RealTimeAuthorizer(this._locationRepository, this._logger);
+    this._authorizers = [_realTimeAuthorizer, ...(authorizers || [])];
```
lydiazcheng, line 211: 🥇

lydiazcheng (approved):
The change looks good! Just have a question.

PR #444, opened 2025-07-18
Tc 51: Updating Dtos

Adex9ja:


00_Base/src/interfaces/dto/evse.dto.ts
```diff
@@ -12,6 +12,7 @@ export interface IEvseDto extends IBaseDto {
   removed?: boolean;
   chargingStation?: IChargingStationDto;
   connectors?: IConnectorDto[] | null;
+  connectorId: number;
```
thanaParis, line 15: this would have to be an array, Evses do not have one connectorId--it's also not necessary to include here and could cause issues (since it's a HasMany relationship, only the connector needs to have the evseId not the other way around)

00_Base/src/ocpp/model/common/enums/index.ts
thanaParis, line 1: Don't put this here. It should go in `src/interfaces/dto/enum`

01_Data/src/layers/sequelize/repository/TransactionEvent.ts
```diff
@@ -208,6 +208,10 @@ export class SequelizeTransactionEventRepository
         finalTransaction = await existingTransaction.update(
           {
             isActive: value.eventType !== OCPP2_0_1.TransactionEventEnumType.Ended,
+            endTime:
```
thanaParis, line 211: this only works for 2.0.1. Please add logic for 1.6 in the StartTransaction and StopTransaction related messages in this class.

thanaParis (changes requested):
some changes needed

thanaParis (approved):
Approved!

thanaParis (changes requested):
Changes required for build issue.

thanaParis (approved):
Approved!

PR #446, opened 2025-07-21
Unable to create Authorizations with latest Core/OperatorUI

deye9:
Fix for Unable to create Authorizations with latest Core/OperatorUI


thanaParis (approved):
approved

PR #528, opened 2026-02-09
Ocpp 1.6 j set charging profile

StefanRiedAV:
as discussed here https://github.com/orgs/citrineos/discussions/176

Here's the missing implementation of the setChargingProfile for OCPP1.6 chargers. Also tested for negative values, discharging bidi chargers.

The module Smartcharging seems to subscribe always the setChargingProfile event as it exists in OCPP2.0.1 as well. The handler shall be in the evdriver for OCPP1.6. 
I've added it there and added an logging nops handler in the smartcharging module to avoid an error of missing implementation. Not sure if this is the way you like. Maybe filtering the subscription of the smartcharging module by protocol and command would be more elegant.


thanaParis (changes requested):
Contains changes related to https://github.com/citrineos/citrineos/issues/179
Please only put changes related to a single issue per PR.

Overlaps with https://github.com/citrineos/citrineos-core/pull/526
Please see the feedback I gave there.

PR #540, opened 2026-02-17
Handlers and modules 2.1

devonNewsomDevelop:
Adds collective OCPP_CALLACTION enum for all possible ocpp call actions to compartmentalize calling for versions that will use the same functionality.
Creates OCPP allowed actions record to parse out version action mismatches.

Creates exported list of all ocpp2 versions for usage.

Updates handler function to accept a list of versions instead of single version so that we can spin up handlers for different versions. Note: This will grow large eventually, one day. Sometime, we may want to tweak how these decorators work. Perhaps instead have one function that works for all common versions.

Changes modules to use 2.1 schema all where applicable. Updates modules to use the list of ocpp 2 versions where applicable

Updates message APIs and various other files to use collective ocpp call action

Tweaks typing for call action and call action schema (simplified)

Changes module maps to records
Changes accessors referencing these

Creates unified version list and exports it for use in schema maps, like websocketServerInputSchema

TODOs for further implementation


02_Util/src/certificate/CertificateAuthority.ts
```diff
@@ -134,11 +134,12 @@ export class CertificateAuthorityService {
...
   public async validateCertificateChainPem(
     certificateChainPem: string,
-  ): Promise<OCPP2_0_1.AuthorizeCertificateStatusEnumType> {
+  ): Promise<OCPP2_1.AuthorizeCertificateStatusEnumType> {
     const certificatePems: string[] = parseCertificateChainPem(certificateChainPem);
     this._logger.debug(`Found ${certificatePems.length} certificates in chain.`);
     if (certificatePems.length < 1) {
-      return OCPP2_0_1.AuthorizeCertificateStatusEnumType.NoCertificateAvailable;
+      return OCPP2_1.AuthorizeCertificateStatusEnumType.NoCertificateAvailable;
+      return OCPP2_1.AuthorizeCertificateStatusEnumType.NoCertificateAvailable;
```
thanaParis: Duplicate return statement

01_Data/src/interfaces/repositories.ts
```diff
@@ -396,11 +397,11 @@ export interface IVariableMonitoringRepository
   ): Promise<EventData>;
 }
 
-export interface IMessageInfoRepository extends CrudRepository<OCPP2_0_1.MessageInfoType> {
+export interface IMessageInfoRepository extends CrudRepository<OCPP2_1.MessageInfoType> {
   deactivateAllByStationId(tenantId: number, stationId: string): Promise<void>;
   createOrUpdateByMessageInfoTypeAndStationId(
     tenantId: number,
-    value: OCPP2_0_1.MessageInfoType,
+    value: OCPP2_1.MessageInfoType,
```
thanaParis: Is there a reason (here and elsewhere) that sometimes we're updating to 2.1 and sometimes we're using union types? it is that the union types are cases where they are mutually different whereas the updates are cases where there's no differences or 2.1 covers all of 2.0.1? It's a bit confusing trying to discern why union types aren't just used everywhere

03_Modules/OcppRouter/src/module/router.ts
```diff
@@ -247,11 +247,15 @@ export class MessageRouterImpl extends AbstractMessageRouter implements IMessage
...
+            case OCPPVersion.OCPP1_6: {
               errorCode = ErrorCode.FormationViolation;
               break;
             }
-            case 'ocpp2.0.1': {
+            case OCPPVersion.OCPP2_0_1: {
+              errorCode = ErrorCode.FormatViolation;
+              break;
+            }
+            case OCPPVersion.OCPP2_1: {
```
thanaParis, line 255: small thing, but if we combine the cases it's clear they're the same

```
            case OCPPVersion.OCPP2_0_1: 
            case OCPPVersion.OCPP2_1: {
              errorCode = ErrorCode.FormatViolation;
              break;
            }
```

Server/src/config/envs/local.ts
```diff
@@ -85,66 +79,63 @@ export function createLocalConfig() {
...
         endpointPrefix: '/reporting',
         responses: [
-          OCPP2_0_1_CallAction.CustomerInformation,
-          OCPP2_0_1_CallAction.GetLog,
-          OCPP2_0_1_CallAction.GetReport,
-          OCPP2_0_1_CallAction.GetBaseReport,
-          OCPP2_0_1_CallAction.GetMonitoringReport,
-          OCPP1_6_CallAction.GetDiagnostics,
+          OCPP_CallAction.CustomerInformation,
+          OCPP_CallAction.GetLog,
```
thanaParis, line 112: GetDiagnostics appears to be missing here

Server/src/config/envs/local.ts
```diff
@@ -85,66 +79,63 @@ export function createLocalConfig() {
...
+          OCPP_CallAction.GetMonitoringReport,
         ],
         requests: [
-          OCPP2_0_1_CallAction.LogStatusNotification,
-          OCPP2_0_1_CallAction.NotifyCustomerInformation,
-          OCPP2_0_1_CallAction.NotifyReport,
-          OCPP2_0_1_CallAction.SecurityEventNotification,
-          OCPP2_0_1_CallAction.NotifyMonitoringReport,
-          OCPP1_6_CallAction.DiagnosticsStatusNotification,
+          OCPP_CallAction.LogStatusNotification,
```
thanaParis, line 119: DiagnosticsStatusNotification is missing

eslint.config.js
```diff
@@ -51,6 +51,8 @@ export default tseslint.config(
       'eslint.config.js',
       '00_Base/json-schema-processor-1.6.js',
```
thanaParis: we might want to remove these now that the files are gone? also, we should probably adjust the readme in this pr if we're removing the processors.

thanaParis (changes requested):
looking strong, some small changes requested.

01_Data/src/interfaces/repositories.ts
devonNewsomDevelop: Ah yes. I can understand where you're coming from. I'll explain the reasoning and I'm curious about your thoughts.

For areas of the code where the 2.1 version is backwards compatible with the 2.0.1 version, I simply changed the references to use OCPP2.1 and delivers the message with 2.1's structure.

So if you're on protocol 2.0.1, but we know it is 2.1 compatible, to save code duplication the handlers simply use 2.1 to call methods from repositories like this, if that makes sense? This way we don't have to have one handler that calls 2.0.1 and one handler that calls for 2.1 with case statements and etc.

Right now, the way I have the handlers set up, they basically just make two copies of the same function and use it for both protocols. It's essentially syntax sugar for writing the same @asHandler twice if that makes sense? So in that definition, we would either need to make a case statement for the protocol or do as I have done and pass the highest version that is backwards compatible for all previous versions. This should scale well as more versions are created.

When there are differences and the code should diverge, like I believe Authorize will be for 2.1, then we will have one @AsHandler for 2.0.1 and one @AsHandler for 2.1, much like the original ticket was going to be.

03_Modules/OcppRouter/src/module/router.ts
devonNewsomDevelop: I thought that this might need to be different for a reason I didn't understand since it seemed like 1.6 was in its own section. I figured that perhaps the errorCode was different with each version. Should I also bring 1.6 into this fallthru?

Server/src/config/envs/local.ts
devonNewsomDevelop: I'm silly and thought this was referencing another part of the code. Will fix!

Server/src/config/envs/local.ts
devonNewsomDevelop: See comment about GetDiagnostics.

01_Data/src/interfaces/repositories.ts
devonNewsomDevelop: I realized I didn't address the union type part of the question. If you could point an example out to me that you're curious about, I can take a look and remember my reasoning and reflect on if it makes sense to me still!

thanaParis: I think we should create union type shorthands for these scenarios
reading the code, this looks like a problem--the reader doesn't know why only 2.1 types are being used in a handler for both.
we either need to use union types everywhere just for consistency, or we could create an 'OCPP2' namespace for common types. For an OCPP2 namespace, we could just make it a bunch of union types that we already know to be equivalent... I'm not sure what other options we may have
but this current approach will be confusing to maintain

devonNewsomDevelop: Hey hey. I have a little time to respond between moving stuff so I will go ahead and do that here. 

The thing is, there are going to be situations when the handling for 2.1 and 2.0.1 have to be different. That was my point. We won't be able to utilize union types for when the code should diverge. The places where I did not use a union type (or just send 2.1 by itself), we diverged between 2.1 and 2.0.1.

I was thinking maybe a way to make it more obvious that we are using a backwards compatible version would be to do something similar to the namespace stuff you suggested, except we have a variable set to the latest version of OCPP 2. 

At the current version, we could do OCPP2_LATEST = OCPPVersion.OCPP2.1 or whatever the enum is called

My worry is that if we pass union types, the handling logic will be inundated with case / if statements as we go forward. For 2.1, they add a lot of *optional* properties and I can foresee them doing this in future minor versions as well. If we do union types, we will have to add case / if statements to handle those optional params. If we send the latest version that works backwards, we can fill those out without worry about the version and send it back. The prior versions wont reference the optional properties and all should be well. The only time I would diverge this behavior is if an optional property was expensive to set for whatever reason.

That's my thoughts and reasoning, if you believe that union types are better after this consideration, I can go through and change how we have this set up.

thanaParis (approved):
approved!

PR #553, opened 2026-02-21
Fix #163, #164: Add null safety to RealTimeAuthorizer

rishabhvaish:
Fixes #163
Fixes #164

## Problem

The `RealTimeAuthorizer` has two runtime exceptions that crash the authorization flow instead of producing helpful error messages:

1. **#163 — Null `locationId` when EVSE has no Location**
   `chargingStation!.locationId!.toString()` throws `TypeError: Cannot read properties of undefined` when a charging station has no associated Location. This crashes the entire real-time auth flow.

2. **#164 — Malformed response causes crash**
   The response from the real-time auth URL is type-cast (`as RealTimeAuthorizationResponse`) without validation. If the external service returns an unexpected shape (e.g., `{error: "server error"}`), accessing `realTimeAuth.data.allowed` throws `TypeError: Cannot read properties of undefined (reading 'allowed')`.

### Null-safe `locationId` access (#163)
- Use optional chaining (`chargingStation?.locationId?.toString()`) instead of non-null assertions
- Log a warning when `locationId` is missing, but proceed with the auth request (the field is already optional in `RealTimeAuthorizationRequestBody`)
- The request body sends `locationId: undefined` which serializes as the field being absent — the external auth service can handle this

### Response validation (#164)
- Validate response structure before type casting: check for `data` object with `allowed` property
- On malformed response: log a descriptive error including the actual response JSON, then fall back to `AllowedOffline` behavior if configured (matching the existing `catch` behavior)
- Only type-cast and process the response after validation passes

## Testing

Both fixes handle edge cases that previously caused unhandled exceptions:

```
# Before fix #163:
TypeError: Cannot read properties of undefined (reading 'toString')

# After fix #163:
WARN: Charging station 'CP-001' has no associated location. Proceeding with Realtime Auth without locationId.

# Before fix #164:
TypeError: Cannot read properties of undefined (reading 'allowed')

# After fix #164:
ERROR: Malformed Realtime Auth response for authorization 4: expected { data: { allowed: string } } but received: {"error":"server error"}
```

### DIFF
--- a/02_Util/src/authorizer/RealTimeAuthorizer.ts
+++ b/02_Util/src/authorizer/RealTimeAuthorizer.ts
@@ -131,11 +131,20 @@ export class RealTimeAuthorizer implements IAuthorizer {
         }
       }
 
+      // Build locationId safely — EVSE may not have an associated Location (#163)
+      const locationId = chargingStation?.locationId?.toString();
+      if (!locationId) {
+        this._logger.warn(
+          `Charging station '${context.stationId}' has no associated location. ` +
+            `Proceeding with Realtime Auth for authorization ${authorization.id} without locationId.`,
+        );
+      }
+
       const payload: RealTimeAuthorizationRequestBody = {
         tenantPartnerId: authorization.tenantPartnerId!, // Required if authorization has RealTimeAuth
         idToken: authorization.idToken,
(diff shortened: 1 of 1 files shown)


thanaParis (approved):
approved

PR #558, opened 2026-02-25
fix/validation-issue-in-data-transfer-not-accepting-data-values

elliot-sabitov:
fix: DataTransfer schema data blocks to be more flexible on validation to resolve schema parsing errors during data transfer request, where previously any values for data seemed to not be accepted


thanaParis (approved):
approved

PR #560, opened 2026-02-26
feat: add maxChargingStations for public demo

vicky-ahmed:


thanaParis:
will not affect charging stations created from operator-ui.

PR #566, opened 2026-03-02
Fix inverted concurrentTransaction logic in authorization (#185)

### LINKED ISSUE #185: CitrineOS ConcurrentTX issue
•	citrineos-core

⸻

Version/Branch Information
	•	Component version/tag: v1.8.3

⸻

Describe the bug

When an authorization tag already has an active session, CitrineOS returns ConcurrentTx when attempting to start another session — even though the tag is configured with Allow Concurrent Transactions enabled.

Based on the configuration, the tag should be allowed to initiate multiple transactions simultaneously. However, CitrineOS still rejects the new StartTransaction request with ConcurrentTx.

This behavior makes it impossible to use the same tag for multiple connectors, even when explicitly configured to allow it.

⸻

To Reproduce
	1.	Connect an OCPP 1.6 charger to CitrineOS.
	2.	Create an authorization tag.
	3.	Enable “Allow Concurrent Transactions” for that tag.
	4.	Start a charging session using that tag (Session A).
	5.	While Session A is still active, attempt to start a second session using the same tag (Session B).
	6.	Observe that CitrineOS responds with ConcurrentTx.
	7.	Session B does not start.

⸻

Expected behavior

If “Allow Concurrent Transactions” is enabled for a tag, CitrineOS should return Accepted for additional StartTransaction requests using the same tag, even when another session is active.

Instead, the system returns ConcurrentTx regardless of the configuration.

rishabhvaish:
## Summary

- **OCPP 2.0.1 bug:** `concurrentTransaction === true` caused `ConcurrentTx` rejection when concurrent sessions were *allowed*. Tokens with the flag enabled got blocked; tokens without it bypassed the check entirely.
- **OCPP 1.6 bug:** The `concurrentTransaction` flag was never checked. Every token with an active transaction was unconditionally rejected with `ConcurrentTx`, regardless of whether concurrent use was permitted.
- **Root cause:** Condition inversion in `authorizeOcpp201IdToken` (line 126: `=== true` should be `!== true`) and missing guard in `authorizeOcpp16IdToken`.

## Changes

| File | Change |
|------|--------|
| `03_Modules/Transactions/src/module/TransactionService.ts` | Invert condition from `=== true` to `!== true` (OCPP 2.0.1); wrap OCPP 1.6 concurrent check with `concurrentTransaction !== true` guard |
| `03_Modules/Transactions/test/module/TransactionService.test.ts` | Update existing tests to assert correct semantics; add new OCPP 1.6 test for `concurrentTransaction=true` |

## How it works

The OCPP spec defines `concurrentTransaction` as: *"Whether the IdToken can be used for multiple concurrent charging sessions."*

- `concurrentTransaction = true` → token **CAN** charge on multiple connectors simultaneously → **do NOT** return `ConcurrentTx`
- `concurrentTransaction = false/undefined` → token **CANNOT** → **return `ConcurrentTx`** if active transactions exist

The fix aligns the code with this spec behavior for both OCPP 1.6 and 2.0.1.

## Test plan

- [x] All 46 existing tests pass (`vitest run` in Transactions module)
- [x] OCPP 2.0.1: `concurrentTransaction=false` + active transactions → `ConcurrentTx`
- [x] OCPP 2.0.1: `concurrentTransaction=true` + active transactions → `Accepted`
- [x] OCPP 1.6: `concurrentTransaction` unset + active transactions → `ConcurrentTx`
- [x] OCPP 1.6: `concurrentTransaction=true` + active transactions → `Accepted`

Fixes citrineos/citrineos#185


thanaParis (approved):
approved!

PR #567, opened 2026-03-02
Fix OCPP 1.6 StatusNotification crash: Connector.evseId cannot be null (#160)

### LINKED ISSUE #160: OCPP 1.6 StatusNotification fails: Connector model requires non-null evseId/evseTypeConnectorId
## **Which repository?**

* `citrineos-core`

---

## **Describe the bug**

In CitrineOS v1.8, OCPP 1.6 `StatusNotification` messages fail with an `InternalError` because the **Connector Sequelize model requires non-null `evseId` and `evseTypeConnectorId`**, even though:

* OCPP 1.6 does not define EVSEs
* the OCPP 1.6 handler does not populate these fields
* the PostgreSQL schema allows these fields to be NULL

This is a regression from v1.7.x, where OCPP 1.6 stations worked normally.

As a result, CitrineOS cannot create connectors for OCPP 1.6 stations, and all subsequent messages (e.g., StartTransaction) fail.

---

## **To Reproduce**

1. Start CitrineOS-Core v1.8 with a clean PostgreSQL database.
2. Connect any OCPP 1.6 charge point (e.g., MicroOcpp Simulator).
3. Observe:

   * `BootNotification` → **Accepted**
   * First `StatusNotification` → **InternalError**
4. Check server logs:

   ```
   SequelizeValidationError: notNull Violation: Connector.evseId cannot be null
   notNull Violation: Connector.evseTypeConnectorId cannot be null
   ```
5. Check the DB:

   ```
   SELECT COUNT(*) FROM "Connectors";  -- returns 0
   ```

---

## **Expected behavior**

* OCPP 1.6 `StatusNotification` should be processed successfully.
* CitrineOS should create/update connectors based solely on `connectorId`, as defined in the OCPP 1.6 protocol.
* No EVSE-related fields should be required for OCPP 1.6 stations.

---

## **Screenshots**

---

## **Desktop:**

* **OS:** macOS ARM64
* **Environment:** Docker
* **CitrineOS version:** `ghcr.io/citrineos/citrineos-server:v1.8.0`

---

### Root Cause

`Connector` model (`Location/Connector.ts`) defines:

```ts
@Column({ allowNull: false })
evseId: number;

@Column({ allowNull: false })
evseTypeConnectorId: number;
```

…but the database schema has:

```
evseId              | integer | YES
evseTypeConnectorId | integer | YES
```

…and the OCPP 1.6 handler constructs a connector without EVSE information, which is correct per the OCPP 1.6 specification.

This mismatch triggers a Sequelize `notNull Violation`.

### Impact

* ❌ OCPP 1.6 chargers cannot create connectors
* ❌ `StatusNotification` always returns `InternalError`
* ❌ `StartTransaction` fails (`Unable to find connector`)
* ✅ OCPP 2.0.1 chargers work (they include EVSE data)

### Tested Fix

Changing both fields in the Sequelize model to `allowNull: true` resolves the issue:

```ts
@Column({ allowNull: true })
evseId?: number;

@Column({ allowNull: true })
evseTypeConnectorId?: number;
```

This allows OCPP 1.6 stations to operate normally again.

---

## **Proposed solution**

Make `evseId` and `evseTypeConnectorId` optional in the Connector model for compatibility with OCPP 1.6, which does not provide EVSE information.

This restores the behavior from v1.7.x while keeping full OCPP 2.0.1 support.

rishabhvaish:
## Summary

- **Bug:** Every OCPP 1.6 `StatusNotification` crashes with `notNull Violation: Connector.evseId cannot be null` because OCPP 1.6 has no EVSE concept — connectors are created without `evseId`.
- **Cascade:** No connectors are created in the database, so `StartTransaction` fails with `Unable to find connector`. OCPP 1.6 chargers cannot transact at all.
- **Root cause:** Both the Zod DTO schema (`connector.dto.ts`) and the Sequelize model (`Connector.ts`) enforce `evseId` as non-nullable, but the database column already allows NULL.

## Changes

| File | Change |
|------|--------|
| `00_Base/src/interfaces/dto/connector.dto.ts` | Make `evseId` `.nullable().optional()` in the Zod schema |
| `01_Data/src/layers/sequelize/model/Location/Connector.ts` | Add `allowNull: true` to the `@Column` decorator; update type to `number \| null` |
| `03_Modules/Transactions/test/module/StatusNotificationService.test.ts` | Add test verifying OCPP 1.6 connectors are created without `evseId` |

## Why this is safe

- The database already allows `NULL` for `evseId` — this was purely an application-layer constraint
- OCPP 2.0.1 chargers are unaffected — they always include `evseId` in `StatusNotification`
- No migration needed

## Test plan

- [x] All 46 existing tests pass (`vitest run` in Transactions module)
- [x] New test: OCPP 1.6 connector created without `evseId` — verifies no `notNull` violation
- [x] Existing OCPP 2.0.1 StatusNotification tests still pass

Fixes citrineos/citrineos#160

### DIFF
--- a/00_Base/src/interfaces/dto/connector.dto.ts
+++ b/00_Base/src/interfaces/dto/connector.dto.ts
@@ -16,7 +16,7 @@ import {
 export const ConnectorSchemaWithoutParent = BaseSchema.extend({
   id: z.number().int().optional(),
   stationId: z.string(),
-  evseId: z.number().int(),
+  evseId: z.number().int().nullable().optional(),
   connectorId: z.number().int(),
   evseTypeConnectorId: z.number().int().optional(),
   status: ConnectorStatusEnumSchema.default('Unknown').nullable().optional(),
--- a/01_Data/src/layers/sequelize/model/Location/Connector.ts
+++ b/01_Data/src/layers/sequelize/model/Location/Connector.ts
@@ -44,8 +44,9 @@ export class Connector extends Model implements ConnectorDto {
   @Column({
     unique: 'evseId_evseTypeConnectorId',
     type: DataType.INTEGER,
+    allowNull: true,
   })
-  declare evseId: number;
+  declare evseId: number | null;
 
   @Column({
     unique: 'stationId_connectorId',
--- a/03_Modules/Transactions/test/module/StatusNotificationService.test.ts
+++ b/03_Modules/Transactions/test/module/StatusNotificationService.test.ts
@@ -173,6 +173,24 @@ describe('StatusNotificationService', () => {
       expect(locationRepository.createOrUpdateConnector).toHaveBeenCalled();
     });
 
+    it('should create connector without evseId for OCPP 1.6 chargers', async () => {
+      locationRepository.readChargingStationByStationId.mockResolvedValue(aChargingStation());
+      vi.spyOn(StatusNotification, 'build').mockImplementation(() => {
+        return aStatusNotification();
+      });
+
+      await statusNotificationService.processOcpp16StatusNotification(
+        DEFAULT_TENANT_ID,
+        MOCK_STATION_ID,
+        aOcpp16StatusNotificationRequest(),
+      );
+
+      const connectorArg = (locationRepository.createOrUpdateConnector as any).mock.calls[0][1];
+      expect(connectorArg.evseId).toBeUndefined();
+      expect(connectorArg.connectorId).toBeDefined();
+      expect(connectorArg.stationId).toBe(MOCK_STATION_ID);
+    });
+
     it('should not save StatusNotification or connector when Charging Station does not exist', async () => {
       componentRepository.readAllByQuery.mockResolvedValue([aComponent()]);
 


thanaParis:
Despite OCPP 1.6 not having the concept of 'EVSEs' in its message structure, the concept still exists for the hardware. I have needed to manage 1.6 DCFC chargers with 2 connectors where the charger can only charge one vehicle at a time (1 evse) and where the charger can charge two vehicles at a time (2 evses). This difference has to be represented at the data level for CitrineOS to function correctly. For example, say there's a 1.6 charging station that has 2 connectors. If a transaction is open on one connector, and a start is received on the other, should Citrine assume that the open transaction has been closed and is now 'stuck' and needs to be marked inactive (a correct assumption if the station has 1 evse) or should it treat this as a new, unrelated transaction and leave the old one be?

This 'bug' is easy to avoid by correctly commissioning chargers in the operator ui. the ability of CitrineOS to accommodate un-commissioned charging stations exists only to reduce friction during ad-hoc testing. At a certain point, adaptions like this start to degrade Citrine's production readiness by letting it handle scenarios where it should break in order to bring the issue to the attention of network admins.

Bottom line, this change would let a potentially serious issue silently persist; I don't think it's a good idea.

PR #569, opened 2026-03-03
Fix SetNetworkProfile API 500 error: parameter order mismatch (#180)

rishabhvaish:
## Summary
- SetNetworkProfile API endpoint returns 500 on every call
- Root cause: parameter order mismatch between `AbstractModuleApi._handler` and `setNetworkProfile`
- `_handler` passes `(identifiers, body, callbackUrl, tenantId, extraQueries)` but `setNetworkProfile` declares `(identifier, request, callbackUrl, extraQueries, tenantId)` — the last two are swapped
- Additionally, `stationId: identifier` passes the full `string[]` array instead of `identifier[0]`

## Changes
- `03_Modules/Configuration/src/module/2.0.1/MessageApi.ts`:
  - Swapped `tenantId` and `extraQueries` parameter order to match `_handler`'s call signature
  - Changed `stationId: identifier` to `stationId: identifier[0]` so a single string is stored, not the array

## Test plan
- [ ] Call SetNetworkProfile API with valid parameters — should succeed instead of 500
- [ ] Verify stationId is correctly stored as a string, not an array
- [ ] Verify other API endpoints still work (no regression)

Fixes #180

### DIFF
--- a/03_Modules/Configuration/src/module/2.0.1/MessageApi.ts
+++ b/03_Modules/Configuration/src/module/2.0.1/MessageApi.ts
@@ -55,15 +55,15 @@ export class ConfigurationOcpp201Api
     identifier: string[],
     request: OCPP2_0_1.SetNetworkProfileRequest,
     callbackUrl?: string,
-    extraQueries?: Record<string, any>,
     tenantId: number = DEFAULT_TENANT_ID,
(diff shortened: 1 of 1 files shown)


thanaParis (approved):
approved

PR #577, opened 2026-03-09
DataTransfer improvements

thanaParis:
- adding data transfer handling for 1.6
- adjusting data transfer handling to manage either request or response.

### DIFF
--- a/00_Base/src/interfaces/modules/AbstractModule.ts
+++ b/00_Base/src/interfaces/modules/AbstractModule.ts
@@ -211,9 +211,7 @@ export abstract class AbstractModule implements IModule {
    */
 
   /**
-   * Unimplemented method to handle incoming {@link IMessage}.
-   *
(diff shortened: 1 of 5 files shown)


lydiazcheng (approved):

mull:
I think the changes done to Server/src/config/envs/{docker.ts,local.ts,swarm.docker.ts} need to be done for 1.6 as well? OCPP1_6_CallAction.DataTransfer

thanaParis:
> I think the changes done to Server/src/config/envs/{docker.ts,local.ts,swarm.docker.ts} need to be done for 1.6 as well? OCPP1_6_CallAction.DataTransfer

As it happens, the DataTransfer call action is the same between the protocol versions ("DataTransfer" in 1.6 and "DataTransfer" in 2.0.1), so there's no need for both in the routing config. The current enum arrangement does cause confusion--this will actually be changing in our next minor version release which has support for 2.1.

PR #581, opened 2026-03-11
Updates to Everest's latest image and sets up server connection for OCPP 2.1 (PR version 2)

devonNewsomDevelop:
Updates Everest image to their current latest
Exposes ports on 8083 for OCPP2.1 unauthorized messaging
Updates checks to allow OCPP2.1 to connect. Further functionality to be added later.
Adjusts start.sh script, docker environment, and package.json to more dynamically work with OCPP versions.
Completely adjusted how we approach changing the port for Everest's config


thanaParis (approved):
approved

PR #582, opened 2026-03-12
Refactor(Overhaul): Restructuring And Remove Circular Dependencies

vicky-ahmed:


elliot-sabitov (changes requested):
@vicky-ahmed it looks like we did not implement the TS `@aliases`. Can you look at how Operator UI handles aliases. We are expecting all imports to use aliases similar to `import { MenuSection } from '@lib/client/components/main-menu/main.menu';`  this is accomplished by adding paths to the tsconfig - ```"paths": {
      "@*": ["./src/*"]
    },``` so the base, core and server modules would each use this. Does that make sense? We can create additional aliases for commonly used folders for example `./src/models/...` or similar can be aliased as `@models/...`, but at very least we should use the `@/...` source alias by default.

elliot-sabitov:
Good news, I ran madge again using `npx madge --circular --extensions ts ./`  and where previously we saw 300+ circular dependencies, we now see only 21. Can we make sure to address those as well? 🙏

PR #586, opened 2026-03-16
improving rabbit mq production readiness.

thanaParis:
https://github.com/citrineos/citrineos/issues/191


02_Util/src/queue/rabbit-mq/receiver.ts
```diff
@@ -84,7 +84,7 @@ export class RabbitMqReceiver extends AbstractMessageHandler {
     // Assert exchange and queue
     await channel.assertExchange(this.exchange, 'headers', { durable: false });
```
jskee256: i guessing the exchange need to be durable also.

If false, the results would be queue survives restart, but the exchange disappears, so bindings are lost and messages won’t route until you re‑declare the exchange and re‑bind.

yuvenus (approved):

PR #588, opened 2026-03-22
fix: skip reservation deactivation on rejected OCPP 1.6 StartTransaction

rishabhvaish:
I noticed that `_handleOcpp16StartTransaction` deactivates the reservation unconditionally after sending the StartTransaction response — even when the transaction was rejected due to auth failure, a missing charging station, or a DB error. This means the reservation gets consumed despite no transaction being created, so the charger can’t retry and no other idTag can use the reservation either.

The fix adds a guard that only calls `deactivateReservation` when `response.idTagInfo.status` is `Accepted`, which means both authorization and transaction creation succeeded. The OCPP 2.0.1 handler doesn’t have this issue because TransactionEvent semantics are different — the CS reports the event and the CSMS just acknowledges it.

Verified the change compiles and matches the repo’s existing code style.


thanaParis (approved):
approved

PR #593, opened 2026-03-26
feat(build): add madge circular dependency check

vicky-ahmed:


package.json
```diff
@@ -6,7 +6,7 @@
   "scripts": {
     "install-all": "npm i --verbose",
     "copy-assets": "npm run copy-assets --prefix ./Server",
-    "build": "tsc --build --verbose",
+    "build": "npm run madge-circular-deps && tsc --build --verbose",
```
elliot-sabitov: lets leave it out of build only keep in lint (pre-commit check) and github pipelines for PRs... because otherwise it will be running on every file change and that is too much and unnecessary

vicky-ahmed: Updated.

thanaParis:
Closing for now because circular dependency issue with sequelize remains unresolved.

PR #594, opened 2026-03-26
Enhancement: Use Column Wildcards Hasura Metadata

yuvenus:
This PR attempts to simplify the Hasura metadata by using wildcard for the column declarations so we don't miss out on new columns.


elliot-sabitov (approved):

thanaParis (approved):
approved

PR #603, opened 2026-03-30
Adding Message API 2.1

devonNewsomDevelop:
Changes how we do abstract APIs for the message api- allows for piping of multiple versions via instantiation
Altered the way we get the schema for message apis to allow for dynamic checks against the instance's version- this way we can connect on 2.1 or 2.0.1 depending on what version the class is instantiated

Creates a grouping for all common types and etc. for the 2.x versions

Creates a method and records to determine which schema version a message api should reflect with

Updates references across the board to new refs

Updates server configuration to use new instantiation


thanaParis (approved):
approved!

PR #604, opened 2026-03-31
Feat/ocpp16 signed firmware status notification

MatthiasContribute:
Integration of signed firmware status notification for ocpp v.1.6 security Whitepaper.


thanaParis (changes requested):
Nothing wrong with the content here.
This PR would be simple to update--apologies for the slow response; the refactor that broke it was open in another PR at the time and I meant to circle back and mention that you should repoint, but lost it in the shuffle.

MatthiasContribute:
Ok, everything done from my site, or shoud i do something?

PR #605, opened 2026-03-31
Fix csr parse error

lydiazcheng:
When there are `\r` in csr, citrine cannot parse it correctly. This PR is to fix this issue. And it also remove the hardcoded default hubject token and change to fetch it instead since the token used for testing need to be refreshed every day.


elliot-sabitov (approved):

lydiazcheng:
HI @elliot-sabitov @thanaParis , I don't have merge permission. Pls feel free to merge it. Thank you!

PR #608, opened 2026-04-01
Feature/set variables now updates device model

elliot-sabitov:
Current:
It was discovered that when SetVariables is sent directly to the charger via RabbitMQ, the device model in the DB never gets updated with the latest value if it is accepted. Instead, in the message API, the record is created when the SetVariables endpoint is hit. 

New:
This adjustment removes the previous logic to create device model variables in the DB in the message API, and instead, when the SetVariables response confirmation from the charger is received, we now get the original SetVariables request using the OCPPMessage table and use that to update / create the variable attribute record in the DB. This way, no matter how the SetVariables request originally entered the system, the SetVariables confirmation will handle updating the values in the DB.


lydiazcheng (approved):
The change looks good. 👍

PR #611, opened 2026-04-02
feat: migrate ChargingStations from id to pkId primary key

vicky-ahmed:


migrations/20260330100000-add-charging-station-pk-id.ts
```diff
@@ -0,0 +1,967 @@
...
+// SPDX-License-Identifier: Apache-2.0
+'use strict';
+
+import { DataTypes, type ModelAttributeColumnOptions, QueryInterface, QueryTypes } from 'sequelize';
+
+export default {
+  up: async (queryInterface: QueryInterface) => {
+    const migrationName = '20260330100000-add-charging-station-pk-id';
+
+    // ── Idempotency guard (read-only, outside the transaction) ─────────────────
```
thanaParis, line 12: why is this here? sequelize already guarantees this out of the box...

01_Data/src/layers/sequelize/model/Certificate/DeleteCertificateAttempt.ts
```diff
@@ -61,6 +64,19 @@ export class DeleteCertificateAttempt extends Model {
   @BelongsTo(() => Tenant)
   declare tenant?: TenantDto;
 
+  @BeforeCreate
```
thanaParis, line 67: These @BeforeCreate hooks everywhere rather than changing the original calls is a bad approach. It adds a query per create and is harder to notice or debug.

01_Data/src/layers/sequelize/model/Location/ChargingStation.ts
```diff
@@ -44,8 +46,21 @@ import { InstalledCertificate } from '../Certificate/index.js';
 export class ChargingStation extends Model implements ChargingStationDto {
   static readonly MODEL_NAME: string = Namespace.ChargingStation;
 
+  @AutoIncrement
   @PrimaryKey
-  @Column(DataType.STRING(36))
+  @Column(DataType.INTEGER)
+  declare pkId: number;
```
thanaParis, line 52: I understand why you took this approach, but it's very disappointing. Keeping the old 'id' column data untouched might be the simplest way to solve this, since it reduces the migration complexity and the need for upstream changes, but it's very confusing to anyone unfamiliar with the code and clearly isn't the way this table should have been designed if we were aware of the constraints from the beginning.

01_Data/src/layers/sequelize/model/Location/ChargingStation.ts
```diff
@@ -44,8 +46,21 @@ import { InstalledCertificate } from '../Certificate/index.js';
...
+
+  /**
+   * The tenant-scoped charging station identifier — used in WebSocket routing
+   * (the charger appends this to the end of the WebSocket URL on connect).
+   * Unique per tenant, but two different tenants may share the same value.
+   */
+  @Index
+  @Column({
+    type: DataType.STRING(36),
+    unique: 'ChargingStations_id_tenantId_key',
```
elliot-sabitov, line 62: nitpick can we please rename `ChargingStations_id_tenantId_key` -> `'id_tenantId'` for consistency 🙏

01_Data/src/layers/sequelize/model/Location/Connector.ts
```diff
@@ -42,18 +48,21 @@ export class Connector extends Model implements ConnectorDto {
...
-    unique: 'stationId_connectorId',
+    unique: 'stationPkId_connectorId',
+    allowNull: false,
     type: DataType.INTEGER,
   })
   declare connectorId: number; // This is the serial int starting at 1 used in OCPP 1.6 to refer to the connector, unique per Charging Station.
 
   @Column({
     unique: 'evseId_evseTypeConnectorId',
+    allowNull: false,
```
elliot-sabitov, line 65: I dont think this is supposed to be nullable since the `evseTypeConnectorId` is `?` optional

thanaParis (approved):
This PR is a tech-debt-heavy shortcut. It looks like it will work, so I'm giving a reluctant approval. These comments will hopefully remind us in the future to properly redesign our tables.

01_Data/src/layers/sequelize/model/TransactionEvent/Transaction.ts
```diff
@@ -47,12 +47,18 @@ export class Transaction extends Model implements TransactionDto {
...
+  })
+  declare stationPkId?: number;
+
   @Column({
     type: DataType.STRING,
-    unique: 'stationId_transactionId',
   })
-  @ForeignKey(() => ChargingStation)
-  declare stationId: string;
+  stationId!: string;
```
elliot-sabitov, line 61: for consistency, please re-add declare, we can consider removing all `declare` statements if we agree them to be unnecessary

01_Data/src/layers/sequelize/model/VariableMonitoring/EventData.ts
```diff
@@ -109,12 +115,28 @@ export class EventData extends Model implements EventDataDto {
...
     onUpdate: 'CASCADE',
     onDelete: 'RESTRICT',
+    unique: 'stationId_tenantId_eventId',
   })
   declare tenantId: number;
 
   @BelongsTo(() => Tenant)
   declare tenant?: TenantDto;
 
+  @BeforeCreate
```
elliot-sabitov, line 125: @vicky-ahmed I think we should add @BeforeUpdate as well right? Because I would assume we would want something similar for records that are being updated, what do you think?

01_Data/src/layers/sequelize/model/Certificate/DeleteCertificateAttempt.ts
elliot-sabitov: I do want to add that it may not be so terrible, it will only run if the stationPkId isnt set, idk why it checks for `null` case only, but likely should be checking to see if it is truthy in case it is undefined because please remember `undefined !== null` also in general if I remember correctly using `!==` and `===` over `!=` and `==` is recommended... but I was thinking it is messy but it does add a nice safety net, and I was proposing we do it for @BeforeUpdate as well, but it only adds the additional `instance.stationPkId == null && instance.stationId && instance.tenantId != null` lookup in most cases, which should not add to the overhead in any significant way

PR #589, opened 2026-03-25
[OCPP21] Update N02

lydiazcheng:
## Changes
This PR,
1. add a OCPP 2.1 handler for `NotifyMonitoringReportRequest`
2. add a new column for the enum `EventNotificationEnumType` and related repository
3. fix a request validation issue

## Tests
1. connect to 2.1, sends a request to citrine and get a response Screenshot 2026-03-24 at 6 03 06 PM
2. the new enum value is stored in db Screenshot 2026-03-24 at 6 04 53 PM

PR #617, opened 2026-04-07
Feature/ocpp2.1 with overhaul

devonNewsomDevelop:
This includes the package overhaul feature as a dependency. The code is already merged into this branch (as well as next)  but it would probably be best to get the overhaul feature into next first so that the original authors can dictate how conflicts are resolved. I will adjust this PR afterwards.

This feature introduces OCPP 2.1 basic message handling and transaction functionality. 

What this feature introduces: 

Module handlers will now accept a list of versions instead of a single version. This allows for a single source of truth for versions that are functionally the same. The memory impact is similar to duplicating the code with the current implementation, but it is more compact. Eventually we may want to consider refactoring how this work if memory becomes a concern. 

Message APIs now work with constructed versions if provided. What this allows for is spinning up multiple of the same message apis but with different versions. Added functionality to fetch the schema for a given message api dynamically based on the constructed version for the decorators. 

Added Everest 2.1 OCPP image as well as adjusting our shell script to tweak everest. We now modify their json file directly in their image directory, allowing us to both modify the url of our CSMS and the protocol that the charger should be created on.

Adds common indices for OCPP2 version types and schemas. 
Adds utility method to package group calls for messages.
Adds forceProtocol property to webserver config so that developers can test a specific protocol when many are offered and accepted by both the CS and CSMS.

The flow of the project is very similar to before. 
Simply change the target version variable for Everest in the "start-everest" command definition within the npm package file and run from there.


lydiazcheng:
Hi Devon, this change is amazing. I manage to build and run the service in docker smoothly. I randomly test few cases and found 2 issues, 
1. changes in this PR is missing, https://github.com/citrineos/citrineos-core/pull/589. Could you pls help add them?
2. the StatusNotification doesn't work. I'm not sure whether it is caused by this change. If not, we might need another ticket for it. Screenshot 2026-04-08 at 5 47 58 PM

thanaParis (approved):
approved!

PR #620, opened 2026-04-09
adding auto-commissioning behavior modulation

thanaParis:


yuvenus (approved):

devonNewsomDevelop:
If it's easily done, it looks like we could perhaps create a common method for the connection get / set. Only a suggestion though, I'm not strongly opinionated one way or the other. Just might make maintaining that little piece of code easier in the future if we ever complicate it more, but if it seems pointless with context then I get that.

PR #621, opened 2026-04-12
feat: add SECP192R1 support for SignedMeterValues to support Alfen ch…

LoveKapibarasan:
## Summary
This PR adds support for the `SECP192R1` signing method to improve interoperability with Alfen (Eichrecht-compliant) chargers.

## Changes
1. **Config/Types**: Added `SECP192R1` to the `signingMethod` Zod enum to allow the server to start when this method is configured.
2. **SignedMeterValuesUtil**: Added a case for `SECP192R1` in the verification logic. For now, it bypasses the cryptographic check (returns true) to allow meter values from Alfen chargers to be accepted without SecurityErrors.

## Background & Necessity
This is a critical workaround for environments using Alfen chargers. Without this change, our Alfen fleet is completely unusable with CitrineOS:
- If we configure the signing method correctly, the server fails to boot due to **Zod Schema Error**.
- If we use existing options, it results in **"unsupported" security errors** during transaction processing.

To prevent our Alfen chargers from being dead in a CitrineOS environment, this support for elliptic curve cryptography (ECC) strings and the corresponding logic bypass is essential.

### DIFF
--- a/00_Base/src/config/types.ts
+++ b/00_Base/src/config/types.ts
@@ -158,7 +158,7 @@ export const systemConfigInputSchema = z.object({
       signedMeterValuesConfiguration: z
         .object({
           publicKeyFileId: z.string(),
-          signingMethod: z.enum(['RSASSA-PKCS1-v1_5', 'ECDSA', 'SECP192R1']),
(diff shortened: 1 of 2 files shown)


thanaParis:
closed in favor of https://github.com/citrineos/citrineos-core/pull/654

PR #626, opened 2026-04-13
feat: add configuration option to enable/disable GetChargingProfiles …

Adex9ja:
This pull request introduces a new configuration option to control whether the EVDriver module should request charging profiles after a start transaction. The change is implemented in the configuration schemas, module logic, and all relevant environment configuration files. By default, this feature is enabled, but it can be explicitly disabled through configuration.

**Configuration changes:**

* Added a new boolean configuration option, `enableGetChargingProfilesOnStartTransaction`, to the `evdriver` module in both the input and runtime config schemas, with a default value of `true`. 

**Module logic updates:**

* Updated the `EVDriverModule` to only request charging profiles after a start transaction if `enableGetChargingProfilesOnStartTransaction` is not set to `false`. If disabled, a log message is generated to indicate the request is being skipped. 

**Environment configuration updates:**

* Set `enableGetChargingProfilesOnStartTransaction: true` by default in the local, docker, and swarm docker environment config files (`local.ts`, `docker.ts`, `swarm.docker.ts`). 
* Updated the system config test provider to include the new option with a value of `true`.…on StartTransaction

### DIFF
--- a/Server/src/config/envs/docker.ts
+++ b/Server/src/config/envs/docker.ts
@@ -80,6 +80,7 @@ export function createDockerConfig() {
       },
       evdriver: {
         endpointPrefix: '/evdriver',
+        enableGetChargingProfilesOnStartTransaction: true,
         responses: [
           OCPP_CallAction.CancelReservation,
           OCPP_CallAction.ClearCache,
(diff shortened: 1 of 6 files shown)


thanaParis (approved):
approved!

PR #632, opened 2026-04-14
fix: ensuring to use idToken.type for authorization query string

yuvenus:
Minor fix to authorization query string to ensure idToken.type is passed in properly


yuvenus:
already represented in another PR, closing!

PR #634, opened 2026-04-15
feat: add prepaid fields to Authorizations and implement prepaid card authorization logic

Adex9ja:
This pull request introduces support for prepaid card authorization within the authorization and transaction flow, including full backend, DTO, and database support, as well as comprehensive tests. The main changes include the addition of `isPrepaid` and `prepaidBalance` fields to the authorization model, logic to handle prepaid balances during authorization, and OCPP 2.1-specific behavior for transaction limits.

**Prepaid Card Authorization Support**

* Added new fields `isPrepaid` (boolean) and `prepaidBalance` (number or null) to the `Authorization` DTO (`base/src/interfaces/dto/authorization.dto.ts`) and Sequelize model (`core/src/dal/layers/sequelize/model/Authorization/Authorization.ts`) to track prepaid status and balance. 
* Database migration introduced to add the `isPrepaid` and `prepaidBalance` columns to the `Authorizations` table.

**Authorization and Transaction Flow Logic**

* Implemented logic in `EVDriverModule` and `TransactionService` to check for prepaid status:
  - If a token is prepaid and has a positive balance, authorization is accepted and `cacheExpiryDateTime` is set to now.
  - If the balance is zero or negative, the status is set to `NoCredit` and `cacheExpiryDateTime` is set to now.
  - For OCPP 2.1, if the token is prepaid and accepted, the response includes a `transactionLimit.maxCost` equal to the prepaid balance. 

**Testing**

* Added comprehensive tests in `PrepaidAuthorization.test.ts` to verify:
  - Correct handling of prepaid tokens with positive, zero, negative, or null balances.
  - OCPP 2.1-specific behavior for `transactionLimit.maxCost`.
  - Ensured non-prepaid tokens are unaffected by the new logic.


core/src/modules/EVDriver/src/module/module.ts
```diff
@@ -464,6 +464,30 @@ export class EVDriverModule extends AbstractModule {
           response.idTokenInfo.status =
             OCPP2_0_1_Mapper.AuthorizationMapper.fromAuthorizationStatusEnumType(result);
         }
+
```
lydiazcheng, line 594: The authorize handler for OCPP 2.1 has some new logics than OCPP 2.0.1. How about adding a new OCPP 2.1 handler for now? If we would like to reuse duplicated codes, then we can merge them later after we implemented all new 2.1 cases related to `authorize` so that we can make a better decision.

core/src/modules/Transactions/src/module/TransactionService.ts
lydiazcheng, line 89: It would be great to update the name if we would like to reuse it for 2.1 as well.

There are also 2.0.1 object used in this method, like line 104, 109. It would be better to use some common object for 2.X. Or create a new method just for 2.1.

core/src/modules/Transactions/src/module/TransactionService.ts
```diff
@@ -165,6 +166,104 @@ export class TransactionService {
...
 
+  async authorizeOcpp21IdToken(
+    tenantId: number,
+    transactionEvent: OCPP2_request_types.TransactionEventRequest,
+    messageContext: IMessageContext,
+  ): Promise<OCPP2_1.TransactionEventResponse> {
+    const idToken = transactionEvent.idToken!;
+    const authorizations = await this._authorizeRepository.readAllByQuerystring(tenantId, {
+      idToken: idToken.idToken,
+      type: idToken as unknown as IdTokenEnumType,
```
thanaParis: This line is a bug copied from the other authorize handler, fixed in recent pr. Please update to idToken.type.

thanaParis (changes requested):

core/src/modules/EVDriver/src/module/module.ts
lydiazcheng: Hi @Adex9ja , thanks for the update. In another PR, https://github.com/citrineos/citrineos-core/pull/638/changes#diff-5166a0945362d2c781b8a0f29e5f52e120910052c96e3c10152148a782269b77, I also added a new 2.1 authorize handler and also update this handler using 2.1 specific methods like validateOcpp21IdToken and 2.1 object mappers. To avoid merge conflicts, how about we merge that PR at first and then add C17 related logic into that 2.1 handler? 🙏  CC: @thanaParis

thanaParis (approved):

PR #642, opened 2026-04-20
ensuring tls certificates can be updated without restarts

thanaParis:


devonNewsomDevelop (approved):
The only thing I can see is that the constructor / method cb(null, ctx) suggests to me that the first param might be better as an optional param but that's nitpicky

Adex9ja (approved):

thanaParis:
> The only thing I can see is that the constructor / method cb(null, ctx) suggests to me that the first param might be better as an optional param but that's nitpicky

Not my method! Part of node's https implementation, so I can't change it.

PR #638, opened 2026-04-17
C18: Authorization using locally connected payment terminal

lydiazcheng:
## Context
This PR is to implement handlers and request based on **OCPP 2.1, C18**.

## Changes
1. add new `_handleOcpp21Authorize` handler. 
   1. Take most logic from OCPP 2.0.1 but add new implementation for `TariffType` and 2.1 token validation, `validateOcpp21IdToken`.
   2. Extend `Tariff` table with 2.1 fields. Add a foreign key `tariffId` in `Authorization` table to find non-default tariff for customer
   3. update the `upsert tariff` endpoint for new fields
2. add new `_handelVatNumberValidation` handler
   1. implement a free EU VIES REST API provided by European Commission for VAT validation
3. add new `SetDefaultTariff` endpoint

## Tests
1. send `SetDefaultTariff` request to charger Screenshot 2026-04-16 at 3 36 09 PM
2. create a new tariff Screenshot 2026-04-16 at 12 08 48 PM
3. manually create a `DirectPayment` token and associate it with the new created tariff 
4. set tariff enable to true Screenshot 2026-04-17 at 1 43 20 PM
5. charger sends a `DirectPayment` authorize request to citrine and it accepts the request and return the tariff as well Screenshot 2026-04-17 at 11 33 24 AM
6. charger sends a `VatNumberValidation` request to citrine and get accept response with expected company info. Screenshot 2026-04-17 at 1 46 10 PM

PR #644, opened 2026-04-21
I07: Set Default Tariff

lydiazcheng:
## Context
This PR is to implement OCPP2.1, I07 Set Default Tariff
## Changes
1. Store default tariff in the db
2. Add validation for times in TariffType
3. Map tariffId (string) in TransactionEventRequest with tariff.id (db id) and store in db
## Tests
1. Send SetDefaultTariff request to charger Screenshot 2026-04-21 at 11 46 28 AM
2. The tariff is stored in db Screenshot 2026-04-21 at 12 24 54 PM
3. Citrine validate the times in the request and returns error messages Screenshot 2026-04-21 at 11 14 45 AM
4. When receiving the TransactionEventRequest from charger, the tariffId is mapped to tariff db id Screenshot 2026-04-21 at 12 26 29 PM

### DIFF
--- a/core/src/dal/layers/sequelize/repository/TransactionEvent.ts
+++ b/core/src/dal/layers/sequelize/repository/TransactionEvent.ts
@@ -19,8 +19,7 @@ import type {
   IChargingStationSequenceRepository,
   ITransactionEventRepository,
 } from '../../../interfaces/repositories.js';
-import { AuthorizationMapper } from '../mapper/2.1/AuthorizationMapper.js';
-import { MeterValueMapper } from '../mapper/2.0.1/MeterValueMapper.js';
+import { MeterValueMapper } from '../mapper/2/MeterValueMapper.js';
 import { Authorization } from '../model/Authorization/Authorization.js';
 import { EvseType } from '../model/DeviceModel/EvseType.js';
 import { ChargingStation } from '../model/Location/ChargingStation.js';
@@ -192,9 +191,7 @@ export class SequelizeTransactionEventRepository
             where: {
               tenantId,
               idToken: value.idToken.idToken,
-              idTokenType: AuthorizationMapper.fromIdTokenEnumType(
-                value.idToken.type as OCPP2_1.IdTokenEnumType,
-              ), // use 2.1 to cover both OCPP versions
+              idTokenType: value.idToken.type,
             },
             transaction: sequelizeTransaction,
           });
@@ -274,9 +271,7 @@ export class SequelizeTransactionEventRepository
             where: {
               tenantId,
               idToken: value.idToken.idToken,
-              idTokenType: AuthorizationMapper.fromIdTokenEnumType(
-                value.idToken.type as OCPP2_1.IdTokenEnumType,
-              ),
+              idTokenType: value.idToken.type,
             },
             transaction: sequelizeTransaction,
           });
@@ -322,9 +317,7 @@ export class SequelizeTransactionEventRepository
           where: {
             tenantId,
             idToken: value.idToken.idToken,
-            idTokenType: AuthorizationMapper.fromIdTokenEnumType(
-              value.idToken.type as OCPP2_1.IdTokenEnumType,
-            ),
+            idTokenType: value.idToken.type,
           },
           transaction: sequelizeTransaction,
         });
@@ -342,8 +335,7 @@ export class SequelizeTransactionEventRepository
 
       if (value.meterValue && value.meterValue.length > 0) {
         const meterValueTypes = value.meterValue.map((meterValue) =>
-          // meterValueType is effectively the same between OCPP 2.1 and 2.0.1
-          MeterValueMapper.fromMeterValueType(meterValue as OCPP2_0_1.MeterValueType),
+          MeterValueMapper.fromMeterValueType(meterValue),
         );
         const newMeterValues = await Promise.all(
           meterValueTypes.map(async (meterValueType) => {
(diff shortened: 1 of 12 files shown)


lydiazcheng:
⚠️ This branch is now branch off of another feature branch (https://github.com/citrineos/citrineos-core/pull/638). Please update the destination branch before merging.

core/src/dal/layers/sequelize/repository/TransactionEvent.ts
```diff
@@ -171,14 +175,26 @@ export class SequelizeTransactionEventRepository
...
         let authorizationId = existingTransaction.authorizationId;
         if (!authorizationId && value.idToken) {
           // Find Authorization by IdToken
           const authorization = await Authorization.findOne({
             where: {
               tenantId,
               idToken: value.idToken.idToken,
-              idTokenType: AuthorizationMapper.fromIdTokenEnumType(value.idToken.type),
+              idTokenType: AuthorizationMapper.fromIdTokenEnumType(
+                value.idToken.type as OCPP2_1.IdTokenEnumType,
```
devonNewsomDevelop: Instead of using OCPP2_1.IdTokenEnumType, I believe we have a common enum defined in enums.ts for this instead.

core/src/dal/layers/sequelize/repository/TransactionEvent.ts
```diff
@@ -249,7 +274,9 @@ export class SequelizeTransactionEventRepository
...
```
devonNewsomDevelop: Same enum thing here

core/src/dal/layers/sequelize/repository/TransactionEvent.ts
```diff
@@ -295,7 +322,9 @@ export class SequelizeTransactionEventRepository
           where: {
             tenantId,
             idToken: value.idToken.idToken,
-            idTokenType: AuthorizationMapper.fromIdTokenEnumType(value.idToken.type),
+            idTokenType: AuthorizationMapper.fromIdTokenEnumType(
+              value.idToken.type as OCPP2_1.IdTokenEnumType,
```
devonNewsomDevelop: Same enum here

core/src/dal/layers/sequelize/repository/TransactionEvent.ts
```diff
@@ -313,7 +342,8 @@ export class SequelizeTransactionEventRepository
 
       if (value.meterValue && value.meterValue.length > 0) {
         const meterValueTypes = value.meterValue.map((meterValue) =>
-          MeterValueMapper.fromMeterValueType(meterValue),
+          // meterValueType is effectively the same between OCPP 2.1 and 2.0.1
+          MeterValueMapper.fromMeterValueType(meterValue as OCPP2_0_1.MeterValueType),
```
devonNewsomDevelop: Can we add this to the common types under OCPP2_common_types like the other types of this nature? Thank you very much

core/src/modules/Transactions/src/module/module.ts
```diff
@@ -818,6 +820,55 @@ export class TransactionsModule extends AbstractModule {
     props?: HandlerProperties,
   ): Promise<void> {
     this._logger.debug('OCPP 2.1 SetDefaultTariff response received:', message, props);
+
+    if (message.payload.status !== OCPP2_1.TariffSetStatusEnumType.Accepted) {
```
devonNewsomDevelop: Probably good to add this enum to enums.ts if it's not there already! If it is, please change to using that, Thanks so much

core/src/util/util/validator.ts
```diff
@@ -717,6 +717,46 @@ export function validateMessageContentType(
...
+ */
+export function validateTimeOfDay(time: string): boolean {
+  return /^([0-1][0-9]|2[0-3]):[0-5][0-9]$/.test(time);
+}
+
+/**
+ * Validate all startTimeOfDay and endTimeOfDay fields in a TariffType per I07.FR.17.
+ * Checks every TariffConditionsType and TariffConditionsFixedType across all price arrays.
+ */
+export function validateTariffConditionsTimeFields(tariff: OCPP2_1.TariffType): ValidationResult {
```
devonNewsomDevelop, line 731: Instead of OCPP2_1.TarrifType, is this better as a common type? Or is this one specific only to OCPP2_1? If it's specific to 2_1 and doesn't exist in 2_0_1, then I agree with this typing. Otherwise, we should make a union type under common types!

devonNewsomDevelop (changes requested):
Approved with suggestions! Minor type changes. Everything else looks good

core/src/dal/layers/sequelize/repository/TransactionEvent.ts
lydiazcheng: Thanks for the review, Devon. I create a new mapper for the common types and use it instead.

core/src/util/util/validator.ts
lydiazcheng: TariffType is a new object introduced by OCPP 2.1, so this validation is also only for 2.1.

devonNewsomDevelop (approved):
Wow, you went above and beyond! For the mapper I had planned to do that some time, if one of my comments had sent you down this rabbit hole Im sorry I just meant for easy type changes. Thanks for going ahead out of your way and doing all that

lydiazcheng:
> Wow, you went above and beyond! For the mapper I had planned to do that some time, if one of my comments had sent you down this rabbit hole Im sorry I just meant for easy type changes. Thanks for going ahead out of your way and doing all that

Thank you, Devon. Yes, it would be great we have a ticket and update these mappers all together.

thanaParis (approved):

PR #651, opened 2026-04-22
C-22: enhance NotifySettlement handling to support rejected and faile…

Adex9ja:
This pull request enhances the handling of settlement notifications in the `TransactionsModule` to support new settlement statuses and ensure correct storage and response logic. The main focus is on properly differentiating between successful, rejected, and failed settlements, especially regarding receipt information.

**Settlement status handling improvements:**

* Added explicit handling for new settlement statuses: `Rejected` and `Failed` in addition to the existing `Settled` and `Cancelled` statuses, updating documentation and logic to reflect these cases.

* For settlements with status `Rejected` or `Failed`, the system logs a warning and stores settlement data without receipt information (`receiptId`, `receiptUrl`, `vatNumber`), ensuring only successful settlements include receipts.

**Receipt generation logic:**

* Modified the logic so that the receipt is only generated and included in the response for successful (`Settled`) settlements, and not for `Rejected` or `Failed` statuses.


thanaParis (approved):

thanaParis (changes requested):
merge conflicts

core/src/modules/Transactions/src/module/module.ts
```diff
@@ -618,22 +621,41 @@ export class TransactionsModule extends AbstractModule {
...
+          `amount=${request.settlementAmount}. ` +
+          `statusInfo=${request.statusInfo ?? 'none'}. ` +
+          `CPO may need to manually capture the amount via PSP using the pspRef.`,
+      );
+    }
+
     // Store settlement data on the transaction if a transactionId is provided
     if (request.transactionId) {
       try {
+        const settlementData: Record<string, unknown> = {
```
devonNewsomDevelop, line 651: Is it possible to type the unknown as `Partial<Transaction>? `

core/src/modules/Transactions/src/module/module.ts
```diff
@@ -618,22 +621,41 @@ export class TransactionsModule extends AbstractModule {
         `amount=${request.settlementAmount}, transactionId=${request.transactionId ?? 'none'}`,
     );
 
+    const isSettled = request.status === OCPP2_1.PaymentStatusEnumType.Settled;
+    const isRejected = request.status === OCPP2_1.PaymentStatusEnumType.Rejected;
+    const isFailed = request.status === OCPP2_1.PaymentStatusEnumType.Failed;
+
```
devonNewsomDevelop, line 637: We should probably throw an error if the response isn't one of isSettled, isRejected, or isFailed. Defensive coding against charging station messages in case they mess up the enum

devonNewsomDevelop (changes requested):
Some minor changes requested and then we're good to go.

core/src/modules/Transactions/src/module/module.ts
Adex9ja: @devonNewsomDevelop, It's stored in customData field of Transaction entity. Defined as below:

```
  @Column(DataType.JSONB)
  declare customData?: any | null;
```

devonNewsomDevelop: I was going off of the repository method you're using this in, it appears that it expects the type `Partial<Transaction>` right?

Edit: Forgot to tag, sorry. @Adex9ja

Adex9ja: Yes, `updateTransactionByStationIdAndTransactionId` expects `Partial<Transaction>`

core/src/modules/Transactions/src/module/module.ts
devonNewsomDevelop: I'm sorry @Adex9ja if I wasn't clear. I didn't mean to change this to an error, I thought it was fine as a warning.

What I meant was to add something like

``` 
if (!isRejected && !isSettled && !isFailed){
    throw an error saying an unknown state occurred or something like that
}
```

core/src/modules/Transactions/src/module/module.ts
devonNewsomDevelop: Yeah, so my thought was we would make it explicit here what type it was. But if this would cause issues somewhere else in the code, I'm okay with leaving it as is. If you think this typing could cause a problem or cause confusion, you can resolve this comment without changing it

devonNewsomDevelop (approved):
Looks good.

thanaParis (approved):

PR #658, opened 2026-04-23
C-23: increase authorization amount on transaction limit 

Adex9ja:
This pull request adds logic to handle increased authorization amounts during an active transaction for OCPP 2.1 protocol. When a `TransactionEventRequest` with a `triggerReason` of `LimitSet` is received, the system updates the transaction's `maxCost` in the `customData` field and logs the change.

**Enhancements for OCPP 2.1 transaction updates:**

* Added handling for the `LimitSet` trigger reason in OCPP 2.1 `TransactionEventRequest`, updating the transaction's `maxCost` in `customData` when the authorization amount increases. Includes logging and error handling for the update operation.


core/src/modules/Transactions/src/module/module.ts
```diff
@@ -370,6 +370,47 @@ export class TransactionsModule extends AbstractModule {
           );
         }
 
+        // C23: Increasing authorization amount
+        // When CS sends TransactionEventRequest(Updated) with triggerReason=LimitSet (OCPP 2.1),
+        // it indicates the authorization amount has been increased and transactionLimit.maxCost updated.
+        if (message.protocol === OCPPVersion.OCPP2_1 && transaction && transaction.isActive) {
+          const ocpp21Payload = message.payload as unknown as OCPP2_1.TransactionEventRequest;
+          if (
```
devonNewsomDevelop, line 381: It looks like this IF statement could be added to the first check? Thoughts?

core/src/modules/Transactions/src/module/module.ts
```diff
@@ -370,6 +370,47 @@ export class TransactionsModule extends AbstractModule {
...
+                tenantId,
+                {
+                  customData: {
+                    ...existingCustomData,
+                    transactionLimit: {
+                      ...(existingCustomData.transactionLimit ?? {}),
+                      maxCost: newMaxCost,
+                    },
+                  },
+                } as any,
```
devonNewsomDevelop: Are we able to do `as Partial<Transaction>` instead?

core/src/modules/Transactions/src/module/module.ts
devonNewsomDevelop: I understand it's probably structured this way since you're typing the payload. If it's a pain to do, I think it's fine to leave it as is!

devonNewsomDevelop (approved):
this is a test approval. ignore

devonNewsomDevelop (changes requested):
Ok this is the real review. We had to test something with merging, sorry about that. 

Two minor change requests, one of which you probably cannot change (the if statement location). Resolve it if that change is too much trouble @Adex9ja

devonNewsomDevelop (approved):

PR #659, opened 2026-04-26
chore(ocpp-router): remove empty IAdminApi placeholder

camiloh95:
- `IAdminApi` was empty and had a single implementor; `implements` enforced no contract.
- Deleted the interface file, removed the `implements` clause and the public type export.
- Build and OcppRouter tests pass (81/81).


thanaParis (changes requested):
needs to be updated, repository structure is outdated.

thanaParis (approved):

PR #660, opened 2026-04-26
fix(transactions): set Connector FK fields on OCPP 1.6 StatusNotification (#160)

camiloh95:
- Stamps `evseId` and `evseTypeConnectorId` on the Connector record from the matching evse/connector for commissioned stations.
- Auto-commissions a default evse (1 connector → 1 evse for OCPP 1.6) when `allowUnknownChargingStations: true` and no matching evse exists.
- Strict mode preserved: throws when commissioning is required but the flag is off — matches PR #567 maintainer feedback (no nullable FKs).
- Tests: 4 new cases (FK stamping, auto-commission, strict-mode throw, audit-trail behavior); 2 existing tests refactored.

Fixes citrineos/citrineos#160

### DIFF
--- a/core/src/modules/Transactions/src/module/StatusNotificationService.ts
+++ b/core/src/modules/Transactions/src/module/StatusNotificationService.ts
@@ -169,24 +169,15 @@ export class StatusNotificationService {
           (connector) => connector.connectorId === statusNotificationRequest.connectorId,
         ),
       );
-      const statusNotificationInput: Partial<StatusNotification> = {
-        tenantId,
-        ...statusNotificationRequest,
-        ocppConnectionName: ocppConnectionName,
-        connectorStatus: statusNotificationRequest.status,
-      };
-      if (matchingEvse) {
-        statusNotificationInput.evseId = matchingEvse.evseTypeId;
-      }
-      const statusNotification = StatusNotification.build(statusNotificationInput);
-      await this._locationRepository.addStatusNotificationToChargingStation(
-        tenantId,
-        ocppConnectionName,
-        statusNotification,
+      const matchingConnector = matchingEvse?.connectors?.find(
+        (connector) => connector.connectorId === statusNotificationRequest.connectorId,
       );
 
+      // We upsert the Connector BEFORE saving the StatusNotification because
+      // StatusNotifications.connectorId has an FK to Connectors.connectorId.
       const connector = {
         tenantId,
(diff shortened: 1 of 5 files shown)


core/src/modules/Transactions/src/module/StatusNotificationService.ts
```diff
@@ -169,6 +169,87 @@ export class StatusNotificationService {
...
+      // Connector model declares evseId and evseTypeConnectorId as allowNull:false.
+      // For commissioned stations these come from the matching evse/connector;
+      // for ad-hoc 1.6 stations we auto-commission below (citrineos/citrineos#160).
+      // We upsert the Connector BEFORE saving the StatusNotification because
+      // StatusNotifications.connectorId has an FK to Connectors.connectorId.
+      let connectorEvseId: number | undefined = matchingEvse?.id;
+      let connectorEvseTypeConnectorId: number | undefined = matchingConnector?.evseTypeConnectorId;
+
+      if (chargingStation.use16StatusNotification0 && statusNotificationRequest.connectorId === 0) {
+        const broadcastConnector = {
```
thanaParis: I think there's a misunderstanding here--there's no such thing as a "broadcast connector" and putting a connector with id=0 in the db would be a bug. The previous behavior (update all connector statuses when use16StatusNotification0 & connectorId == 0) is correct.

thanaParis (changes requested):
Looks good, I'm more comfortable with this than the other auto-commissioning approach. Auto-commissioning is very dangerous, will lead to unexpected behavior (even this PR, which will lead to very strange OCPI mappings and flawed transaction lifecycles), and should never be done in production ever. However, given that it exists as a feature, this is a definitive improvement, thank you!

One change required, since there was an unrelated behavior change to connectorId=0 handling--that needs to be undone.

core/src/modules/Transactions/src/module/StatusNotificationService.ts
camiloh95: Thanks for the review. Restored the original broadcast-update behavior `use16StatusNotification0` + `connectorId == 0` now goes through `updateAllConnectorsByQuery` with `{ ...connector, connectorId: undefined }` like before, so no `connectorId=0` row gets inserted. The FK stamping is isolated to the specific-connector path.

thanaParis (approved):

PR #663, opened 2026-04-28
Fix/rabbitmq receiver reconnect

elliot-sabitov:
- fix: issue with CitrineOS not re-creating MODULE_MODE queues when reconnecting to RabbitMQ. Adjusted implementeation to keep track of _moduleSubscriptions so that _onReconnect we can _reinitializeModuleQueues and _reinitializeInstanceQueue if in router mode. This ensures that when Connection is somehow terminated between CitrineOS and RabbitMQ, CitrineOS will properly re-establish the needed queues.
- chore: cleaned up implementation in efforts to simplify readability, removed `_pendingInstanceQueueName` in favor of just keeping _instanceQueueName and introducing _instanceQueueReady and _isRouterMode as well as other adjustments to help readability. Ensuring that `this._channelManager.getConnectionManager().on('connected', ...` subscription only exists in one place to reduce the cognitive load as well.
- chore: adjusted all queues to autoDelete true for now since we are not currently implementing logic to persist in flight messages during disconnects, which can be revisited in the future.
- tests: updating tests to account for latest changes
- chore: adjusted RabbitMqReceiver properties to protected access for easier extendability

### DIFF
--- a/core/src/util/queue/rabbit-mq/receiver.ts
+++ b/core/src/util/queue/rabbit-mq/receiver.ts
@@ -30,22 +30,22 @@ import { RabbitMQChannelManager } from './ChannelManager.js';
  * environment variable wired into `SystemConfig.util.messageBroker.amqp.instanceIdentifier`.
  */
 export class RabbitMqReceiver extends AbstractMessageHandler {
-  private static readonly QUEUE_PREFIX = 'rabbit_queue_';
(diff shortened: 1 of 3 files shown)


lydiazcheng (approved):
Changes look good. Only some tests fail.

thanaParis (approved):

PR #665, opened 2026-04-29
C-25: add TOTP utility for QR-code web payment validation and impleme…

Adex9ja:
…nt web payment initiation endpoint


core/src/modules/EVDriver/src/module/2/MessageApi.ts
```diff
@@ -411,6 +417,178 @@ export class EVDriverOcpp2Api
...
+  /**
+   * Registers a custom REST endpoint for initiating a C25 QR-code web payment session.
+   *
+   * POST /evdriver/webpayment/initiate
+   *
+   * This endpoint is called by the CSMS web page after the EV driver scans a QR code.
+   * It validates the TOTP embedded in the QR URL (C25.FR.07-09), caches the driver's
+   * optional transaction limits (C25.FR.03-06), and sends NotifyWebPaymentStartedRequest
+   * to the charging station to lock the EVSE (C25.FR.21-22).
+   *
```
devonNewsomDevelop, line 456: You don't need to add the typing to the actual code, but do you mind to add the type /interface for the body in a comment? Do we have an interface for this?

devonNewsomDevelop: You added all the parameters I know. I meant the overall type / interface to be clear. Specifically the name of the interface / type.

core/src/modules/EVDriver/src/module/2/MessageApi.ts
```diff
@@ -411,6 +417,178 @@ export class EVDriverOcpp2Api
...
+              'QR code may be expired or fraudulent.',
+          );
+          return reply
+            .code(401)
+            .send({ error: 'TOTP validation failed. The QR code may be expired.' });
+        }
+
+        // Cache the QR transaction limits keyed by stationId:evseId (C25.FR.03-06, C25.FR.56-58)
+        // These are read by the TransactionEvent handler to set transactionLimit in the response.
+        const cacheKey = `webpayment:${identifier}:${evseId}`;
```
devonNewsomDevelop: What will happen in a multi-tenant set up?

If we have a tenant group of 1 and a tenant group of 2, and they happen to use the same identifier and evseId names- wouldn't those collide?

core/src/modules/Transactions/src/module/module.ts
```diff
@@ -338,30 +339,71 @@ export class TransactionsModule extends AbstractModule {
...
-              component_name: 'PaymentCtrlr',
-              variable_name: 'AuthorizationAmount',
-              type: AttributeEnum.Actual,
-            });
+        const ocpp21Response = response as OCPP2_1.TransactionEventResponse;
+
+        // C25.FR.03-06: Check cache for QR payment limits set by initiateWebPayment endpoint
+        if (!ocpp21Response.transactionLimit && transactionEvent.evse?.id != null) {
+          try {
+            const cacheKey = `webpayment:${stationId}:${transactionEvent.evse.id}`;
```
devonNewsomDevelop: Same question here regarding multi-tenancy and unique cache keys

devonNewsomDevelop (changes requested):
Some minor changes to work on for now or review. I will need to give this another pass with a fresh set of eyes to make sure I didn't miss anything, but overall looking good and thorough. Nice unit tests.

devonNewsomDevelop (approved):
Good job

PR #666, opened 2026-04-29
improving health check

thanaParis:


Server/src/citrineOSServer.ts
```diff
@@ -294,7 +298,44 @@ export class CitrineOSServer {
...
+        return respond(reply, {
+          status: 'fail',
+          checks: { init: { status: 'fail', error: 'not yet initialized' } },
+        });
+      }
+      return respond(reply, await this._healthCheckService.checkReadiness());
+    };
+
+    this._server.get('/health', liveness);
+    this._server.get('/health/live', liveness);
```
lydiazcheng, line 326: Just be curious, why do we have 2  endpoints for liveness?

lydiazcheng (approved):

Server/src/citrineOSServer.ts
thanaParis: /health/live is correct, but /health is the default for many configurations
If we had to have one, /health/live would be the pick, but supporting both is free and means that custom route mapping for health checks will be significantly less frequent (though, if you're only hitting /health you can only establish that the service responds, not that it's ready--but that's also by design, and provides a way to have the same lightweight health check that existed before this PR)

PR #668, opened 2026-04-29
Feature/drizzle init

elliot-sabitov:
- chore: removed unused base crud repository methods
- feat: added drizzle dependencies, created initial drizzle implementation, introducing base and security event drizzle repositories, creating `DefaultDrizzleInstance` singleton equivalent of `DefaultSequelizeInstance`, and adjusted SecurityEvent handler to use drizzle based on CITRINEOS_USE_DRIZZLE_SECURITY_EVENT env var.
- feat: created e2e test to run rabbitmq, postgresdb, citrine, pre-seed db, establish ws connection to citrine, send security event notification, ensure that the DB has an appropriate security event record created as expected based on handler implementation. The test does this for both Sequelize and Drizzle while toggling `CITRINEOS_USE_DRIZZLE_SECURITY_EVENT` env var to ensure that the handler works the same for both implementations.
- feat: adding drizzle-zod to create entity dtos and mappers to map to existing, potentially external facing DTOs.
- feat: created Explicit type and ensuring that mapping functions utilize explicit to ensure that mapping is enforced to prevent unintended issues where typescript fails to raise build issues resulting in run time bugs


core/src/dal/layers/drizzle/repository/SecurityEvent.ts
```diff
@@ -0,0 +1,107 @@
...
+  tenantSecurityEventTable,
+} from '../schema/SecurityEvent.js';
+import { DrizzleRepository } from './Base.js';
+import type { ISecurityEventRepository } from '@/dal/index.js';
+
+// ─── Mapper ──────────────────────────────────────────────────────────────────
+// Maps a Drizzle entity (DB row, validated by SecurityEventEntitySchema) to the
+// external SecurityEventDto contract. This keeps the ORM type contained to the
+// DAL layer while letting the rest of the system work against stable DTOs.
+export function toSecurityEventDto(entity: SecurityEventEntity): SecurityEventDto {
```
thanaParis, line 28: I think this could be safer

```
// Forces every key of T to be explicitly present in the object literal,
// even optional ones — the value can still be undefined, but it must be declared.
type Explicit<T> = { [K in keyof Required<T>]: T[K] };

export function toSecurityEventDto(entity: SecurityEventEntity, logger?: Logger<ILogObj>): SecurityEventDto | null {
  const dto: Explicit<SecurityEventDto> = {
    id: entity.id,
    stationId: entity.stationId,
    // ...forgetting `updatedAt` here now causes a compile error
  };
  const result = SecurityEventDtoSchema.safeParse(dto);  
  if (!result.success) {
    logger?.error('unable to map row to dto:', result.error);
    return null;  
  } else {  
    return result.data;
  }
}
```
something like the above?

thanaParis (changes requested):
I really like this overall. SecurityEvent is an okay table to start with--I think we want to remove it entirely perhaps even next release, but that's perfect for testing...
One change suggested

core/src/dal/layers/drizzle/repository/SecurityEvent.ts
elliot-sabitov: adjusted 👍 🙇

thanaParis (approved):

PR #701, opened 2026-05-12
Bugfix/rename backend tweak

devonNewsomDevelop:
A few misses from the original backend changes that were caught while testing the frontend.


packages/base/src/interfaces/dto/charging.station.dto.ts
```diff
@@ -14,7 +14,7 @@ import { PointSchema } from './types/location.js';
 import { OCPPVersionSchema } from './types/ocpp.message.js';
 
 export const ChargingStationSchema = BaseSchema.extend({
-  id: z.number().int().optional(),
```
yuvenus, line 17: whoops, i've been testing against the next branch, not this branch 😬 if this is the case, we'll want to change the frontend to extend ChargingStationSchema but make id an optional field 🤔 makes sense why we'd want it like this though

thanaParis (approved):

PR #706, opened 2026-05-15
Replace fs usages and update generate cert chain endpoint

lydiazcheng:
### Changes
1. Update generate cert endpoint: Create certificate chain and key files using file storage interface
2. Update reload endpoint: try loading these generated files at fist from the file storage interface if not exist then load from local disk
3. Remove update cert endpoint since no longer needed
### Tests
1. Generate certs
2. Before reload Screenshot 2026-05-14 at 5 19 36 PM
3. After reload Screenshot 2026-05-14 at 5 19 51 PM


thanaParis (approved):
approved!

PR #728, opened 2026-06-03
Bug/station and status update fix

devonNewsomDevelop:
Adds citrine to transpile list for local running
Changes how we open the hasura websocket for charging stations
Changes how we determine and set the status of a charging station for the UI and backend


apps/operator-ui/src/lib/client/pages/overview/online-status/online.status.card.tsx
```diff
@@ -27,15 +27,37 @@ export const OnlineStatusCard = () => {
   const translate = useTranslate();
 
   const {
-    query: { data, isLoading, error },
-  } = useCustom({
-    meta: {
-      gqlQuery: CHARGING_STATIONS_STATUS_COUNT_QUERY,
-    },
-  } as any);
+    query: { data: onlineData, isLoading: onlineLoading, error: onlineError },
```
yuvenus, line 30: can you not subscribe on the count query, with something like a useOne that always returns the result of the count aggregate? 👀  i guess technically this is similar to what's being done, but just curious why this approach and not use the `CHARGING_STATIONS_STATUS_COUNT_QUERY` (of which you should also be able to pass filters)

devonNewsomDevelop: useCustom doesn't have live updates so we had to move away from that and useOne afaik isn't designed for aggregate queries so I didn't use that. This seemed like the easiest way to go about the change from my limited understanding of refine. I haven't used it very long so this solution is really just a combination of stack overflow and refine docs and Claude all working together to help me understand how to best retrieve this data.

If you have a better solution I'm definitely all ears as I'm still learning regarding this framework.

yuvenus: ahh okay got it! that makes sense to me, was just wondering if there were alternatives but it seems like this works as needed 😄

yuvenus (approved):

apps/operator-ui/src/lib/client/pages/overview/online-status/online.status.card.tsx
devonNewsomDevelop: Oh for sure! I meant it, if you see a flaw or know of something better please let me know. I am just going off of what seems right to a Refine newbie but I'm definitely not confident this is the best way hahaha

thanaParis (approved):

PR #730, opened 2026-06-04
  Local Authorization List support for OCPP 1.6

ProgrammerDATCH:
Local Authorization List support for OCPP 1.6
  
  Adds management of the charging station Local Authorization List for OCPP 1.6 chargers, bringing it in line with the existing 2.0.1 support.

  What's included

  - SendLocalList (1.6) — new message endpoint + handler. Validates the request (version must increase, no duplicate idTags), persists it, and forwards it to the station. On the response it
  reconciles the stored LocalListVersion:
    - Accepted → applies the new version
    - VersionMismatch → automatically fires GetLocalListVersion to resync
    - Failed / NotSupported → logged
  - GetLocalListVersion (1.6) — new endpoint + handler that validates/replaces the locally stored list version for the station.
  - Differential deletes — entries sent without idTagInfo are recorded as tombstones and drop the matching authorization from the stored version.
  - 1.6 mapper (LocalAuthListMapper) — maps between Authorization status and 1.6 idTagInfo status, using flat idTag instead of the 2.0.1 IdToken structure.
  - GetLocalListVersion data API now eager-loads LocalListAuthorization entries.

  Fixes

  - SequelizeLocalAuthListRepository now uses LocalListVersion as its primary model (was Authorization), which fixes the "LocalListAuthorization is not associated to Authorization!" error when
  including authorization rows in queries.
  - Config loader hardening — config validation is now two-pass: stored configs that predate a newer optional-with-default field (e.g. websocketServers[*].protocols) auto-heal instead of crashing
  strict validation, while unknown/optional top-level keys are preserved instead of being stripped by Zod.


thanaParis:
@ProgrammerDATCH please repoint this to next.

PR #739, opened 2026-06-05
Fix case-sensitive apps/Server paths in docker compose files

SolidadeFabio:
## Problem

Both root-level compose files reference `./apps/server` (lowercase) while the actual directory in the repository is `apps/Server`. On case-sensitive filesystems (Linux), this breaks the Quick Start flow from the README in two ways:

1. **`docker-compose.local.yml`**  the `citrine` service fails to build:
   ```
   resolve : lstat /<repo>/apps/server: no such file or directory
   ```
   The same applies to the rabbitmq/postgres/minio volume mounts, which end up creating a new `apps/server` directory instead of reusing `apps/Server/data`.

2. **`docker-compose.yml`**  more subtle: Docker silently creates an empty `apps/server/hasura-metadata` directory and bind-mounts it into the `graphql-engine` container, so the Hasura metadata never loads (no error is raised).

On macOS and Windows the mismatch goes unnoticed because the default filesystems are case-insensitive, which is likely why this hasn't surfaced in CI or for most contributors.

## Fix

Reference `apps/Server` with the correct casing in both files (6 occurrences). Verified with `docker compose config` and a full `docker compose -f docker-compose.local.yml up -d --build` on Linux (Ubuntu, kernel 6.17).

### DIFF
--- a/docker-compose.local.yml
+++ b/docker-compose.local.yml
@@ -10,7 +10,7 @@ services:
   citrine:
     build:
       context: .
-      dockerfile: ./apps/server/deploy.Dockerfile
+      dockerfile: ./apps/Server/deploy.Dockerfile
     environment:
       APP_NAME: "all"
       APP_ENV: "docker"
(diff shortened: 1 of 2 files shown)


thanaParis:
Will not be relevant after https://github.com/citrineos/citrineos-core/pull/725

PR #746, opened 2026-06-11
Enable set password via set variable request

lydiazcheng:
### Changes
1. Remove crypto when store password
2. Compare plain password string with the stored password

### Test
1. send a set variable request contains BasicAuthPassword Screenshot 2026-06-11 at 4 59 30 PM
2. password is stored in db Screenshot 2026-06-11 at 5 00 56 PM


packages/core/src/util/networkconnection/authenticator/BasicAuthenticationFilter.ts
```diff
@@ -60,9 +60,18 @@ export class BasicAuthenticationFilter extends AuthenticatorFilter {
...
+            if (storedPassword === password) {
+              return true;
+            }
+            // Fall back to legacy PBKDF2 hash comparison for passwords stored before plain-text migration
+            try {
+              return CryptoUtils.isPasswordMatch(storedPassword, password);
+            } catch (error) {
+              this._logger.warn('Failed to compare password hash for', username, error);
+              return false;
+            }
```
lydiazcheng: I keep this as fallback for existed hashed password. Please let me know if I should remove it. @thanaParis Thank you!

thanaParis: Please remove it! A good idea, but I'd prefer to be rid of the bcrypt dependency and now is a good time for this shakeup given it's a new major version.

When we return to encrypted pw at rest, we'll probably try to use a different approach/library anyways

thanaParis (changes requested):

thanaParis (approved):

PR #747, opened 2026-06-15
Feature: Redefine object storage arguments

lydiazcheng:
### Changes
1. Sync arguments in file storage methods: use bucket and key instead of filePath and fileName
2. Rename configDir to configBucket
3. Make sue the fileStorage config is pass to module via CertificateAuthorityService 
### Test (Updated Jul 7)
1. generate cert chain via local and reload them
Screenshot 2026-07-07 at 9 50 55 AM
Screenshot 2026-07-07 at 9 51 41 AM
2. generate cert chain via s3 and reload them
Screenshot 2026-07-07 at 10 19 09 AM
Screenshot 2026-07-07 at 10 19 55 AM


lydiazcheng:
This PR doesn't change any frontend codes. But after pull the latest next, the e2e flow is triggered and failed with E2E-018 which is fixed in https://github.com/citrineos/citrineos-core/pull/757

packages/core/src/util/files/s3Storage.ts
```diff
@@ -179,7 +226,11 @@ export class S3Storage implements ConfigStore {
       const command = new CreateBucketCommand({ Bucket: bucket });
       await this.s3Client.send(command);
       this._logger.info(`Bucket "${bucket}" created successfully.`);
-    } catch (error) {
+    } catch (error: any) {
+      if (error.name === 'BucketAlreadyOwnedByYou' || error.name === 'BucketAlreadyExists') {
+        this._logger.debug(`Bucket "${bucket}" already exists.`);
+        return;
```
devonNewsomDevelop, line 232: Will this fail silently other than the log? Like is there any way for the user to know that this happened other than log monitoring?

devonNewsomDevelop: And does that matter if they don't know? It could be that this behavior is fine when running, just double checking!

devonNewsomDevelop (commented):
One small question on the PR otherwise everything looks good to me

packages/core/src/util/files/s3Storage.ts
lydiazcheng: Hi Devon, thanks for reviewing! This `createBucket` method is used when saving file while the bucket doesn't exist. So in general, this already exists error case won't happen.(I met once during my development and testing so add it just in case) And it is fine to just log, since saving file still works.

devonNewsomDevelop (approved):

PR #751, opened 2026-06-23
Fix: avoid overwrite existing variable attribute value when reject

lydiazcheng:
### Context
When get a rejected in response of getVariables.req, it should not overwrite the existing value for variable attribute.
### Tests
1. Set password Screenshot 2026-06-23 at 10 15 05 AM
2. Get password and the value is not overwrote after receiving a rejected response.  Screenshot 2026-06-23 at 10 16 04 AM Screenshot 2026-06-23 at 10 16 49 AM
3. [Regression] Manually update a variable attribute value in db  Screenshot 2026-06-23 at 10 25 52 AM
4. [Regression] Get variable and the value is updated. Screenshot 2026-06-23 at 10 27 33 AM Screenshot 2026-06-23 at 10 26 48 AM


apps/operator-ui/src/lib/cls/meter.value.dto.tsx
```diff
@@ -2,11 +2,7 @@
 //
```
lydiazcheng, line 2: This file contains only lint fixes. They are not related to this change, but after pulling the origin next, it seems that lint thinks they belongs to this change.

thanaParis (approved):

PR #757, opened 2026-06-30
Harden flaky locale e2e test (E2E-018)

AhmedAboulezz:
E2E-018 (overview › locale) is flaky across PRs (fails on next/ocpi/renovate, passes on others). Root cause is the in-place language switch: the switcher writes the NEXT_LOCALE cookie then `router.refresh()`es the server tree, and (a) the Radix dropdown option can be clicked before the menu finishes opening, and (b) the in-place RSC refresh can lag under CI load — so waiting on a single localized heading times out.

Fix (tests-only): wait for the option to be visible before clicking, then wait on the authoritative `<html lang>` flip as the signal the locale applied, with a reload fallback (the cookie write is the real contract, so a reload re-renders deterministically). Never masks a real break — if the language never applies even after reload, it still fails.

Verifying with a 7x re-run before merge.

### DIFF
--- a/apps/operator-ui/tests/e2e/specs/overview/locale.spec.ts
+++ b/apps/operator-ui/tests/e2e/specs/overview/locale.spec.ts
@@ -3,11 +3,50 @@
 // SPDX-License-Identifier: Apache-2.0
 
 import { test, expect } from '../../fixtures';
+import type { Page } from '@playwright/test';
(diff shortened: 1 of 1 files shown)


thanaParis:
test was flaky due to underlying issue with maps crashing; this has been fixed. the existing version of the test is arguably better due to catching this--we will keep it as-is.

PR #759, opened 2026-07-01
Change security event enum error handling

devonNewsomDevelop:
Introduces common method for checking the security event "enum" and the handling associated with that as well as unit tests exercising the code

### DIFF
--- a/packages/base/src/interfaces/dto/types/enums.ts
+++ b/packages/base/src/interfaces/dto/types/enums.ts
@@ -485,6 +485,29 @@ export const ReserveNowStatusEnumSchema = z.enum([
 
 export const ResetEnumSchema = z.enum(['Immediate', 'OnIdle', 'ImmediateAndResume']);
 
+export const SecurityEventNotificationTypeEnumSchema = z.enum([
+  'FirmwareUpdated',
+  'FailedToAuthenticateAtCsms',
+  'CsmsFailedToAuthenticate',
+  'SettingSystemTime',
+  'StartupOfTheDevice',
+  'ResetOrReboot',
+  'SecurityLogWasCleared',
+  'ReconfigurationOfSecurityParameters',
+  'MemoryExhaustion',
+  'InvalidMessages',
+  'AttemptedReplayAttacks',
+  'TamperDetectionActivated',
+  'InvalidFirmwareSignature',
+  'InvalidFirmwareSigningCertificate',
+  'InvalidCsmsCertificate',
+  'InvalidChargingStationCertificate',
+  'InvalidTLSVersion',
+  'InvalidTLSCipherSuite',
+  'MaintenanceLoginAccepted',
+  'MaintenanceLoginFailed',
+]);
+
 export const SendLocalListStatusEnumSchema = z.enum(['Accepted', 'Failed', 'VersionMismatch']);
 
 export const SetMonitoringStatusEnumSchema = z.enum([
@@ -614,6 +637,7 @@ export const RequestStartStopStatusEnum = RequestStartStopStatusEnumSchema.enum;
 export const ReservationUpdateStatusEnum = ReservationUpdateStatusEnumSchema.enum;
 export const ReserveNowStatusEnum = ReserveNowStatusEnumSchema.enum;
 export const ResetEnum = ResetEnumSchema.enum;
+export const SecurityEventNotificationTypeEnum = SecurityEventNotificationTypeEnumSchema.enum;
 export const SendLocalListStatusEnum = SendLocalListStatusEnumSchema.enum;
 export const SetMonitoringStatusEnum = SetMonitoringStatusEnumSchema.enum;
 export const SetNetworkProfileStatusEnum = SetNetworkProfileStatusEnumSchema.enum;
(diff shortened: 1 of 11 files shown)


lydiazcheng (approved):
The changes look good. Just wondering if we update the property definition in SecurityEventNotificationRequest.json instead. In this case, we don't need to add the corner case check everywhere the validation exists.

devonNewsomDevelop:
> The changes look good. Just wondering if we update the property definition in SecurityEventNotificationRequest.json instead. In this case, we don't need to add the corner case check everywhere the validation exists.

@lydiazcheng  This is an interesting idea. I didn't think this is something we could do because of the spec. Could you explain a little more what you mean? Do you mean we would get rid of the check altogether from the JSON schema? I think we still want that, just not to throw an error anymore. Still, I like the idea of not validating everywhere...so if you don't mind could you explain what you were thinkin' for this one?

lydiazcheng:
> > The changes look good. Just wondering if we update the property definition in SecurityEventNotificationRequest.json instead. In this case, we don't need to add the corner case check everywhere the validation exists.
> 
> @lydiazcheng This is an interesting idea. I didn't think this is something we could do because of the spec. Could you explain a little more what you mean? Do you mean we would get rid of the check altogether from the JSON schema? I think we still want that, just not to throw an error anymore. Still, I like the idea of not validating everywhere...so if you don't mind could you explain what you were thinkin' for this one?

Hi @devonNewsomDevelop , since we rely on the json schema for ocpp validation, my idea is to comment the `enum` part in SecurityEventNotificationRequest.json for 2.0.1 and 2.1. Then we don't need to have `isUnknownSecurityEventTypeOnly`. We also have some custom updates in json schemas like authorize and transaction event request. So I think such update is acceptable. But I am also good with the current solution.
```
      "enum": [
        "FirmwareUpdated",
        "FailedToAuthenticateAtCsms",
        "CsmsFailedToAuthenticate",
        "SettingSystemTime",
        "StartupOfTheDevice",
        "ResetOrReboot",
        "SecurityLogWasCleared",
        "ReconfigurationOfSecurityParameters",
        "MemoryExhaustion",
        "InvalidMessages",
        "AttemptedReplayAttacks",
        "TamperDetectionActivated",
        "InvalidFirmwareSignature",
        "InvalidFirmwareSigningCertificate",
        "InvalidCsmsCertificate",
        "InvalidChargingStationCertificate",
        "InvalidTLSVersion",
        "InvalidTLSCipherSuite",
        "MaintenanceLoginAccepted",
        "MaintenanceLoginFailed"
      ]

devonNewsomDevelop:
> > > The changes look good. Just wondering if we update the property definition in SecurityEventNotificationRequest.json instead. In this case, we don't need to add the corner case check everywhere the validation exists.
> > 
> > 
> > @lydiazcheng This is an interesting idea. I didn't think this is something we could do because of the spec. Could you explain a little more what you mean? Do you mean we would get rid of the check altogether from the JSON schema? I think we still want that, just not to throw an error anymore. Still, I like the idea of not validating everywhere...so if you don't mind could you explain what you were thinkin' for this one?
> 
> Hi @devonNewsomDevelop , since we rely on the json schema for ocpp validation, my idea is to comment the `enum` part in SecurityEventNotificationRequest.json for 2.0.1 and 2.1. Then we don't need to have `isUnknownSecurityEventTypeOnly`. We also have some custom updates in json schemas like authorize and transaction event request. So I think such update is acceptable. But I am also good with the current solution.
> 
> ```
>       "enum": [
>         "FirmwareUpdated",
>         "FailedToAuthenticateAtCsms",
>         "CsmsFailedToAuthenticate",
>         "SettingSystemTime",
>         "StartupOfTheDevice",
>         "ResetOrReboot",
>         "SecurityLogWasCleared",
>         "ReconfigurationOfSecurityParameters",
>         "MemoryExhaustion",
>         "InvalidMessages",
>         "AttemptedReplayAttacks",
>         "TamperDetectionActivated",
>         "InvalidFirmwareSignature",
>         "InvalidFirmwareSigningCertificate",
>         "InvalidCsmsCertificate",
>         "InvalidChargingStationCertificate",
>         "InvalidTLSVersion",
>         "InvalidTLSCipherSuite",
>         "MaintenanceLoginAccepted",
>         "MaintenanceLoginFailed"
>       ]
> ```

Oh, okay I see what you mean. I think that per the OCPP spec we still have to acknowledge this enum and emit a log for it, but I get what you're saying. I think I'd prefer your solution but I'm not sure that we can do that and still be considered OCPP compliant in this aspect?

lydiazcheng:
> > > > The changes look good. Just wondering if we update the property definition in SecurityEventNotificationRequest.json instead. In this case, we don't need to add the corner case check everywhere the validation exists.
> > > 
> > > 
> > > @lydiazcheng This is an interesting idea. I didn't think this is something we could do because of the spec. Could you explain a little more what you mean? Do you mean we would get rid of the check altogether from the JSON schema? I think we still want that, just not to throw an error anymore. Still, I like the idea of not validating everywhere...so if you don't mind could you explain what you were thinkin' for this one?
> > 
> > 
> > Hi @devonNewsomDevelop , since we rely on the json schema for ocpp validation, my idea is to comment the `enum` part in SecurityEventNotificationRequest.json for 2.0.1 and 2.1. Then we don't need to have `isUnknownSecurityEventTypeOnly`. We also have some custom updates in json schemas like authorize and transaction event request. So I think such update is acceptable. But I am also good with the current solution.
> > ```
> >       "enum": [
> >         "FirmwareUpdated",
> >         "FailedToAuthenticateAtCsms",
> >         "CsmsFailedToAuthenticate",
> >         "SettingSystemTime",
> >         "StartupOfTheDevice",
> >         "ResetOrReboot",
> >         "SecurityLogWasCleared",
> >         "ReconfigurationOfSecurityParameters",
> >         "MemoryExhaustion",
> >         "InvalidMessages",
> >         "AttemptedReplayAttacks",
> >         "TamperDetectionActivated",
> >         "InvalidFirmwareSignature",
> >         "InvalidFirmwareSigningCertificate",
> >         "InvalidCsmsCertificate",
> >         "InvalidChargingStationCertificate",
> >         "InvalidTLSVersion",
> >         "InvalidTLSCipherSuite",
> >         "MaintenanceLoginAccepted",
> >         "MaintenanceLoginFailed"
> >       ]
> > ```
> 
> Oh, okay I see what you mean. I think that per the OCPP spec we still have to acknowledge this enum and emit a log for it, but I get what you're saying. I think I'd prefer your solution but I'm not sure that we can do that and still be considered OCPP compliant in this aspect?

Maybe @thanaParis can help on this question?

thanaParis:
I think it makes more sense to extract the enum definition, alter the json schema to use type string, eliminate the edge condition in the validation, and add a handler for SecurityEventRequest which checks the string against the enum and validates there.

This makes the most sense to me since the type of the field in the actual spec is string--the fact that it's an enum in the json schema is, in my opinion, an outright error.

thanaParis (approved):

PR #772, opened 2026-07-10
fix(router): drop stale OCPP Calls past maxCallLengthSeconds instead of routing them to a later connection

biubiuzhang:
## Problem

`MessageRouter` can deliver a cached/queued OCPP Call to a **later, different** websocket connection with no expiry. During websocket churn a Call is requeued (`channel.nack` defaults to `requeue=true`, and no `x-message-ttl` is set on the queues), and when a new connection for the same station comes up, the aged Call is drained onto it.

Because the router never checks the age of a queued Call before sending it via the network hook, an operator command executes against a runtime state the caller never saw. Concretely:

- A **Reset** requested minutes earlier lands on a station that has since started a *fresh* charging session and kills it.
- A stale **RequestStop / RemoteStopTransaction** closes a *different* transaction than intended.
- Any routed Call (`ChangeAvailability`, `SetChargingProfile`, `UnlockConnector`, …) can act on the wrong session.

This was surfaced by a SIL regression test where a `Reset` POSTed during connection churn was delivered **~4 minutes later** to the next connection (the router logged `Call sent successfully with 0 ms of lag` — the "0 ms" only measures the final send, not the true queue age, which hid the staleness).

## Root cause

There is no age/TTL check anywhere between enqueue and `_networkHook`. `sendCall` sets a `maxCallLengthSeconds` TTL only on the *transaction-namespace tracker* keyed by `correlationId` (the "call already in progress" guard) — it does not bound the lifetime of the Call itself.

## Fix

In `AbstractMessageRouter.handle()`, drop any Request-state Call whose `context.timestamp` is older than `maxCallLengthSeconds` before routing it. Once a Call is older than the window in which its response would still be accepted, delivery is never correct, so it is discarded (and logged) instead of routed on.

`handle()` is a single choke point that:
- covers **every** Call action (not just Reset), and
- catches **requeued** Calls on redelivery (the nack/requeue loop re-enters `handle()`), so the fix holds regardless of the transport's requeue behavior.

`context.timestamp` is set once at publish time (`RequestBuilder.buildCall`) and preserved across requeues, so it reflects the true issue time.

## Scope / follow-ups (not in this PR)

- Only Request-state Calls are gated here (the operator-command path). Response routing is left unchanged.
- Making the REST caller distinguish "delivered" vs "queued/dropped" (e.g. an error via the callback URL) is a larger, separate change.

## Testing

- Added unit tests in `MessageRouterImpl.test.ts`: a Call older than `maxCallLengthSeconds` is dropped (neither `sendCall` nor the network hook is invoked); a fresh Call is routed normally.
- Full `MessageRouterImpl` suite passes (57 tests).

---

Provided by the **Enteligent team** (internal ref: TSW-2676).

### DIFF
--- a/packages/base/src/config/types.ts
+++ b/packages/base/src/config/types.ts
@@ -306,6 +306,7 @@ export const systemConfigInputSchema = z.object({
   logLevel: z.number().min(0).max(6).default(0).optional(),
   maxCallLengthSeconds: z.number().int().min(1).default(20).optional(),
   maxCachingSeconds: z.number().int().min(1).default(30).optional(),
+  staleCallMaxAgeSeconds: z.number().int().min(1).optional(),
   maxReconnectDelay: z.number().int().min(1).default(30).optional(),
   shutdownGracePeriodSeconds: z.number().int().min(1).default(30).optional(),
   ocpiServer: z.object({
@@ -632,6 +633,7 @@ export const systemConfigSchema = z
     logLevel: z.number().min(0).max(6),
     maxCallLengthSeconds: z.number().int().min(1),
     maxCachingSeconds: z.number().int().min(1),
+    staleCallMaxAgeSeconds: z.number().int().min(1).optional(),
     maxReconnectDelay: z.number().int().min(1).default(30),
     shutdownGracePeriodSeconds: z.number().int().min(1).default(30),
     ocpiServer: z.object({
--- a/packages/base/src/interfaces/router/AbstractRouter.ts
+++ b/packages/base/src/interfaces/router/AbstractRouter.ts
@@ -125,6 +125,28 @@ export abstract class AbstractMessageRouter implements IMessageRouter {
           );
         }
 
+        // Optionally drop Calls that have aged past `staleCallMaxAgeSeconds` while queued. A
+        // Call can sit in the broker (e.g. requeued during websocket churn while the station
+        // has no live or valid connection) and then be delivered to a later, different
+        // connection — acting on a runtime state the caller never saw (a stale Reset killing a
+        // fresh session, a stale RemoteStop closing someone else's transaction, etc.). This
+        // guard is opt-in: it only drops when `staleCallMaxAgeSeconds` is configured, since some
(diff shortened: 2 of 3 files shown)


thanaParis (changes requested):
Please point this PR at the next branch.

I understand the problem you are trying to address, but this approach has some issues. MaxCallLengthSeconds is an OCPP protocol value. Charging stations will have it configured and respect it when performing retries etc. What you are using it for here--as a maximum wait time for queued requests--is a detrimental overload. These should be separately configurable.

There's also the question of whether this is universally applicable: do we always want 'stale' messages to be dropped? Some messages, such as certificate installations, getting diagnostics, or setting configuration, likely need to be sent no matter what, even if the wait time on the queue is excessive.

I would be okay with this feature as an optional, separately configurable value for dropping stale calls. There's more work that needs to be done here, eventually, to provide robust mechanisms for handling staleness in a fine-grained manner, but that shouldn't get in the way of a simpler approach being available for those who find it sufficient.

biubiuzhang:
Thanks @thanaParis, that all makes sense — reworked along those lines:

1. **Retargeted at `next`.**
2. **No more overloading `maxCallLengthSeconds`.** Added a dedicated, optional `staleCallMaxAgeSeconds` config key. The stale-drop in `AbstractMessageRouter.handle()` now keys off that value and is a no-op unless it's set.
3. **Opt-in / not universally forced.** It's off by default, so existing deployments are completely unchanged, and any deployment that would rather deliver late than not at all simply leaves it unset. That also covers the must-always-send messages (certificate installs, GetDiagnostics, SetConfiguration) for anyone who doesn't opt in.

I agree the robust, fine-grained per-action staleness handling is a larger effort — I've left that as future work (noted in the code comment) so this simpler mechanism is available for those who find it sufficient, per your suggestion.

Rebased onto `next`; unit tests updated to cover drop-when-enabled, route-within-window-when-enabled, and route-aged-when-disabled. Full `MessageRouterImpl` suite passes.

thanaParis (approved):

thanaParis (approved):

PR #778, opened 2026-07-14
adding metrics for websockets

thanaParis:


packages/core/src/modules/OcppRouter/src/module/router.ts
```diff
@@ -434,9 +468,19 @@ export class MessageRouterImpl extends AbstractMessageRouter implements IMessage
...
+        ocpp_version: protocol,
+        outcome: success ? 'sent' : 'send_failed',
+      });
       this._logger.debug(`CallResult sent successfully ${correlationId}`, identifier, message);
       return { success };
     } else {
+      ocppCallResultSentTotal.add(1, {
+        action: String(action),
+        ocpp_version: protocol,
+        outcome: 'action_mismatch',
```
devonNewsomDevelop: Worth making some sort of string "enum" for these outcome strings? I tend to shy away from raw string discriminators

packages/core/src/util/networkconnection/WebsocketNetworkConnection.ts
```diff
@@ -139,33 +152,38 @@ export class WebsocketNetworkConnection implements INetworkConnection {
...
+      wsSendFailureTotal.add(1, { reason: 'no_cache' });
       throw new Error(errorMsg);
     }
 
     const websocketConnection = this._identifierConnections.get(identifier);
     if (!websocketConnection) {
       const errorMsg = 'Websocket connection not found for ' + identifier;
-      this._logger.fatal(errorMsg);
+      connLogger.fatal(errorMsg);
+      wsSendFailureTotal.add(1, { reason: 'no_socket' });
```
devonNewsomDevelop: Same thought with the actions and reasons, worth creating enums for these so we can avoid typos in the future?

packages/core/src/util/metrics.ts
```diff
@@ -0,0 +1,133 @@
...
+
+/** Failures sending to a station, by `reason`: no_cache | no_socket | not_open | send_error. */
+export const wsSendFailureTotal = meter.createCounter('ocpp_ws_send_failure_total', {
+  description: 'Failures sending a message to a charging station, by reason',
+});
+
+// --- OCPP message flow -------------------------------------------------------
+
+/** Every inbound OCPP frame, by `message_type` (Call|CallResult|CallError) and `ocpp_version`. */
+export const ocppMessageReceivedTotal = meter.createCounter('ocpp_message_received_total', {
```
devonNewsomDevelop: It feels like these actions would be good as string type "enums" too for the same reason

devonNewsomDevelop (commented):
Approved with suggestions. Your call on if the enums would be valuable or not since youre closer to the code!

devonNewsomDevelop (approved):

PR #779, opened 2026-07-15
User uploaded filepath validation

devonNewsomDevelop:
This change plugs holes with our file path validation as pointed out by the community member Santosh Kumar Puppala (https://github.com/Santoshkumarpuppala)

We ensure no absolute paths, no relative path parent traversal, and no NUL bytes to break certain file storage solutions. Adds tests to the suite testing each condition. We purposefully do not guard against config files.

### DIFF
--- a/packages/core/src/modules/Certificates/src/module/DataApi.ts
+++ b/packages/core/src/modules/Certificates/src/module/DataApi.ts
@@ -36,7 +36,7 @@ import {
   UploadExistingCertificate,
   UploadExistingCertificateSchema,
 } from '@dal/interfaces/index.js';
-import { generateCertificate } from '@util/index.js';
+import { generateCertificate, validateSafeFilePath } from '@util/index.js';
 import type { FastifyInstance, FastifyRequest } from 'fastify';
 import jsrsasign from 'jsrsasign';
 import moment from 'moment';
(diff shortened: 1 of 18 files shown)


packages/core/src/modules/Certificates/src/module/DataApi.ts
```diff
@@ -300,6 +304,7 @@ export class CertificatesDataApi
 
     let rootCAPem: string;
     if (installReq.fileId) {
+      validateSafeFilePath(installReq.fileId);
```
thanaParis: this approach in general makes sense, but putting the checks here as opposed to in the file storage implementations means we will constantly be at risk of re-opening this issue. we should move the checks there.

thanaParis (changes requested):

packages/core/src/modules/Certificates/src/module/DataApi.ts
devonNewsomDevelop: Yeah, you're right. I was thinking about making some sort of file path save extension for this reason. I think your solution makes sense too.  I did it this way because we have those config paths that I didnt want to get too messy with flagging but I think that small cost is better than having to remember to add the validate to every new file storage implementation

thanaParis (approved):

PR #788, opened 2026-07-20
Fix: sendLocalList Max Entries

vicky-ahmed:


yuvenus:
relevant issue: https://github.com/citrineos/citrineos/issues/203

thanaParis (approved):

PR #789, opened 2026-07-20
Fix: Coordinate Precision For GeoLocation

vicky-ahmed:


packages/ocpi-base/src/mapper/LocationMapper.ts
```diff
@@ -42,6 +42,10 @@ import { ParkingType } from '../model/ParkingType.js';
 import { Facilities } from '../model/Facilities.js';
 import type { Hours } from '../model/Hours.js';
 
+export function formatCoordinate(value: number | string): string {
+  return Number(value).toFixed(5);
```
yuvenus: just curious, are the coordinates in the database stored with at least up to 5 precision as well? would it be worth making sure that they are from the operator-ui (or even up to 7 precision, assuming this even originates from operator-ui) or is that not necessary?

vicky-ahmed: Hey Venus, thanks for the review. The DB stores coordinates very comfortably as a PostGIS, and operator-ui already emits 6–7 decimals via the address autocomplete. Yes, the coordinates do come from the operator UI, and CitrineOS's own GeoLocationSchema actually requires 5–7 decimal places, with coordinates being a mandatory field. So the precision is already handled at the schema level, and this formatter just ensures we're outputting a consistent format. Let me know if that answers your question.

packages/ocpi-base/src/mapper/LocationMapper.ts
```diff
@@ -56,8 +60,8 @@ export class LocationMapper {
       state: location.state,
       country: location.country,
       coordinates: {
-        longitude: location.coordinates.coordinates[0].toString(),
```
thanaParis, line 59: @vicky-ahmed this being the source of the bug, assuming the database has more than 4 digits of precision, doesn't make sense. toString() does not truncate, and toFixed() will manufacture extra digits as 0 if they are not present.

Please check:
- what an example of a full precision value in the db looks like
- what hasura is returning in the graphql body
- what value is reaching this line of code

since toString doesn't truncate afaik, the precision loss must happen somewhere else in the chain

thanaParis (changes requested):

packages/ocpi-base/src/mapper/LocationMapper.ts
vicky-ahmed: Hey Thana, thanks for the feedback. So, I checked and the database value, the GraphQL response, and the value reaching the mapper. All of them look correct and are not causing precision loss on their own.
For example, here's a sample of what the database stores:
SRID=4326;POINT(-119.41796656 36.778365) and that's 8 decimal places of precision, so the data is definitely there at the source.
image


So, as the value with 4 decimal places was being rejected the current PR uses toFixed(5) to format the coordinate. If the value has fewer than 5 decimal places, it adds zeros to meet the schema requirement, but if it has more than 5, it truncates and loses precision.

So, two possible things that we can do:

Keep the current approach so that it doesnt cause a rejection but it will cause precision loss in certain cases.
Add a validation when a location is created that rejects coordinates with less than 5 decimal places.

Or if we could we remove the schema requirement, but since it's a spec requirement, that's not really an option 

Do we have any other ideas or a preferred direction?

packages/ocpi-base/src/mapper/LocationMapper.ts
yuvenus: follow up: if there's the possibility of a data issue, perhaps this format should only occur if:

the coordinates are less than 5 precision OR the coordinates are greater than 7 precision

in which case you'd set to 5/7 precision (or just 5... or 7... whatever is preferred). then we don't punish data that's already formatted properly 🤔

thanaParis (approved):

PR #792, opened 2026-07-22
Fix: Live Logs Toggle Fix On Switching Between Tabs

vicky-ahmed:


apps/operator-ui/src/lib/client/pages/charging-stations/detail/ocpp.messages.tsx
```diff
@@ -68,7 +72,10 @@ export const OCPPMessages: React.FC<OCPPMessagesProps> = ({
   const [selectedOrigin, setSelectedOrigin] = useState<string>(allOption);
   const [filters, setFilters] = useState<LogicalFilter[]>([]);
   const [exportDialogOpen, setExportDialogOpen] = useState(false);
-  const [liveLogEnabled, setLiveLogEnabled] = useState(false);
+  const [internalLiveLogEnabled, setInternalLiveLogEnabled] = useState(false);
```
yuvenus: why does there need to be an internal state? why not just trigger onLiveLogEnabledChange whenever the live log toggle is toggled? is this preventing some sort of weird state thing?

vicky-ahmed: True actually, makes sense. I'll update. Thanks!

apps/operator-ui/src/lib/client/pages/charging-stations/detail/ocpp.messages.tsx
```diff
@@ -60,6 +62,8 @@ export const OCPPMessages: React.FC<OCPPMessagesProps> = ({
   stationId,
   initialStartDate = null,
   initialEndDate = null,
+  liveLogEnabled: liveLogEnabledProp,
```
yuvenus: rather than needing an additional const on line 76, it probably makes more sense to add a default value, i.e. `liveLogEnabled = false`

vicky-ahmed: Makes sense, updated. Thanks!

yuvenus (approved):

PR #795, opened 2026-07-23
fix(auth): surface IdTokenInfo.groupIdToken in OCPP 2.0.1/2.1 Authorize

biubiuzhang:
## Description

`AuthorizationMapper.toIdTokenInfo` (OCPP 2.0.1 and 2.1) never sets `groupIdToken`, even though:

- the `Authorization` model has a self-referential `groupAuthorization` relation (`groupAuthorizationId` → `Authorization`), and
- the OCPP `IdTokenInfo` type supports `groupIdToken`.

As a result, group membership is silently dropped from `AuthorizeResponse` (and the `IdTokenInfo` embedded in `SendLocalList`), so a charging station can never learn a token's group — breaking group-based concurrent-transaction handling and any group-scoped local-authorization logic on the station side.

## Changes

- `mapper/2.0.1/AuthorizationMapper.ts` and `mapper/2.1/AuthorizationMapper.ts`: `toIdTokenInfo` now maps `groupIdToken` from the related `groupAuthorization` (reusing `toIdToken`) when present.
- `repository/Authorization.ts`: eager-load the `groupAuthorization` association in `_constructQuery` so `toIdTokenInfo` can surface it during the authorize lookup.

Guarded by presence of the relation, so tokens without a group are unaffected (still emit no `groupIdToken`).

## Not included

`IdTokenInfo.evseId` (per-EVSE allow-list) is still not mapped. Unlike `groupIdToken` (which had a backing relation that simply wasn't mapped), there is **no backing column** on `Authorization` for `evseId` — surfacing it needs a schema addition, so it's left for a separate feature PR.

## How tested

- Added a unit test (`packages/core/.../mapper/2.0.1/AuthorizationMapper.test.ts`) covering the `groupIdToken` mapping (present-relation and no-relation cases) — passes under vitest.
- `tsc --noEmit` on `@citrineos/core` passes.
- Verified against a running CitrineOS + an OCPP 2.0.1 charging station: provisioned a member Authorization whose `groupAuthorizationId` points at a group Authorization; `Authorize` response now returns `idTokenInfo.groupIdToken = { idToken, type }` of the group. (Before: field absent.)

## Checklist
- [x] Signed-off-by (DCO)
- [x] SPDX headers unchanged
- [x] Unit test for the mapper mapping

### DIFF
--- a/packages/core/src/dal/layers/sequelize/mapper/2.0.1/AuthorizationMapper.ts
+++ b/packages/core/src/dal/layers/sequelize/mapper/2.0.1/AuthorizationMapper.ts
@@ -34,6 +34,10 @@ export class AuthorizationMapper {
       language2: authorization.language2,
       personalMessage: authorization.personalMessage,
       customData: authorization.customData,
+      // groupIdToken must be eager-loaded (include groupAuthorization) to be surfaced here.
+      groupIdToken: authorization.groupAuthorization
+        ? AuthorizationMapper.toIdToken(authorization.groupAuthorization)
+        : undefined,
     };
   }
 
--- a/packages/core/src/dal/layers/sequelize/mapper/2.1/AuthorizationMapper.ts
+++ b/packages/core/src/dal/layers/sequelize/mapper/2.1/AuthorizationMapper.ts
@@ -35,6 +35,10 @@ export class AuthorizationMapper {
       language2: authorization.language2,
       personalMessage: authorization.personalMessage,
       customData: authorization.customData,
+      // groupIdToken must be eager-loaded (include groupAuthorization) to be surfaced here.
+      groupIdToken: authorization.groupAuthorization
+        ? AuthorizationMapper.toIdToken(authorization.groupAuthorization)
+        : undefined,
     };
   }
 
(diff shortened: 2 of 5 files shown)


thanaParis:
Please point to next. Looks good otherwise, though a unit test for the 2.1 mapper change as well would be appreciated.

biubiuzhang:
Thanks for the review — both points addressed.

**Pointed at `next`:** the PR now targets `next` and the branch is rebased onto it, so the diff is back to 5 files / 1 commit.

**2.1 mapper test:** added `packages/core/test/dal/layers/sequelize/mapper/2.1/AuthorizationMapper.test.ts` covering `toIdTokenInfo` — `groupIdToken` mapped when the `groupAuthorization` association is eager-loaded, and omitted when it is not — and added the matching `toIdTokenInfo` cases to the existing 2.0.1 suite (`AuthenticationMapper.test.ts`, which covers `AuthorizationMapper`). Both new assertions fail against the unpatched mappers and pass with the fix; the 2.0.1 file is 55/55 green.

Tests follow the `test/` layout and the `aAuthorization` provider convention already used on `next`.

thanaParis (approved):

PR #798, opened 2026-07-25
feat(operator-ui): generic extension slot system

LoveKapibarasan:
## Summary
- Adds a plugin-style mechanism for embedding operator-facing tools inside Operator-UI without editing its code per tool.
- Extensions are declared via `public/extensions/*.json` manifests (`{id, label, icon, internalUrl}`); installing/removing one is just adding/removing a file.
- `/api/extensions` exposes a public-safe list (internalUrl is stripped server-side) that the sidebar renders automatically via a generic loop.
- `/extensions/[id]` dynamically loads the extension's UI bundle and mounts it directly into the page's own React tree (no iframe).
- `/extensions/[id]/proxy/*` is a same-origin reverse proxy to the extension's backend, so extensions never need their own public route or reverse-proxy config.
- `/api/extensions/token` mints a short-lived signed token (reusing `NEXTAUTH_SECRET`) so an extension can call its own API without a separate login screen.

## Scope
This PR only contains the generic mechanism. No specific extension manifest is included.

## Test plan
- [ ] `pnpm --filter @citrineos/operator-ui build`
- [ ] Manually verify existing pages/navigation are unaffected
- [ ] Drop a sample `public/extensions/*.json` manifest and confirm a sidebar item + `/extensions/{id}` route appear with no other code changes

### DIFF
--- a/apps/operator-ui/src/app/(authenticated)/extensions/[id]/page.tsx
+++ b/apps/operator-ui/src/app/(authenticated)/extensions/[id]/page.tsx
@@ -4,11 +4,7 @@
 import { getExtension } from '@lib/server/extensions';
 import { ExtensionFrame } from '@lib/client/components/extension-frame/extension-frame';
 
-export default async function ExtensionPage({
-  params,
-}: {
-  params: Promise<{ id: string }>;
-}) {
+export default async function ExtensionPage({ params }: { params: Promise<{ id: string }> }) {
   const { id } = await params;
   const extension = getExtension(id);
 
--- a/apps/operator-ui/src/app/(authenticated)/extensions/[id]/page.tsx
+++ b/apps/operator-ui/src/app/(authenticated)/extensions/[id]/page.tsx
@@ -1,0 +1,28 @@
+// SPDX-FileCopyrightText: 2025 Contributors to the CitrineOS Project
+//
+// SPDX-License-Identifier: Apache-2.0
+import { getExtension } from '@lib/server/extensions';
+import { ExtensionFrame } from '@lib/client/components/extension-frame/extension-frame';
+
+export default async function ExtensionPage({
+  params,
+}: {
+  params: Promise<{ id: string }>;
+}) {
+  const { id } = await params;
+  const extension = getExtension(id);
+
+  if (!extension) {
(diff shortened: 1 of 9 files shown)


LoveKapibarasan:
## Extension-side contract

This PR only adds the host-side mechanism. For reference, here is what an extension needs to implement to work with it (not part of this PR — extensions are deployed separately via `public/extensions/*.json`, which is intentionally not included here).

**Manifest** (`public/extensions/{id}.json`):
```json
{
  "id": "my-tool",
  "label": "My Tool",
  "icon": "Wrench",
  "internalUrl": "http://localhost:PORT"
}
```
`icon` is either a name from the built-in map in `main.menu.tsx` (currently `Wrench`, `Puzzle`) or an `https://` image URL.

**Routing**: Operator-UI proxies the *full* incoming path (including the `/extensions/{id}/proxy` prefix) to `internalUrl`. If the extension is a Next.js app, this means it should set a matching `basePath: '/extensions/{id}/proxy'` so its own router strips the same prefix Operator-UI left in place. Any other framework just needs to route on that same prefix.

**UI bundle**: the extension must serve a JS file (e.g. `extension-bundle.js`) that, once loaded via `<script>`, exposes a mount function on `window`:
```ts
window.MyExtension = {
  mount(container: HTMLElement, props: { apiBase: string; getToken: () => Promise<string> }) {
    // render using the *hosts own* window.React / window.ReactDOM
    // (do not bundle your own copy of React/ReactDOM — this keeps it a single
    // React tree, no iframe, no version-mismatch risk)
    return () => { /* unmount */ };
  },
};
```
- `apiBase` — prefix your own API calls with this (e.g. `fetch(`${apiBase}/api/...`)`), so they go through the same-origin proxy.
- `getToken()` — resolves to a short-lived Bearer token signed with the *same* `NEXTAUTH_SECRET`. Your own backend should accept `Authorization: Bearer <token>` on top of (or instead of) its normal session and verify it the same way `/api/extensions/token` mints it (`purpose: 'extension-handoff'`, short `maxAge`).

Happy to add this as a short `docs/extensions.md` in a follow-up if useful.

LoveKapibarasan:
This will still take some time, but I am planning to build a feature like this.

PR #802, opened 2026-07-27
Adding types package and rewiring the project to use it

devonNewsomDevelop:
Cleanup of where we have our types live. Operator UI no longer depends on base, and base + core will use this package now for types.


yuvenus:
woohoo! just to note, in `base/src/money` both `Currency.ts` and `Money.ts` seem like fair candidates to be moved into types -- would we consider moving them in this PR or in another one? to be very fair, i don't know which folder they'd belong to if moved to types 😅

devonNewsomDevelop:
> woohoo! just to note, in `base/src/money` both `Currency.ts` and `Money.ts` seem like fair candidates to be moved into types -- would we consider moving them in this PR or in another one? to be very fair, i don't know which folder they'd belong to if moved to types 😅

@yuvenus  Ah yeah these were a point of deliberation. Chose to leave them out since they are constructed classes that do work at runtime and not just types. They should live in base (in my opinion)!

yuvenus:
> > woohoo! just to note, in `base/src/money` both `Currency.ts` and `Money.ts` seem like fair candidates to be moved into types -- would we consider moving them in this PR or in another one? to be very fair, i don't know which folder they'd belong to if moved to types 😅
> 
> @yuvenus Ah yeah these were a point of deliberation. Chose to leave them out since they are constructed classes that do work at runtime and not just types. They should live in base (in my opinion)!

that's fair! they'll continue holding down the `base/src/money` fort then 😆

thanaParis (approved):

PR #805, opened 2026-07-28
feat(operator-ui): backend sort, type filter, and multi-column search added on Configuration tab

vicky-ahmed:
image

Pre-populated fields in set variable modal.
image

### DIFF
--- a/apps/operator-ui/src/lib/client/pages/charging-stations/detail/charging.station.configuration.tsx
+++ b/apps/operator-ui/src/lib/client/pages/charging-stations/detail/charging.station.configuration.tsx
@@ -2,7 +2,7 @@
 //
 // SPDX-License-Identifier: Apache-2.0
 
-import { OCPP2_0_1, OCPPVersion } from '@citrineos/base';
+import { AttributeEnum, OCPPVersion } from '@citrineos/base';
 import { ChangeConfigurationModal } from '@lib/client/components/modals/1.6/change-configuration/change.configuration.modal';
 import { DebounceSearch } from '@lib/client/components/debounce-search';
 import { MultiSelect } from '@lib/client/components/multi-select';
@@ -75,6 +75,9 @@ interface ChargingStationConfigurationProps {
   id: number;
 }
 
+// Which config view is shown. '1.6' -> ChangeConfigurations; '2.0.1'/'2.1' -> VariableAttributes.
+type ConfigVersion = '1.6' | '2.0.1' | '2.1';
+
 const CONFIG_1_6_COLUMNS = [
   { key: 'key', headerKey: 'ChargingStations.configuration.key', accessor: 'key' },
   {
@@ -116,13 +119,13 @@ const CONFIG_2_0_1_COLUMNS = [
   },
 ];
 
-const ATTRIBUTE_TYPE_OPTIONS = Object.values(OCPP2_0_1.AttributeEnumType);
+const ATTRIBUTE_TYPE_OPTIONS = Object.values(AttributeEnum);
 
 export const ChargingStationConfiguration: React.FC<ChargingStationConfigurationProps> = ({
   id,
 }) => {
   const translate = useTranslate();
-  const [version, setVersion] = useState<'1.6' | '2.0.1'>('1.6');
+  const [version, setVersion] = useState<ConfigVersion>('1.6');
   const [searchTerm, setSearchTerm] = useState('');
   const [selectedTypes, setSelectedTypes] = useState<string[]>([]);
   const [sort, setSort] = useState<{ field: string; order: 'asc' | 'desc' } | null>(null);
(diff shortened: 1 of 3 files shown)


apps/operator-ui/src/lib/client/pages/charging-stations/detail/charging.station.configuration.tsx
```diff
@@ -99,12 +116,16 @@ const CONFIG_2_0_1_COLUMNS = [
   },
 ];
 
+const ATTRIBUTE_TYPE_OPTIONS = Object.values(OCPP2_0_1.AttributeEnumType);
```
devonNewsomDevelop: We are trying to get away from the OCPP2_0_1 specific enums where possible. Do we potentially have this enum in Enums.ts in base? If not, do you think it's reasonable to add this one there?

apps/operator-ui/src/lib/client/pages/charging-stations/detail/charging.station.configuration.tsx
```diff
@@ -129,10 +150,28 @@ export const ChargingStationConfiguration: React.FC<ChargingStationConfiguration
   const station = data?.data;
   const isConnected = !!station?.isOnline;
 
-  const attributeFilters = useMemo<CrudFilter[]>(
-    () => [{ field: 'stationId', operator: 'eq', value: id }],
-    [id],
-  );
+  const attributeFilters = useMemo<CrudFilter[]>(() => {
+    const filters: CrudFilter[] = [{ field: 'stationId', operator: 'eq', value: id }];
+    if (version === '2.0.1') {
```
devonNewsomDevelop: This will miss version 2.1, which I think we don't want. Better as !== '1.6' I think

apps/operator-ui/src/lib/client/pages/charging-stations/detail/charging.station.configuration.tsx
```diff
@@ -422,18 +477,47 @@ export const ChargingStationConfiguration: React.FC<ChargingStationConfiguration
   const renderTable = () => {
     const columns = version === '1.6' ? CONFIG_1_6_COLUMNS : CONFIG_2_0_1_COLUMNS;
     const isLoading = version === '2.0.1' ? isAttributesLoading : isConfigurationsLoading;
+    const isFetching = version === '2.0.1' ? isAttributesFetching : isConfigurationsFetching;
```
devonNewsomDevelop: Same version stuff here. I think we need to update this file to account for 2.1, sorry about that, I know it was like that beforehand

apps/operator-ui/src/lib/client/pages/charging-stations/detail/charging.station.configuration.tsx
```diff
@@ -499,15 +581,28 @@ export const ChargingStationConfiguration: React.FC<ChargingStationConfiguration
...
-              <SelectTrigger className="flex-1">
-                <SelectValue />
-              </SelectTrigger>
-              <SelectContent>
-                <SelectItem value="1.6">OCPP 1.6</SelectItem>
-                <SelectItem value="2.0.1">OCPP 2.0.1</SelectItem>
-              </SelectContent>
-            </Select>
+            <div className="w-full">
+              <Select value={version} onValueChange={(v: '1.6' | '2.0.1') => setVersion(v)}>
```
devonNewsomDevelop: missing 2.1 in this area

devonNewsomDevelop (changes requested):

apps/operator-ui/src/lib/client/pages/charging-stations/detail/charging.station.configuration.tsx
vicky-ahmed: Yes, it is exported from the base Enums.ts. I replaced OCPP2_0_1.AttributeEnumType with the Enums from base.

apps/operator-ui/src/lib/client/pages/charging-stations/detail/charging.station.configuration.tsx
devonNewsomDevelop: I would verify with Mason but right now if someone was to connect on 2.1 with this code, none of this would function right?

vicky-ahmed: Yeah thats true, I mean I'm not sure what we would show for 2.1.
@masonseaman thoughts?

vicky-ahmed: @devonNewsomDevelop or we could show the same data as we are for 201, for now. The check would be if version !==1.6.

apps/operator-ui/src/lib/client/pages/charging-stations/detail/charging.station.configuration.tsx
vicky-ahmed: nice catch. on it

apps/operator-ui/src/lib/client/pages/charging-stations/detail/charging.station.configuration.tsx
devonNewsomDevelop: Yeah, I was thinking that would likely be best unless there is some major discrepancy in 2.0.1 and 2.1 regarding this specific content

devonNewsomDevelop (approved):

thanaParis (approved):

PR #811, opened 2026-07-29
fix: edit button is hidden in case of readonly mode, mutabily column added

vicky-ahmed:
image

### DIFF
--- a/apps/operator-ui/src/lib/client/pages/charging-stations/detail/charging.station.configuration.tsx
+++ b/apps/operator-ui/src/lib/client/pages/charging-stations/detail/charging.station.configuration.tsx
@@ -53,6 +53,8 @@ interface VariableAttribute {
   id: string;
   type: string;
   value: string;
+  mutability?: string | null;
+  constant?: boolean | null;
   Variable: {
     name: string;
     instance: string;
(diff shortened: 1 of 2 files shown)


apps/operator-ui/src/lib/client/pages/charging-stations/detail/charging.station.configuration.tsx
```diff
@@ -284,13 +292,16 @@ export const ChargingStationConfiguration: React.FC<ChargingStationConfiguration
...
+          mutability: attribute.mutability ?? '-',
           component: `${attribute.Component?.name ?? '-'}:${attribute.Component?.instance ?? '-'}`,
           variable: `${attribute.Variable?.name ?? '-'}:${attribute.Variable?.instance ?? '-'}`,
           evse: `${attribute.Evse?.id ?? '-'}:${attribute.Evse?.connectorId ?? '-'}`,
           componentName: attribute.Component?.name ?? '',
           componentInstance: attribute.Component?.instance ?? null,
           variableName: attribute.Variable?.name ?? '',
           variableInstance: attribute.Variable?.instance ?? null,
+          // Whether this attribute can be set on the charger — drives the Edit action.
+          isEditable: attribute.mutability !== 'ReadOnly' && !attribute.constant,
```
devonNewsomDevelop, line 304: The only thing here is that if we are using 'ReadOnly' in other places in this file, we should make it a string literal. If this is the only place, it's fine imo

devonNewsomDevelop (approved):
Approve with suggestion

apps/operator-ui/src/lib/client/pages/charging-stations/detail/charging.station.configuration.tsx
vicky-ahmed: Thanks for the suggestion. ReadOnly isn't anywhere else in the file, so it shouldn't cause any problems, so we're good on that.

PR #813, opened 2026-07-29
Feature/drizzle

thanaParis:
Setup for drizzle refactor. Will be completed across several releases.


apps/ocpp-server/src/container.ts
```diff
@@ -305,11 +308,20 @@ function registerRepositories(container: AwilixContainer): void {
...
   });
 
-  if (process.env.CITRINEOS_USE_DRIZZLE_SECURITY_EVENT === 'true') {
+  if (process.env.CITRINEOS_USE_DRIZZLE === 'true') {
     container.register({
       securityEventRepository: asFunction(
         ({ config, logger }) => new DrizzleSecurityEventRepository(config, logger),
       ).singleton(),
+      subscriptionRepository: asFunction(
+        ({ config, logger }) => new DrizzleSubscriptionRepository(config, logger),
```
devonNewsomDevelop: Just curious, you couldn't use asClass here for it to pick up the config and the logger?

packages/core/src/dal/layers/drizzle/repository/AsyncJobStatus.ts
```diff
@@ -0,0 +1,62 @@
...
+
+export class DrizzleAsyncJobStatusRepository extends DrizzleRepository<
+  typeof asyncJobStatusTable,
+  AsyncJobDto
+> {
+  constructor(
+    config: BootstrapConfig,
+    logger?: Logger<ILogObj>,
+    db?: NodePgDatabase,
+    useTenantSchema = false,
```
devonNewsomDevelop: This is totally up to you and depends on your preferences. I find it confounding that typescript doesn't used named commands so you have to just pass `(....otherArgs, false | true)` as a parameter and then the developer has to go and figure out what the heck false or true is, so recently on that file validation PR, you probably noticed I had made it an Options object. 

If you thought it was valuable, you could do the same here! 

`Interface TenantOptions {
  useTenantSchema: boolean
}`

and usage would look like

`(...otherArgs, {useTenenatSchema: false | true})` 

Where it's more explicit what the boolean is. Not a big deal and not a change needed to be made if you think it's too much boiler plate for this little bit.

devonNewsomDevelop (approved):
Approved with suggestions. 

I didnt scan every single Drizzle schema so if there are any you want me to take a closer look at let me know! Looks cool and I like how you set up the dto pattern.

PR #816, opened 2026-07-30
Feature/ocpp message model refactor

thanaParis:
changes:
- type column added and used in place of 'state' in various places in code; corresponds to messageTypeId field of ocpp json structure
- message column replaced with payload and raw: payload is ocpp payload field or { errorCode, errorDescription, errorDetails } for callErrors, raw is string as received/sent over the wire
- setting action and requestId fields for responses now handled on a db level by triggers with a hash-based transaction lock on tenantId+stationId+correlationId. This eliminates the possibility of race conditions on setting these fields and improves performance.


packages/core/src/modules/OcppRouter/src/module/router.ts
```diff
@@ -958,6 +972,7 @@ export class MessageRouterImpl extends AbstractMessageRouter implements IMessage
     action: CallAction,
     timestamp: Date,
     protocol: OCPPVersionType,
+    rawMessage: string,
   ): Promise<IMessageConfirmation> {
     const messageId = message[1];
     const payload = new OcppError(messageId, message[2], message[3], message[4]);
```
devonNewsomDevelop: We really should have some sort of message util or something that grabs the pieces that we need in a safe way and describes them better than this. Right now this is somewhat black box magic that requires the developer to already know the message array structure and which components arrive when, and to never mistake the ordering. Not a part of this change, exactly, but it seems like we're continuing that pattern. 

If not changed in this PR we should probably add it to a future improvement IMO

packages/core/src/modules/OcppRouter/src/module/router.ts
```diff
@@ -1009,9 +1025,10 @@ export class MessageRouterImpl extends AbstractMessageRouter implements IMessage
       message.context.ocppConnectionName,
       message.context.timestamp,
       message.protocol,
-      message.action,
-      message.state,
+      rawMessage,
+      rpcMessage[0],
```
devonNewsomDevelop: Regarding the message Util mentioned a few lines above-

Like here for example, I don't know what rpcMessage[0] is. I can infer it's *probably* the meat of the message, but I have no idea.

packages/core/src/modules/EVDriver/src/module/module.ts
```diff
@@ -903,7 +903,7 @@ export class EVDriverModule extends AbstractModule {
           reserveStatus: status,
           isActive: status === ReserveNowStatusEnum.Accepted,
         },
-        request.message[3].id,
+        request.payload.reservationId,
```
lydiazcheng: [minor] the name of the field is `id` in `ReserveNowRequest`.

lydiazcheng (approved):

packages/core/src/modules/OcppRouter/src/module/router.ts
```diff
@@ -236,15 +236,27 @@ export class MessageRouterImpl extends AbstractMessageRouter implements IMessage
       messageId = rpcMessage[1];
       switch (messageTypeId) {
         case MessageTypeId.Call: {
-          await this._onCall(identifier, rpcMessage as Call, timestamp, protocol);
+          await this._onCall(identifier, rpcMessage as Call, timestamp, protocol, message);
           break;
         }
         case MessageTypeId.CallResult: {
-          await this._onCallResult(identifier, rpcMessage as CallResult, timestamp, protocol);
+          await this._onCallResult(
```
devonNewsomDevelop: This may be a pre-existing bug, but there appears to be a nested logging issue here.

This calls routeCallresult which will call emitMessage, and supposing that the message emit fails to send (with a success:False) then we will dispatchMessageReceived and log the error properly. 

However, when that success false returns back to onCallResult, we will then throw because the confirmation.success is false, and log AGAIN in the catch for onCallResult, which will cause a second insert with less descriptive data and an orphaned correlation ID.

devonNewsomDevelop (changes requested):

devonNewsomDevelop (approved):
These changes rock

PR #817, opened 2026-07-30
feat(operator-ui): add resizable station preview panel to station lists

vicky-ahmed:
image

Transaction Screen:
image


devonNewsomDevelop:
> image

@masonseaman  @vicky-ahmed  Do you guys think it would be worth changing the nested "Name" to something like "Charger" or something?

apps/operator-ui/src/lib/client/components/authenticated-layout/index.tsx
```diff
@@ -93,15 +108,38 @@ export default function AuthenticatedLayout({
...
-              <div className="content-inner-wrap">
-                <>
-                  <HeaderBanner />
-                  {children}
-                </>
+          <main className={`content-container flex min-h-0 flex-1 flex-col ${routeClassName}`}>
+            <div className="content-outer-wrap flex min-h-0 flex-1 flex-col">
+              <div className="content-inner-wrap flex min-h-0 flex-1 flex-col">
+                <HeaderBanner />
+                <ResizablePanelGroup
```
yuvenus: don't change the authenticated layout, it's probably best to make some sort of wrapper component for charging stations (or perhaps just a generic resizable wrapper component if we wanna do this for other components), then include charging stations list and station preview in there 🤔

apps/operator-ui/src/lib/client/pages/charging-stations/station.preview.panel.tsx
```diff
@@ -0,0 +1,301 @@
...
+
+export interface StationPreviewPanelProps {
+  /** The ChargingStations.id to preview. */
+  stationId: number;
+  /** Called when the panel's close button is clicked. */
+  onClose: () => void;
+}
+
+// Same compact power formatting as connectors/connectors.table.tsx.
+const formatPower = (value?: number | null) =>
```
yuvenus: if this is used more than once, probably worth throwing an exported const utils

apps/operator-ui/package.json
```diff
@@ -84,6 +84,7 @@
     "react-dom": "19.1.4",
     "react-hook-form": "7.65.0",
     "react-redux": "9.2.0",
+    "react-resizable-panels": "^3.0.6",
```
yuvenus: remove caret

apps/operator-ui/src/lib/client/components/authenticated-layout/index.tsx
vicky-ahmed: Hmm that is the better approach, you're right. I just thought of going with this as the changes were not just in charging stations table but also in location and transactions table. But, yeah let me see what I can do. I'll I guess have to wrap all of these tables into the wrapper component.

apps/operator-ui/src/lib/client/pages/transactions/columns.tsx
```diff
@@ -68,12 +71,30 @@ export const getTransactionsColumns = (translate?: TranslateFn): ColumnConfigura
       header: t('Transactions.columns.stationId', 'Station ID'),
       visible: true,
       sortable: true,
-      cellRender: ({ row }: CellContext<TransactionClass, unknown>) => (
-        <TableCellLink
-          path={`/${MenuSection.CHARGING_STATIONS}/${row.original.chargingStation?.id}`}
-          value={row.original.chargingStation?.ocppConnectionName ?? EMPTY_VALUE}
-        />
-      ),
+      cellRender: ({ row }: CellContext<TransactionClass, unknown>) => {
```
yuvenus, line 74: any screenshots of this? 👀

apps/operator-ui/src/lib/client/pages/charging-stations/station.preview.panel.tsx
vicky-ahmed: Good call, it was indeed duplicated in two files, I have extracted it into a shared formatPower util.

apps/operator-ui/src/lib/client/pages/charging-stations/station.preview.panel.tsx
```diff
@@ -0,0 +1,320 @@
...
+                    {serverId}
+                  </Link>
+                ) : (
+                  <span>{NOT_APPLICABLE}</span>
+                )
+              }
+            />
+
+            {/* Collapsible EVSEs & Connectors — operators generally care about this at a glance. */}
+            <div className="border-t pt-3">
```
yuvenus: didn't realize we never had a collapsible component :O can you go ahead and add it from shadcn and use it here instead? https://ui.shadcn.com/docs/components/base/collapsible

yuvenus: the EVSE collapsible and connectors (if they aren't already) should auto default to open i think -- save folks an extra click

vicky-ahmed: The collapsible component is added and set it default to open.

yuvenus (approved):

PR #826, opened 2026-08-04
refactor(operator-ui): extract reusable GraphQL field-set consts with pick/omit

vicky-ahmed:


apps/operator-ui/src/lib/queries/authorizations.ts
```diff
@@ -12,24 +13,11 @@ export const AUTHORIZATIONS_LIST_QUERY = gql`
...
-      groupAuthorizationId
-      additionalInfo
-      concurrentTransaction
-      chargingPriority
-      language1
-      language2
-      personalMessage
-      cacheExpiryDateTime
+      ${AUTHORIZATION_FIELDS}
       allowedConnectorTypes
```
yuvenus: shouldn't these be part of authorization fields?

apps/operator-ui/src/lib/queries/locations.ts
```diff
@@ -162,59 +101,32 @@ export const LOCATIONS_GET_QUERY = gql`
...
-      id
-      name
-      address
-      city
-      postalCode
-      state
-      country
-      coordinates
+      ${LOCATION_CORE_FIELDS}
       facilities
```
yuvenus: should these fields also be in LOCATION_CORE_FIELDS?

apps/operator-ui/src/lib/queries/authorizations.ts
vicky-ahmed: Yeah, they belong in the const. The reason I trimmed it was that transactions and tariffs fetch fewer auth fields but now we can omit in the query.

apps/operator-ui/src/lib/queries/locations.ts
```diff
@@ -13,31 +19,12 @@ export const LOCATIONS_LIST_QUERY = gql`
...
-      postalCode
-      state
-      country
-      coordinates
-      createdAt
-      updatedAt
-      timeZone
-      parkingType
+      ${LOCATION_CORE_FIELDS}
+      ${LOCATION_DETAIL_FIELDS.pick('timeZone', 'parkingType')}
```
yuvenus: for the locations table now we actually show all the detail fields also, so you don't need pick, just include all the detail fields 👍

vicky-ahmed: Done.

thanaParis (approved):

PR #832, opened 2026-08-06
fix(handlers): correlate NotifyCustomerInformation and NotifyDisplayMessages by requestId against the payload column

biubiuzhang:
## Problem

Every `NotifyCustomerInformation` and `NotifyDisplayMessages` sent by a charging station is rejected with `CALLERROR PropertyConstraintViolation` ("RequestId was not provided in a ...Request."), even when the station echoes the exact `requestId` the CSMS sent.

Both handlers validate the notification by looking up the originating request in `OCPPMessages`:

```ts
where: {
  ...
  action: OCPP_CallAction.CustomerInformation,
  message: { requestId: requestId },
}
```

The OCPP message model refactor (#816) replaced the `message` column with `payload` (parsed OCPP payload, JSONB) and `raw` (wire frame, TEXT), but these two queries were carried over unchanged. They now reference a column that no longer exists, so the lookup fails and the handler always errors — the notification data is lost.

This replaces #808, which fixed the same correlation failure in the pre-refactor code (there the column existed but held the raw RPC frame array, so the object match never matched). The refactor moved the logic from `Reporting`/`Configuration` module files into the per-message handler classes, which is why that PR no longer applied. With `payload` now storing the parsed payload object, the fix reduces to matching against the right column.

## Fix

Query `payload: { requestId }` instead of `message: { requestId }` in:

- `packages/core/src/handlers/requests/2/NotifyCustomerInformationRequestOcpp2Handler.ts`
- `packages/core/src/handlers/requests/2/NotifyDisplayMessagesRequestOcpp2Handler.ts`

## Tests

- Added unit tests for both handlers covering the accept path (asserting the repository is queried by `payload.requestId`) and the reject path (no matching request → CallError). `vitest run` on the two files: 4/4 pass.
- `tsc --noEmit` over `packages/core`: identical pre-existing error count before and after the change; no new errors.
- The original defect was observed end-to-end against a real OCPP 2.0.1 station: all `CustomerInformation` variants were `Accepted` by the charger, and every resulting `NotifyCustomerInformation` was rejected by the CSMS. With the correlation fixed (equivalent change on our fork), the station receives a proper `NotifyCustomerInformationResponse` CallResult.

### DIFF
--- a/packages/core/src/handlers/requests/2/NotifyCustomerInformationRequestOcpp2Handler.ts
+++ b/packages/core/src/handlers/requests/2/NotifyCustomerInformationRequestOcpp2Handler.ts
@@ -57,7 +57,7 @@ export class NotifyCustomerInformationRequestOcpp2Handler extends AbstractHandle
           tenantId: message.context.tenantId,
           ocppConnectionName: message.context.ocppConnectionName,
           action: OCPP_CallAction.CustomerInformation,
-          message: {
+          payload: {
             requestId: requestId,
           },
         },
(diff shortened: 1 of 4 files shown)


thanaParis (approved):
Looks great. Thank you for the pr!

PR #833, opened 2026-08-06
Adding Authorization Drizzle Repository

yuvenus:
This PR adds the Drizzle-version of the `Authorization` repository. Changes:

1. Added new Drizzle-version of `Authorization` repository and referenced it in `RepositoryStore` and `container` for dependency injection.
2. Changed repository return type for IAuthorizationRepository interface methods from `Authorization` to `AuthorizationDto`.
3. Any non-data layer usages (i.e. in the handlers, services, etc.) of the Sequelize entity `Authorization` were replaced with bespoke repository interface methods and therefore implemented in both the Sequelize and Drizzle versions of the repositories.
4. `AuthorizationQuerystring` was made more flexible by including the `id` field and making the `idToken` field optional -- this supports the usage of the repository methods across more use cases.
5. Optional `tariff` relationship field added to `AuthorizationDto` to assist with one of the handler use cases.

### DIFF
--- a/apps/ocpp-server/src/container.ts
+++ b/apps/ocpp-server/src/container.ts
@@ -121,6 +121,7 @@ import {
   WebhookDispatcher,
   WebsocketNetworkConnection,
 } from '@citrineos/core';
+import { DrizzleAuthorizationRepository } from '@citrineos/core/dist/src/dal/layers/drizzle/index.js';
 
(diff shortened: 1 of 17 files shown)


thanaParis (approved):
You fixed the security event test! thank you 🙏

PR #838, opened 2026-08-07
Fix: Avoid overwrite value of write-only variable attributes by GetBaseReport

lydiazcheng:


packages/core/src/dal/layers/sequelize/repository/DeviceModel.ts
```diff
@@ -156,14 +156,23 @@ export class SequelizeDeviceModelRepository
           },
         );
         if (!variableAttributeCreated) {
+          const mutability = variableAttribute.mutability ?? savedVariableAttribute.mutability;
+          // B08.FR.03 The Charging Station sends the requested information,
+          // excluding the value of WriteOnly variables
+          // the existing value should not be overwritten in this case.
+          const preserveValue =
+            variableAttribute.value === undefined &&
```
thanaParis: This check seems brittle. If it's write only, we shouldn't update--value could be undefined, null, empty string, 0--it shouldn't matter I don't think.

thanaParis (changes requested):

packages/core/src/dal/layers/sequelize/repository/DeviceModel.ts
lydiazcheng: Thanks for the review, Thana! I updated the check.

thanaParis (approved):

PR #843, opened 2026-08-10
Adding Boot (and VariableAttribute) Drizzle Repository

yuvenus:
This PR introduces a Drizzle-version of the Boot entity. Since Boot happens to use VariableAttribute, VariableAttribute was also Drizzle-ified. Changes:

1. Implemented IBootRepository methods into the Drizzle-version of Boot.
2. Simplified the IBootRepository "update" methods into one singular `updateByKey` method, since `updateStatusByKey` wasn't used, and there were more usages for a general update method.
3. Implemented a general `updateAllByQueryString` method into the Drizzle-version of VariableAttribute, but left a TODO in the conditions helper method for another PR since there aren't currently more usages of the conditions.
4. Updated all relevant imports (RepositoryStore.ts, container.ts, etc.) to include imports for DrizzleBootRepository and DrizzleVariableAttributeRepository. (and also fixed the DrizzleAuthorizationRepository import 😳)


packages/core/src/dal/interfaces/repositories.ts
```diff
@@ -89,26 +88,16 @@ export interface IAuthorizationRepository {
...
-    statusInfo: OCPP2_common_types.StatusInfoType | undefined,
-    key: string,
-  ) => Promise<Boot | undefined>;
-  updateLastBootTimeByKey: (
-    tenantId: number,
-    lastBootTime: string,
-    key: string,
-  ) => Promise<Boot | undefined>;
-  readByKey: (tenantId: number, key: string) => Promise<Boot | undefined>;
+  ) => Promise<BootDto | undefined>;
```
devonNewsomDevelop, line 95: Just checking, but I assume that Create is out of scope for this context? Like we wouldn't create a charging station record here? Just checking since it's the only real CRUD missing, but I figure it's very likely intentional

packages/core/src/dal/layers/drizzle/repository/Boot.ts
```diff
@@ -33,7 +37,33 @@ export function toBootDto(entity: BootEntity): BootDto {
   return dto;
 }
 
-export class DrizzleBootRepository extends DrizzleRepository<typeof bootTable, BootDto> {
+// Required to convert lastBootTime back into Date
+function toBootEntity(value: object): BootEntity {
+  const v = value as { lastBootTime?: unknown };
+  if (typeof v.lastBootTime === 'string') {
```
devonNewsomDevelop, line 43: The typing here has me concerned, haha. How many different types can lastBootTime be?

packages/core/src/dal/layers/drizzle/repository/Boot.ts
```diff
@@ -42,5 +72,149 @@ export class DrizzleBootRepository extends DrizzleRepository<typeof bootTable, B
     return toBootDto(row);
   }
 
-  // Domain query/write methods intentionally omitted — stub outline only.
+  // ──────────── IBootRepository methods ─────────────────────────────
+  // Note that for many of the methods below, we purposely DO NOT use the equivalent Drizzle methods
+  // because they only accept key as a number, but Boot stores id as a string equal to the
+  // ocppConnectionName of the relevant station.
```
devonNewsomDevelop, line 78: Just curious, and I understand if this is how it has to be done without getting real gnarly with it- but can we not use the pkid for the charging stations here instead of the ocppConnectionName? That way, we could just use the built in drizzle stuff, right? Does the flow makes it gross to try and achieve this?

packages/core/src/dal/layers/drizzle/repository/Boot.ts
```diff
@@ -42,5 +72,149 @@ export class DrizzleBootRepository extends DrizzleRepository<typeof bootTable, B
...
+  }
+
+  async deleteByKey(tenantId: number, key: string): Promise<BootDto | undefined> {
+    const rows = (await this.db
+      .delete(bootTable)
+      .where(and(eq(bootTable.tenantId, tenantId), eq(bootTable.id, key)))
+      .returning()) as BootEntity[];
+
+    if (!rows[0]) return undefined;
+    const dto = this.toDto(rows[0]);
```
devonNewsomDevelop, line 158: If rows gets deleted above this, does it ever return not undefined? Could be that drizzle creates a copy of the row before deleting it and that's what's going on here, just wanted to verify.

mrnonaki:
While preparing a fork port we independently hit the missing `DrizzleAuthorizationRepository` barrel export and had a standalone fix ready — dropping it in favor of this PR since it covers the same gap (plus the `container.ts` deep-import cleanup). 👍 from a downstream consumer's perspective: deep-importing from `dist/` is the only workaround today, so getting this merged unblocks embedding the Drizzle repos properly.

packages/core/src/dal/layers/drizzle/repository/VariableAttribute.ts
```diff
@@ -50,5 +52,36 @@ export class DrizzleVariableAttributeRepository extends DrizzleRepository<
...
+  }
+
+  async updateAllByQueryString(
+    query: VariableAttributeQuerystring,
+    value: object,
+  ): Promise<VariableAttributeDto[]> {
+    const rows = (await this.db
+      .update(variableAttributeTable)
+      .set(value)
+      .where(and(...this.createVariableAttributeConditions(query)))
```
devonNewsomDevelop, line 78: If createVariableAttributeConditions is empty, it just becomes an unfiltered update, right? Do we want that behavior?

devonNewsomDevelop (commented):
Some questions here and there

packages/core/src/dal/interfaces/repositories.ts
yuvenus: Yup, that's it! We just don't happen to create BootNotifications in a context that's outside of the DAL (phew!). If we did, we'd want to add an interface method here.

packages/core/src/dal/layers/drizzle/repository/VariableAttribute.ts
yuvenus: Yup that's fine! The `and()` is smart enough to not freak out on empty list, and simply grabs everything. 👍

packages/core/src/dal/layers/drizzle/repository/Boot.ts
yuvenus: LOL I was being a little paranoid here, but you're right lastBootTime can only be a string or Date -- updated accordingly. 👌

packages/core/src/dal/layers/drizzle/repository/Boot.ts
yuvenus: The biggest hurdle is honestly just that it's a little annoying to get the charging station pkId, because most of these operations happen inside the BootNotificationRequest handlers, which will come with the ocppConnectionName 😅 But it's not so devastating that it's a dealbreaker...I think. This could be something we take back and figure out if it's worth the effort for sure though 🤔

packages/core/src/dal/layers/drizzle/repository/Boot.ts
yuvenus: Do you mean if the row gets deleted by a different transaction (like I guess two people just trying to delete the Boot at the same time)? 

AFAIK since this is a PostgreSQL invention (delete... returning) and Drizzle's just wrapping around it, in theory if someone stealth deleted before you, the transaction would fail because it wasn't able to delete in the first place 🤔

thanaParis (approved):

PR #846, opened 2026-08-11
fix(core): don't let a failed message-audit insert kill the process

biubiuzhang:
## Problem

A single failed insert into `OCPPMessages` terminates the whole CitrineOS process.

`WebhookDispatcher.dispatchMessageReceived` and `dispatchMessageSent` both `await this._ocppMessageRepository.createOCPPMessage(...)` **before** their `try` blocks, so a persistence failure is unguarded:

https://github.com/citrineos/citrineos-core/blob/main/packages/core/src/modules/OcppRouter/src/module/webhook.dispatcher.ts#L170-L196

The sibling method `dispatchMessageReceivedUnparsed` in the same file wraps its whole body in `try`/`catch`, which is what makes the omission visible.

This is made fatal rather than merely lossy by the call site: `dispatchMessageReceived` is awaited from inside a **`finally`** block in `MessageRouterImpl.onMessage`. A throw inside `finally` bypasses the surrounding `try`/`catch` entirely, so the rejection escapes `onMessage` and reaches Node as an unhandled rejection:

```
node:internal/process/promises:394
    triggerUncaughtException(err, true /* fromPromise */);

SequelizeForeignKeyConstraintError
  insert or update on table "OCPPMessages"
  violates foreign key constraint "OCPPMessages_requestMessageId_fkey"
    at WebhookDispatcher.dispatchMessageReceived (webhook.dispatcher.js:103)
    at MessageRouterImpl.onMessage (router.js:197)
```

## Impact

In our deployment this crashed the server **313 times**. Every crash drops the websocket for every connected charging station mid-session. Because authorization on our chargers travels over OCPP, a crash during a session stalls authorization and the session times out before charging starts.

The failure is intermittent and looks like flaky charging rather than a CSMS fault, which made it expensive to track down.

## Fix

Guard both inserts so a persistence failure is logged and routing continues. The audit record is a side effect of routing, not a precondition for it — a CSMS should not stop routing OCPP traffic because one row could not be written.

Also reorders the `action` fallback so the missing-record guard runs *before* the debug log that dereferences `messageRecord.action`.

## Notes

- The FK violation itself is only the trigger; this PR deliberately does not address why a `requestMessageId` can reference a missing request row. Even with that fixed, an unguarded audit insert should not be able to terminate the process.
- No behaviour change on the success path.


thanaParis (approved):
approved! good catch, thank you for the PR.

PR #847, opened 2026-08-11
fix(core): guard null generatedAt in VariableAttribute getter to prevent 500 on device-model read

vicky-ahmed:


packages/core/src/dal/layers/sequelize/model/DeviceModel/VariableAttribute.ts
```diff
@@ -167,7 +167,7 @@ export class VariableAttribute
   @Column({
     type: DataType.DATE,
     get() {
-      return this.getDataValue('generatedAt').toISOString();
+      return this.getDataValue('generatedAt')?.toISOString() ?? null;
     },
   })
   declare generatedAt: string;
```
yuvenus, line 173: is it necessary to change generatedAt to be nullable? i.e. `declare generatedAt?: string | null` ?

vicky-ahmed: I feel like it’s not necessary for this fix. Making it string | null or optional causes other parts of the code to break because they expect generatedAt to always be a string. The current getter safely handles `null` at runtime, So for this one-line crash fix, I recommend keeping the current string type. Thoughts?

yuvenus: gotcha, then in that case should empty string be returned instead to avoid unintended side effects of returning null? i.e. `return this.getDataValue('generatedAt')?.toISOString() ?? '';` ? just trying to avoid a different null check issue 🤔 

also, can generatedAt be null in the DB, or is it actually an empty string?

vicky-ahmed: generatedAt is NULL in the DB (the timestamp column cannot contain an empty string), so getDataValue() returns Date | null. I'd avoid returning empty string because it isn't a valid ISO datetime and would fail DTO validation, null correctly represents "no timestamp" and is consistent with the other DAL date fields.

yuvenus: cool, makes sense to me!

yuvenus (approved):

PR #848, opened 2026-08-11
fix(core): don't reject an OCPP 1.6 idTag that matches several authorizations

biubiuzhang:
## Problem

`TransactionService.authorizeOcpp16IdToken` rejects a perfectly valid idTag whenever more than one `Authorizations` row shares that token value — even when **every** matching row says `Accepted`.

https://github.com/citrineos/citrineos-core/blob/main/packages/core/src/modules/Transactions/src/module/TransactionService.ts#L316-L327

```ts
const authorizations = await this._authorizeRepository.readAllByQuerystring(tenantId, {
  idToken: idToken,          // no type — OCPP 1.6 has none to give
});
if (authorizations.length !== 1) {
  this._logger.error(`Found invalid authorizations ... for idToken: ${idToken}`);
  return response;           // idTagInfo.status = Invalid
}
```

Two things make this reachable rather than a corrupt-data case:

1. `Authorizations` has **no unique constraint** on `idToken` (only the `id` primary key), so the same value under two `idTokenType`s is a legitimate state.
2. An **OCPP 1.6 `idTag` is a bare string with no token type**, so the lookup cannot disambiguate.

The `length !== 1` guard is sound for the 2.0.1 and 2.1 paths a few lines above, because those query with `idToken.idToken` **and** `idToken.type`:

```ts
readAllByQuerystring(tenantId, { idToken: idToken.idToken, type: idToken.type });
```

The 1.6 path reuses the same guard without the type filter, so the identical database contents work over 2.0.1 and fail over 1.6.

## Impact

Observed on a local deployment with two `Accepted` rows for the same idTag (`ISO14443` and `NoAuthorization`):

- **every** `StartTransaction` answered `{"idTagInfo":{"status":"Invalid"},"transactionId":0}` — 15 for 15
- libocpp correctly honours that per OCPP 1.6 and stops the transaction with reason `DeAuthorized`, ~100 ms after it started
- no session ever reached charging; the charge point behaved correctly throughout

Diagnosing this is unnecessarily hard because a duplicated token and an unknown token produce the same log line (`Found invalid authorizations …`), and the CSMS answers `Invalid` in both cases.

## Fix

- Report the no-match case separately from the multiple-match case.
- On multiple matches, resolve deterministically and **fail safe**: honour the first non-`Accepted` match, so a `Blocked`/`Expired` duplicate still blocks. Only when every match is `Accepted` does the token authorize.
- Log which authorization was used and the types that matched, so the duplicate is visible.

This keeps the security property of the original code (an ambiguous set containing a restrictive entry is not silently upgraded to `Accepted`) while no longer rejecting a token whose matches unanimously say `Accepted`.

## Notes

- Scoped to the 1.6 path only; the 2.0.1 and 2.1 paths are untouched since their type filter already makes the single-match assumption hold.
- An alternative would be a unique constraint on `(tenantId, idToken)`, but that would be a breaking migration for existing deployments and would remove a per-type distinction that 2.x legitimately relies on.
- No behaviour change when exactly one authorization matches.

### DIFF
--- a/packages/core/src/modules/Transactions/src/module/TransactionService.ts
+++ b/packages/core/src/modules/Transactions/src/module/TransactionService.ts
@@ -318,13 +318,32 @@ export class TransactionService {
       const authorizations = await this._authorizationRepository.readAllByQuerystring(tenantId, {
         idToken: idToken,
       });
-      if (authorizations.length !== 1) {
-        this._logger.error(
-          `Found invalid authorizations ${JSON.stringify(authorizations)} for idToken: ${idToken}`,
-        );
+      if (authorizations.length === 0) {
+        this._logger.error(`No authorization found for idToken: ${idToken}`);
         return response;
       }
-      const authorization = authorizations[0];
+      // An OCPP 1.6 idTag is a bare string with no token type, so the same value can
+      // legitimately be stored under several idTokenTypes (Authorizations has no unique
+      // constraint on idToken, and the 2.0.1/2.1 paths above disambiguate with
+      // idToken.type — 1.6 has nothing to disambiguate with). Resolve deterministically
+      // and fail safe: honour the first non-Accepted match so a Blocked/Expired duplicate
+      // still blocks, rather than rejecting a token whose matches all say Accepted.
+      let authorization = authorizations[0];
+      if (authorizations.length > 1) {
+        const restrictive = authorizations.find(
+          (candidate) =>
+            candidate.status &&
+            OCPP1_6_Mapper.AuthorizationMapper.toStartTransactionResponseStatus(
+              candidate.status,
+            ) !== OCPP1_6.StartTransactionResponseStatus.Accepted,
+        );
+        authorization = restrictive ?? authorization;
+        this._logger.warn(
+          `Found ${authorizations.length} authorizations for idToken ${idToken} ` +
+            `(types: ${authorizations.map((candidate) => candidate.idTokenType).join(', ')}); ` +
+            `using authorization ${authorization.id} (status ${authorization.status}).`,
+        );
+      }
 
       // Check expiration and status
       if (!authorization.status) {


thanaParis:
I don't agree with the logic here.

If we take the scenario being brought up seriously, the proposed fix doesn't seem an improvement. If there are multiple matching authentications, why would "honour the first non-Accepted match, so a Blocked/Expired duplicate still blocks. Only when every match is Accepted does the token authorize" be the expected behavior? What would be the purpose of this overload--how do you get into this situation without it being a issue that needs to be fixed? The log that is emitted when this condition is triggered can be used to create an alert in a production environment, which could then prompt an investigation to avoid this scenario, which is the best outcome in my opinion.

It is possible to add a uniqueness constraint on the IdToken column alone when operating a 1.6 stack if this is a serious concern for a specific deployment, and I would be very interested in hearing more about why it is a concern. I have also considered adding an implicit type for 1.6 authorizations, such as forcing them all to be 'Central' or adding a 'OCPP16' type, but it's not clear to me that this is a realistic scenario that is worth complicating authorization management for.

biubiuzhang:
Thanks for the pushback — I think we agree on more than it looks, so let me answer the "how do you get into this situation" part directly.

**How the state arises.** `Authorizations` has no unique constraint on `idToken`, and the 2.0.1/2.1 paths in this same file query with `idToken.idToken` **and** `idToken.type`. Two rows with the same value and different `idTokenType` are therefore a supported shape for the 2.x paths, not corruption. On our system that is exactly what happened through ordinary use, months apart: the same identifier stored once as `ISO14443` (the RFID card) and once as `NoAuthorization` (the free-vend/autocharge authorization), both `Accepted`, nothing hand-edited.

The part that looks like a defect is the asymmetry: with those contents a station negotiated as 2.0.1 authorizes normally, and a station on 1.6 is rejected — a 1.6 `idTag` is a bare string with no token type, so the 1.6 path reuses the `length !== 1` guard with nothing to disambiguate with. Same database, same token, opposite answer depending on the protocol the station speaks.

**What we ran into.** The CSMS answered `Invalid` for a valid card, libocpp then correctly issued `stop_transaction(DeAuthorized)`, and the session was torn down shortly after `PrepareCharging`, so charging never started. It presents exactly like an unknown token, which is what made it slow to track down: today the no-match and the multiple-match case emit the same `Found invalid authorizations …` line, so you cannot alert specifically on the duplicate condition. Separating those two diagnostics is part of this PR, and I think it serves the alerting story you describe better than the current shared message.

**On your alternatives.** A unique constraint on `idToken` alone works for a pure-1.6 deployment and we can add it locally, but it does not generalise: for a CSMS serving 1.6 **and** 2.x it makes the legitimate typed-duplicate case unrepresentable, and it cannot be applied where such duplicates are already in use over 2.x. An implicit type for 1.6 authorizations is something we would happily implement if you prefer an explicit model — though it does not help rows that already exist, and even then you still need a deterministic rule when a 1.6 lookup matches several rows.

**What the change guarantees.** It can never be more permissive than the strictest matching row: the first non-`Accepted` match wins, so a `Blocked`/`Expired` duplicate still blocks, and only a unanimous `Accepted` authorizes. It cannot authorize anything the current code denied on the merits.

Happy to shape it however you like — for example raising the multiple-match log from `warn` to `error` and documenting it as a deployment misconfiguration worth alerting on, while still honouring the unanimous `Accepted` case. Or if you would rather go the implicit-1.6-type route, say so and I will rework it that way.

PR #852, opened 2026-08-12
fix(core): drop unmapped OCPP measurands instead of relabeling them as the energy register

vicky-ahmed:


packages/core/src/dal/layers/sequelize/mapper/2.0.1/MeterValueMapper.ts
```diff
@@ -199,7 +199,9 @@ export class MeterValueMapper {
       case OCPP2_0_1.MeasurandEnumType.Voltage:
         return 'Voltage';
       default:
-        return 'Energy.Active.Import.Register';
+        // Unrepresentable measurand: return undefined rather than mislabeling it as the energy
+        // register (which would corrupt the stored measurand and the transaction kWh).
+        return undefined;
```
thanaParis, line 204: This is a protocol violation. When the measurand is not provided, it must default to Energy.Active.Import.Register

We should account for that case, add the missing 2.1 measurands, and include a warning if there are unknown measurands.

packages/core/src/dal/layers/sequelize/mapper/2/MeterValueMapper.ts
```diff
@@ -32,17 +38,24 @@ export class MeterValueMapper {
       OCPP2_common_types.SampledValueType,
       ...OCPP2_common_types.SampledValueType[],
     ],
-  ): [SampledValue, ...SampledValue[]] {
+  ): SampledValue[] {
```
thanaParis: Why are we changing the type here? empty arrays should not be returned. it would make more sense to adjust to return undefined in that case, and this should only happen in the case where measurands are not known--not at all in the protocol. we must support all protocol measurands.

thanaParis (changes requested):

packages/core/src/dal/layers/sequelize/mapper/2.0.1/MeterValueMapper.ts
vicky-ahmed: The missing measurands is already handled at the top of the function -> if (!measurand) return 'Energy.Active.Import.Register';. But yeah it is true that we should add all the missing 2.1 measurands and also add a warning in the default.

vicky-ahmed: if we add Energy.Active.Import.Register in the default. it would incorrectly treat bad input as billing energy and could recreate the original bug

packages/core/src/dal/layers/sequelize/mapper/2/MeterValueMapper.ts
vicky-ahmed: Good call. Added support for all OCPP 2.1 measurands. So reverted it back

thanaParis (approved):

PR #855, opened 2026-08-12
returning state and message columns to ocpp messages

thanaParis:
For backwards compatibility; columns may be removed again later and have been marked as deprecated.


packages/core/src/modules/OcppRouter/src/module/webhook.dispatcher.ts
```diff
@@ -511,6 +519,23 @@ export class WebhookDispatcher {
     }
   }
 
+  /**
+   * Maps an RPC messageTypeId onto the deprecated `state` column so rows stay readable by
+   * consumers written before `type` existed. A message that could not be parsed far enough to
+   * have a role is Unknown, which is what the router used to pass here explicitly.
+   */
+  private _messageStateFromType(type?: MessageTypeId): MessageState {
```
devonNewsomDevelop, line 527: We should probably also make this as @ deprecated ?

devonNewsomDevelop (approved):
approve with suggestion

PR #857, opened 2026-08-13
fix(ocpi): convert energy charging-period dimensions to kWh

vicky-ahmed:


packages/ocpi-base/src/mapper/SessionMapper.ts
```diff
@@ -383,20 +384,20 @@ export class SessionMapper extends BaseTransactionMapper {
           break;
         case OCPP2_0_1.MeasurandEnumType.Energy_Active_Import_Register:
           if (!sampledValue.phase) {
```
devonNewsomDevelop, line 386: Curious, do we know what happens if there is a phase value? It seems like it just goes unhandled? Obviously this predates your changes, I'm just curious about this condition here since there doesn't seem to be handling for when there is a phase in this block.

packages/ocpi-base/src/mapper/SessionMapper.ts
```diff
@@ -383,20 +384,20 @@ export class SessionMapper extends BaseTransactionMapper {
...
-            });
-            const previousEnergyImport = this.getEnergyImportForMeterValue(previousMeterValue);
-            if (
-              previousEnergyImport !== undefined &&
-              !isNaN(Number(previousEnergyImport)) &&
-              !isNaN(Number(sampledValue.value))
-            ) {
+            // reading (commonly Wh) via the same conversion that feeds session.kwh.
+            const energyImportKwh = MeterValueUtils.normalizeToKwh(sampledValue);
+            if (energyImportKwh !== null && !isNaN(energyImportKwh)) {
```
devonNewsomDevelop: Hmm. I feel like the return type of this utility method possibly being null or NaN or a number is too broad for what it does. 

Would it make sense to have the method instead return -1 when it fails to convert and then have all checks using the method check for -1 instead?

That said, depending on how the numbers work, -1 could be a valid response. In that case, I suppose we would need to leave it the way it currently is.

packages/ocpi-base/src/mapper/SessionMapper.ts
```diff
@@ -383,20 +384,20 @@ export class SessionMapper extends BaseTransactionMapper {
...
               cdrDimensions.push({
-                type: CdrDimensionType.ENERGY,
-                volume: Number(sampledValue.value) - Number(previousEnergyImport),
+                type: CdrDimensionType.ENERGY_IMPORT,
+                volume: energyImportKwh,
               });
+              const previousEnergyImport = this.getEnergyImportForMeterValue(previousMeterValue);
+              if (previousEnergyImport !== undefined && !isNaN(previousEnergyImport)) {
```
devonNewsomDevelop: Can previousEnergyImport be null? Will that fail the check or does !NaN(null) cover that properly?

devonNewsomDevelop (approved):
Approved with questions that may lead to possible suggested changes depending on the answers.

packages/ocpi-base/src/mapper/SessionMapper.ts
vicky-ahmed: it's skipped on purpose. We only take the no phase (total) energy reading, the per phase are ignored because OCPI energy dimension is a single session total with no phase concept including them would triple count.

packages/ocpi-base/src/mapper/SessionMapper.ts
vicky-ahmed: The previousEnergyImport can't be null as the helper only ever returns a number or undefined. The !== undefined check is what keeps it safe. (!isNaN wouldn't catch null on its own, but null never reaches here anyways).

packages/ocpi-base/src/mapper/SessionMapper.ts
vicky-ahmed: Same here, the method never returns null it always gives back a number, or throws an error if the unit is unknown. NaN would only happen if the input value were already bad data. So -1 isn't needed, and -1 could be a real energy value so it might be unsafe to use.

packages/ocpi-base/src/mapper/SessionMapper.ts
devonNewsomDevelop: energyImportKwh can be null but previousEnergyImport cannot be null?

vicky-ahmed: Both cannot be null.

devonNewsomDevelop: `if (energyImportKwh !== null && !isNaN(energyImportKwh)) {`

Should we change this line to exclude the null check in that case? Perhaps it should be checking undefined instead?

vicky-ahmed: Good point. Done.

packages/base/src/util/MeterValueUtils.ts
```diff
@@ -214,10 +214,7 @@ export class MeterValueUtils {
 
       let sum = 0;
       for (const value of phaseNeutralValues) {
-        const normalizedValue = this.normalizeToKwh(value);
-        if (normalizedValue !== null) {
-          sum += normalizedValue;
-        }
+        sum += this.normalizeToKwh(value);
```
devonNewsomDevelop, line 217: I apologize if this comes up twice. I had added the comment on the commit but it doesn't seem to come through on the PR. 

I like where we are going with these changes, but since we have checks to ensure normalizeToKwh doesn't return NaN- that implies it can return NaN. If we return NaN here, this summation will throw an error

devonNewsomDevelop (changes requested):

packages/base/src/util/MeterValueUtils.ts
vicky-ahmed: I don’t think those checks were necessary here. normalizeToKwh returns a number or throws, so normalizedValue !== null was always true, and it wouldn’t guard against NaN anyway since NaN !== null is true. The actual NaN guard is !isNaN(energyImportKwh), which skips the value if it’s NaN. Makes sense?

devonNewsomDevelop: In that case, we should remove the check in sessionMapper right?

vicky-ahmed: Done

packages/ocpi-base/src/mapper/SessionMapper.ts
```diff
@@ -415,14 +415,17 @@ export class SessionMapper extends BaseTransactionMapper {
...
-    return (
-      meterValue?.sampledValue.find(
-        (sampledValue) =>
-          sampledValue.measurand === OCPP2_0_1.MeasurandEnumType.Energy_Active_Import_Register &&
-          !sampledValue.phase,
-      )?.value ?? undefined
+  private getEnergyImportForMeterValue(meterValue?: MeterValueDto): number | undefined {
+    const sampledValue = meterValue?.sampledValue.find(
+      (sampledValue) =>
+        sampledValue.measurand === OCPP2_0_1.MeasurandEnumType.Energy_Active_Import_Register &&
```
devonNewsomDevelop: Sorry to keep poking at small things, but I think we probably want an enum for this in Enums.ts instead of using OCPP2_0_1 specifically. The enum would want to combine the properties  of OCPP2_0_1 and OCPP2_1 if they differ

packages/base/src/util/MeterValueUtils.ts
```diff
@@ -226,21 +223,19 @@ export class MeterValueUtils {
...
-   * Convert a sampled value to kWh, applying unit multipliers.
+   * Convert a sampled energy value to kWh, applying the unit multiplier and Wh→kWh factor.
+   * A missing unit is treated as Wh. Throws for a genuinely unknown energy unit.
    * @param value A SampledValueType entry.
-   * @returns The converted value in kWh, or null if unit is missing.
+   * @returns The converted value in kWh.
    */
-  private static normalizeToKwh(value: SampledValue): number | null {
+  public static normalizeToKwh(value: SampledValue): number {
     let powerOfTen = value.unitOfMeasure?.multiplier ?? 0;
```
devonNewsomDevelop, line 239: It worries me that we have removed the check here. We are without any doubt sure that this will always be defined by every caller? The previous entry made sure to account for missing data but now we do not. If we are positive it cannot be missing an entry, that's fine. If it does seem like it can miss an entry, then we should revert the changes and adjust the flow as originally asked.

packages/ocpi-base/src/mapper/SessionMapper.ts
vicky-ahmed: Good catch. I replaced the OCPP2_0_1 specific enum with the shared MeasurandEnum from @citrineos/types.
It’s already version neutral and works for both OCPP2_0_1 and OCPP2_1  so there’s no need for a new combined enum.

packages/base/src/util/MeterValueUtils.ts
vicky-ahmed: No behavior was removed so no need to worry. We only removed the unnecessary null checks.

devonNewsomDevelop: @vicky-ahmed  

Ah sorry I missed where we were coalescing to 0. Good deal.

packages/ocpi-base/src/mapper/SessionMapper.ts
devonNewsomDevelop: Ah this is out of scope for you and you've already done enough so ima make a ticket for it but the way this enum is currently used across our project is not my favorite notation lol. Thanks for changing it to the common though!

devonNewsomDevelop (approved):
Thanks for the work.

PR #860, opened 2026-08-14
fix(operator-ui): make the start/stop transaction controls multi-EVSE aware

sjh37:
Three related changes to the Charging Station transaction controls, all rooted in the same assumption that a station has one EVSE. Two commits; happy to split if you would rather review them separately.

1. **Fix** — start and stop were mutually exclusive per station, so one busy EVSE hid Start for all of them.
2. **Feature** — a start/stop button on each EVSE row, preselecting that EVSE.
3. **Fix** — nothing refreshed the station after either command, so the buttons went stale.

---

## 1. Start and Stop are not mutually exclusive

The Actions column and the station detail card both choose between the remote start and remote stop buttons using a single station-wide test:

```tsx
// columns.tsx:198
const hasActiveTransactions = !isEmpty(row.original.transactions);
{!hasActiveTransactions && <StartTransactionButton station={row.original} />}
{hasActiveTransactions  && <StopTransactionButton  station={row.original} />}

// charging.station.detail.card.tsx:174 — same shape
const hasActiveTransactions = station.transactions && station.transactions.length > 0;
```

A Charging Station may expose several EVSEs, each able to run its own transaction. Treating the two commands as mutually exclusive means **one busy EVSE hides Start for the entire station**, leaving every other EVSE unstartable from the UI. On a two-EVSE station an operator can start exactly one transaction and then has no way to start the second — the button has become Stop.

What makes this a bug rather than a deliberate simplification is that everything *behind* those buttons already handles multiple EVSEs:

- the start modal renders an `EvseSelector` and sends the chosen `evseId`;
- the stop modal enumerates the active transactions and labels each with its EVSE so the operator picks which to stop (`remote-stop/2.0.1/index.tsx:104-106`).

Only the button visibility collapses the multi-EVSE case.

image

Replace the single boolean with independent `canStart` / `canStop`, in one helper shared by both call sites so they cannot drift:

```ts
export function getTransactionCommandAvailability(
  station: TransactionCommandAvailabilityInput,
): TransactionCommandAvailability {
  // The station queries already constrain this relation to `isActive: { _eq: true }`.
  const activeTransactionCount = station.transactions?.length ?? 0;
  const evseCount = station.evses?.length ?? 0;

  return {
    canStart: evseCount === 0 || activeTransactionCount < evseCount,
    canStop: activeTransactionCount > 0,
  };
}
```

Stop is offered whenever a transaction is running; Start until every EVSE has one. A caller that did not select the `evses` relation keeps the previous behaviour of offering Start, rather than hiding a command the station can still accept.

The input type is structural rather than a `Pick` of a station DTO because the two call sites pass `ChargingStationDto` and `ChargingStationDetailsDto`, which disagree on whether these relations are optional.

image

### Verification

Built and run against three two-EVSE OCPP 2.0.1 stations, which between them cover every branch:

| Station | EVSEs | Active transactions | Buttons shown |
|---|---|---|---|
| SIMON_EVEREST_2 | 2 | 1 | Start **and** Stop |
| SIMON_EVEREST_1 | 2 | 2 | Stop only |
| SIMON_EVEREST_3 | 2 | 1 | Start **and** Stop |

Before this change all three showed Stop only.

---

## 2. A start/stop button per EVSE row

With several EVSEs on a station, the station-level buttons still make the operator re-pick from a combobox the EVSE or transaction they were already looking at. Each row of the EVSE tab now carries its own button beside Add Connector: Stop when that EVSE has an active transaction, Start when it does not, each preselecting what the row already identifies.

Preselection is threaded through as optional props — `evse` for remote start, `transactionId` for remote stop — so the station-level buttons are unchanged. OCPP 1.6 has no EVSE concept and ignores the new prop; its modal continues to select a connector.

One detail worth flagging for review: the preselected start value is built exactly as `EvseSelector` builds its option values. Any other shape leaves the combobox with nothing to match, so it renders its placeholder over a form that does hold the value — it looks broken without being broken.

### Verification

On a station whose EVSE 1 was charging and EVSE 2 idle, the rows rendered Stop and Start respectively, and opening Start from the EVSE 2 row arrived with EVSE already set to 2.

---

## 3. Nothing refreshed the station after a command

Stopping one of two transactions left the operator looking at a stale Stop button for an EVSE that had already stopped.

`liveMode` is `'auto'` globally (`providers/index.tsx:98`), which suggests live updates are in play — but `@refinedev/hasura` needs a `meta.gqlSubscription` document to subscribe with, and no query in the app supplies one; every query passes `gqlQuery` only. **So no page in the operator UI live-updates at all.** The stale button is one visible symptom of that.

Both transaction modals now `invalidate({ invalidates: ['all'] })` after the command, matching the existing idiom in the network-profile modals.

**This is a mitigation, not a cure, and I would rather say so than imply otherwise.** `RequestStartTransaction` / `RequestStopTransaction` resolve on the charger's *acceptance*, not on the `TransactionEvent` that actually opens or closes the transaction, so an immediate refetch races it. The modals invalidate twice — once immediately, once after a short settle period. A page reload remains the guarantee. Giving the queries real subscription documents is the proper fix and is a much larger change; I have not attempted it here.

---

## Notes

No unit test accompanies this: `apps/operator-ui` currently has only Playwright e2e specs and no unit test runner, and introducing one felt out of scope. Happy to add a spec under `tests/e2e/specs/charging-stations/` if you would prefer.

The OCPP 1.6 modals are untouched, so 1.6 stations do not get the post-command invalidation either. Straightforward follow-up if wanted.

Item 1 shares its underlying assumption — one EVSE per Charging Station — with citrineos/citrineos-core#859, in a different layer.

### DIFF
--- a/apps/operator-ui/src/lib/client/components/modals/remote-start/2.0.1/index.tsx
+++ b/apps/operator-ui/src/lib/client/components/modals/remote-start/2.0.1/index.tsx
@@ -25,7 +25,7 @@ import { ResourceType } from '@lib/utils/access.types';
 import type { MessageConfirmation } from '@lib/utils/MessageConfirmation';
 import { triggerMessageAndHandleResponse } from '@lib/utils/messages.utils';
 import { closeModal } from '@lib/utils/store/modal.slice';
-import { useCustom, useInvalidate, useSelect, useTranslate } from '@refinedev/core';
+import { useCustom, useSelect, useTranslate } from '@refinedev/core';
 import { useForm } from '@refinedev/react-hook-form';
 import { plainToInstance } from 'class-transformer';
 import React, { useEffect, useMemo, useState } from 'react';
@@ -38,7 +38,6 @@ import { useTenantId } from '@lib/client/hooks/useTenantId';
 
 export interface OCPP2_0_1_RemoteStartProps {
   station: ChargingStationDto;
-  /** Preselects the EVSE, for callers that already know which one - e.g. the per-EVSE start button. */
   evse?: EvseDto;
 }
 
@@ -48,16 +47,11 @@ export type RemoteStartFormData = {
   evse?: string;
 };
 
-/** Grace period for the charger to report the transaction actually started or ended. */
-
 export const OCPP2_0_1_RemoteStart = ({ station, evse }: OCPP2_0_1_RemoteStartProps) => {
   const dispatch = useDispatch();
   const translate = useTranslate();
-  const invalidate = useInvalidate();
   const [loading, setLoading] = useState<boolean>(false);
 
-  // Must match the option value EvseSelector builds, or the combobox has nothing to match against
-  // and renders its placeholder despite the form holding a value.
   const preselectedEvse = useMemo(
     () => (evse?.id !== undefined ? buildEvseOptionValue(evse) : ''),
     [evse],
@@ -179,7 +173,6 @@ export const OCPP2_0_1_RemoteStart = ({ station, evse }: OCPP2_0_1_RemoteStartPr
     }).then(() => {
       form.reset();
       dispatch(closeModal());
-      invalidate({ invalidates: ['all'] });
     });
   };
 
--- a/apps/operator-ui/src/lib/client/components/modals/remote-start/2.0.1/index.tsx
+++ b/apps/operator-ui/src/lib/client/components/modals/remote-start/2.0.1/index.tsx
@@ -7,6 +7,7 @@ import {
   type AuthorizationDto,
   type ChargingStationDto,
   type ChargingStationSequenceDto,
+  type EvseDto,
   AuthorizationProps,
   BaseProps,
   ChargingStationSequenceTypeEnum,
@@ -23,7 +24,7 @@ import { ResourceType } from '@lib/utils/access.types';
 import type { MessageConfirmation } from '@lib/utils/MessageConfirmation';
 import { triggerMessageAndHandleResponse } from '@lib/utils/messages.utils';
 import { closeModal } from '@lib/utils/store/modal.slice';
-import { useCustom, useSelect, useTranslate } from '@refinedev/core';
+import { useCustom, useInvalidate, useSelect, useTranslate } from '@refinedev/core';
 import { useForm } from '@refinedev/react-hook-form';
 import { plainToInstance } from 'class-transformer';
 import React, { useEffect, useMemo, useState } from 'react';
@@ -36,6 +37,8 @@ import { useTenantId } from '@lib/client/hooks/useTenantId';
 
 export interface OCPP2_0_1_RemoteStartProps {
   station: ChargingStationDto;
+  /** Preselects the EVSE, for callers that already know which one - e.g. the per-EVSE start button. */
+  evse?: EvseDto;
 }
 
 export type RemoteStartFormData = {
@@ -44,10 +47,24 @@ export type RemoteStartFormData = {
   evse?: string;
 };
 
-export const OCPP2_0_1_RemoteStart = ({ station }: OCPP2_0_1_RemoteStartProps) => {
+/** Grace period for the charger to report the transaction actually started or ended. */
+const TRANSACTION_SETTLE_MS = 4000;
+
+export const OCPP2_0_1_RemoteStart = ({ station, evse }: OCPP2_0_1_RemoteStartProps) => {
   const dispatch = useDispatch();
   const translate = useTranslate();
+  const invalidate = useInvalidate();
   const [loading, setLoading] = useState<boolean>(false);
+
+  // Must match the option value EvseSelector builds, or the combobox has nothing to match against
+  // and renders its placeholder despite the form holding a value.
+  const preselectedEvse = useMemo(
+    () =>
+      evse?.id !== undefined
+        ? JSON.stringify({ id: evse.id, evseTypeId: evse.evseTypeId })
+        : '',
+    [evse],
+  );
 
   const tenantId = useTenantId();
 
@@ -70,7 +87,7 @@ export const OCPP2_0_1_RemoteStart = ({ station }: OCPP2_0_1_RemoteStartProps) =
     defaultValues: {
       remoteStartId: 0,
       authorization: '',
-      evse: '',
+      evse: preselectedEvse,
     },
   });
 
@@ -165,6 +182,11 @@ export const OCPP2_0_1_RemoteStart = ({ station }: OCPP2_0_1_RemoteStartProps) =
     }).then(() => {
       form.reset();
       dispatch(closeModal());
+      // Same reasoning as the remote stop modal: nothing refreshes the station's active
+      // transactions on its own, and RequestStartTransaction resolves on acceptance rather than on
+      // the charger's TransactionEvent(Started), so refetch once now and once after it settles.
+      invalidate({ invalidates: ['all'] });
+      setTimeout(() => invalidate({ invalidates: ['all'] }), TRANSACTION_SETTLE_MS);
     });
   };
 
--- a/apps/operator-ui/src/lib/client/components/modals/remote-start/2.0.1/index.tsx
+++ b/apps/operator-ui/src/lib/client/components/modals/remote-start/2.0.1/index.tsx
@@ -15,6 +15,7 @@ import {
 import { zodResolver } from '@hookform/resolvers/zod';
 import { Form } from '@lib/client/components/form';
 import { ComboboxFormField, FormField } from '@lib/client/components/form/field';
+import { buildEvseOptionValue } from '@lib/client/components/modals/shared/evse-selector/evse.option.value';
 import { EvseSelector } from '@lib/client/components/modals/shared/evse-selector/evse.selector';
 import { Input } from '@lib/client/components/ui/input';
 import { ChargingStationSequenceClass } from '@lib/cls/charging.station.sequence.dto';
@@ -48,7 +49,6 @@ export type RemoteStartFormData = {
 };
 
 /** Grace period for the charger to report the transaction actually started or ended. */
-const TRANSACTION_SETTLE_MS = 4000;
 
 export const OCPP2_0_1_RemoteStart = ({ station, evse }: OCPP2_0_1_RemoteStartProps) => {
   const dispatch = useDispatch();
@@ -59,10 +59,7 @@ export const OCPP2_0_1_RemoteStart = ({ station, evse }: OCPP2_0_1_RemoteStartPr
   // Must match the option value EvseSelector builds, or the combobox has nothing to match against
   // and renders its placeholder despite the form holding a value.
   const preselectedEvse = useMemo(
-    () =>
-      evse?.id !== undefined
-        ? JSON.stringify({ id: evse.id, evseTypeId: evse.evseTypeId })
-        : '',
+    () => (evse?.id !== undefined ? buildEvseOptionValue(evse) : ''),
     [evse],
   );
 
@@ -182,11 +179,7 @@ export const OCPP2_0_1_RemoteStart = ({ station, evse }: OCPP2_0_1_RemoteStartPr
     }).then(() => {
       form.reset();
       dispatch(closeModal());
-      // Same reasoning as the remote stop modal: nothing refreshes the station's active
-      // transactions on its own, and RequestStartTransaction resolves on acceptance rather than on
-      // the charger's TransactionEvent(Started), so refetch once now and once after it settles.
       invalidate({ invalidates: ['all'] });
-      setTimeout(() => invalidate({ invalidates: ['all'] }), TRANSACTION_SETTLE_MS);
     });
   };
 
--- a/apps/operator-ui/src/lib/client/components/modals/remote-start/remote.start.modal.tsx
+++ b/apps/operator-ui/src/lib/client/components/modals/remote-start/remote.start.modal.tsx
@@ -2,7 +2,7 @@
 //
 // SPDX-License-Identifier: Apache-2.0
 
-import { type ChargingStationDto, OCPPVersion } from '@citrineos/types';
+import { type ChargingStationDto, type EvseDto, OCPPVersion } from '@citrineos/types';
 import { ChargingStationClass } from '@lib/cls/charging.station.dto';
 import { plainToInstance } from 'class-transformer';
 import { useMemo } from 'react';
@@ -12,9 +12,17 @@ import { OCPP2_0_1_RemoteStart } from './2.0.1';
 
 export interface RemoteStartTransactionModalProps {
   station: any;
+  /**
+   * Preselects the EVSE. OCPP 1.6 has no EVSE concept and ignores this - its modal selects a
+   * connector instead.
+   */
+  evse?: EvseDto;
 }
 
-export const RemoteStartTransactionModal = ({ station }: RemoteStartTransactionModalProps) => {
+export const RemoteStartTransactionModal = ({
+  station,
+  evse,
+}: RemoteStartTransactionModalProps) => {
   const translate = useTranslate();
   const parsedStation: ChargingStationDto = useMemo(
     () => plainToInstance(ChargingStationClass, station),
@@ -28,7 +36,7 @@ export const RemoteStartTransactionModal = ({ station }: RemoteStartTransactionM
         return <OCPP1_6_RemoteStart station={parsedStation} />;
       case OCPPVersion.OCPP2_0_1:
       case OCPPVersion.OCPP2_1:
-        return <OCPP2_0_1_RemoteStart station={parsedStation} />;
+        return <OCPP2_0_1_RemoteStart station={parsedStation} evse={evse} />;
       default:
         return (
           <div>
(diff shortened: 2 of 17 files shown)


apps/operator-ui/src/lib/client/components/modals/remote-start/2.0.1/index.tsx
```diff
@@ -165,6 +180,11 @@ export const OCPP2_0_1_RemoteStart = ({ station }: OCPP2_0_1_RemoteStartProps) =
     }).then(() => {
       form.reset();
       dispatch(closeModal());
+      // Same reasoning as the remote stop modal: nothing refreshes the station's active
+      // transactions on its own, and RequestStartTransaction resolves on acceptance rather than on
+      // the charger's TransactionEvent(Started), so refetch once now and once after it settles.
+      invalidate({ invalidates: ['all'] });
+      setTimeout(() => invalidate({ invalidates: ['all'] }), TRANSACTION_SETTLE_MS);
```
devonNewsomDevelop: We need to use Hasura subscriptions or queries instead of a guessing timeout. This is too much of a work-around bandaid to be permissible.

apps/operator-ui/src/lib/client/components/modals/remote-stop/2.0.1/index.tsx
```diff
@@ -68,15 +74,31 @@ export const OCPP2_0_1_RemoteStop = ({ station }: OCPP2_0_1_RemoteStopProps) =>
...
+      // refreshes them on its own: liveMode is inert here because the Hasura live provider has no
+      // gqlSubscription to subscribe with. Without this the operator is left looking at Stop for an
+      // EVSE that has already stopped.
+      //
+      // RequestStopTransaction only resolves to the charger's *acceptance*; the transaction is not
+      // closed until it sends TransactionEvent(Ended) a moment later. So refetch again shortly
+      // after - the first pass keeps the UI honest if the stop is rejected, the second picks up the
+      // actual end. This is a heuristic, and a page reload remains the guarantee.
+      invalidate({ invalidates: ['all'] });
+      setTimeout(() => invalidate({ invalidates: ['all'] }), TRANSACTION_SETTLE_MS);
```
devonNewsomDevelop: Same here. We cannot have workarounds of this nature in the project- at least not going forward

apps/operator-ui/src/lib/client/pages/charging-stations/detail/evses/evses.list.tsx
```diff
@@ -246,6 +299,24 @@ export const EVSESList: React.FC<EVSESListProps> = ({ id }) => {
                           >
                             {translate('ChargingStations.connectors.addConnector')}
                           </Button>
+                          {activeTransaction ? (
```
devonNewsomDevelop, line 299: We need the CanAccess wrappers around these to ensure the user has the proper permissions to start and stop transactions

devonNewsomDevelop (changes requested):
Thank you for the work. Some changes requested for this one, the timeout stuff may take some effort to address.

sjh37:
@devonNewsomDevelop Both addressed - thank you for pushing back on the timeout, you were right and the workaround was hiding a wrong assumption.

**Subscriptions instead of the timeout.** The timed refetch is gone from both modals, along with `TRANSACTION_SETTLE_MS`. Instead, `useOne` for the station detail now carries `liveMode: 'auto'`.

That works because `@refinedev/hasura`'s `generateUseOneSubscription` builds its subscription from `meta.gqlQuery` (`gqlOperation = meta?.gqlQuery ?? meta?.gqlMutation`, then `getOperationFields(gqlOperation)`), and `CHARGING_STATIONS_GET_QUERY` already selects `transactions: Transactions(where: { isActive: { _eq: true } })`. So the subscription covers the station's active transactions, and the start/stop controls follow the charger without the page guessing.

Worth flagging: an earlier comment of mine in the remote-stop modal asserted that `liveMode` was inert here because the live provider had no `gqlSubscription` to subscribe with. **That was wrong** - it reads `gqlQuery`. I only checked properly because you questioned the timeout, so the bandaid was covering a mistaken premise rather than a real gap. The comment has gone with the code.

I kept the single immediate `invalidate({ invalidates: [all] })`. That is not a guess - it is for the views that are not subscribed, such as the stations list. Happy to drop it too if you would rather those went live as well, which would be a slightly wider change.

**CanAccess.** Start and stop are each wrapped now, using `ActionType.COMMAND` with `commandType: CommandType.START_TRANSACTION` / `STOP_TRANSACTION` and the station id, matching the other-commands modals. The shipped rules resolve `COMMAND` by action alone, but passing `commandType` means a deployment with finer rules can distinguish starting from stopping, which I think is what you were after.

Verified: `tsc --noEmit` clean on operator-ui, `eslint` 0 errors (3 pre-existing react-hooks warnings elsewhere in the app), and the full workspace suite green at 2142 passed / 4 skipped. I have not been able to exercise the subscription against a live Hasura from here, so a sanity check of the start/stop controls updating without a reload would be worth someone's five minutes before merge.

apps/operator-ui/src/lib/client/components/modals/remote-start/2.0.1/index.tsx
sjh37: The timeout is gone. `TRANSACTION_SETTLE_MS` and both `setTimeout` calls are removed, and the modal now relies on the station detail card's `useOne` running with `liveMode: 'auto'`, so the controls follow what the charger actually reports.

Worth flagging that my original defence of the timeout was wrong. I had claimed `liveMode` was inert without an explicit `gqlSubscription` in `meta`, but `generateUseOneSubscription` reads `meta.gqlQuery ?? meta.gqlMutation` - so the subscription was available the whole time and the timeout was papering over a mistaken premise rather than a real gap.

apps/operator-ui/src/lib/client/components/modals/remote-stop/2.0.1/index.tsx
sjh37: Same change here: `TRANSACTION_SETTLE_MS` and the `setTimeout` are removed, and the stop modal follows the same live subscription. There is no arbitrary delay left in either modal - `grep -c 'TRANSACTION_SETTLE_MS\|setTimeout'` over the branch returns 0.

apps/operator-ui/src/lib/client/pages/charging-stations/detail/evses/evses.list.tsx
sjh37: Added. Both controls are now wrapped in `CanAccess` with `action={ActionType.COMMAND}` and the matching `commandType` in `params` - `CommandType.START_TRANSACTION` for start and `CommandType.STOP_TRANSACTION` for stop - so each button is gated on the permission for the command it actually issues rather than on the pair.

apps/operator-ui/src/lib/client/components/modals/remote-stop/2.0.1/index.tsx
```diff
@@ -68,15 +73,21 @@ export const OCPP2_0_1_RemoteStop = ({ station }: OCPP2_0_1_RemoteStopProps) =>
       setLoading,
     }).then(() => {
       dispatch(closeModal());
+      invalidate({ invalidates: ['all'] });
```
devonNewsomDevelop: This isn't going to work the way it's intended. This is far out of scope and will work on any and all queries in the app. The refresh needs more thought.

apps/operator-ui/src/lib/client/components/modals/remote-stop/2.0.1/index.tsx
```diff
@@ -14,21 +14,26 @@ import { closeModal } from '@lib/utils/store/modal.slice';
...
+  /** Preselects the transaction, for callers that already know which one - e.g. a per-EVSE button. */
+  transactionId?: string;
 }
 
 type RemoteStopFormData = {
   transactionId: string;
 };
 
-export const OCPP2_0_1_RemoteStop = ({ station }: OCPP2_0_1_RemoteStopProps) => {
+/** Grace period for the charger to report the transaction actually started or ended. */
```
devonNewsomDevelop: Dangling comment

apps/operator-ui/src/lib/client/components/modals/remote-start/remote.start.modal.tsx
```diff
@@ -12,9 +12,17 @@ import { OCPP2_0_1_RemoteStart } from './2.0.1';
 
 export interface RemoteStartTransactionModalProps {
   station: any;
+  /**
+   * Preselects the EVSE. OCPP 1.6 has no EVSE concept and ignores this - its modal selects a
```
devonNewsomDevelop: Unneeded information.

apps/operator-ui/src/lib/client/components/modals/shared/evse-selector/evse.option.value.ts
```diff
@@ -0,0 +1,20 @@
+// SPDX-FileCopyrightText: 2025 Contributors to the CitrineOS Project
+//
+// SPDX-License-Identifier: Apache-2.0
+
+import type { EvseDto } from '@citrineos/types';
+
+/**
+ * Builds the value {@link EvseSelector} puts on each combobox option, and the value a caller must
```
devonNewsomDevelop: This is a massive wall of text for a 2 line lambda

apps/operator-ui/src/lib/client/pages/charging-stations/detail/evses/evses.list.tsx
```diff
@@ -26,6 +33,10 @@ interface EVSESListProps {
 
 export const evsesFormUpsertGrid = 'grid grid-cols-2 xs:grid-cols-1 gap-6';
 
+// The station detail query omits `station` and `location` from each transaction, so this is
+// narrower than TransactionDto. Derive it rather than restate it.
```
devonNewsomDevelop: Unneeded comment.

apps/operator-ui/src/lib/client/pages/charging-stations/detail/evses/evses.list.tsx
```diff
@@ -46,10 +57,52 @@ export const EVSESList: React.FC<EVSESListProps> = ({ id }) => {
   });
 
   const station = React.useMemo(() => {
-    const station = { ...data?.data } as ChargingStationDto;
+    const station = { ...data?.data } as ChargingStationDetailsDto;
     return station;
   }, [data?.data]);
 
+  // Transaction.evseId is the Evses primary key, the same value keying these rows. The query
+  // already restricts the relation to active transactions, and an EVSE runs at most one.
```
devonNewsomDevelop: Unneeded comment

apps/operator-ui/src/lib/client/pages/charging-stations/detail/charging.station.detail.card.tsx
```diff
@@ -73,6 +74,9 @@ export const ChargingStationDetailCard = ({
   } = useOne<ChargingStationDetailsDto>({
     resource: ResourceType.CHARGING_STATIONS,
     id,
+    // The query selects the station's active transactions, so subscribing keeps the start/stop
+    // controls in step with the charger without the page having to poll or guess.
```
devonNewsomDevelop: Unneeded comment.

apps/operator-ui/src/lib/client/pages/charging-stations/transaction.command.availability.ts
```diff
@@ -0,0 +1,41 @@
+// SPDX-FileCopyrightText: 2025 Contributors to the CitrineOS Project
+//
+// SPDX-License-Identifier: Apache-2.0
+
+/**
+ * Which of the remote start / remote stop commands a Charging Station can currently accept.
+ *
+ * A Charging Station may expose several EVSEs, each able to run its own transaction, so the two
```
devonNewsomDevelop: Too much information.

apps/operator-ui/src/lib/client/pages/charging-stations/transaction.command.availability.ts
```diff
@@ -0,0 +1,41 @@
...
+ * both started and stopped. Deciding on the presence of any active transaction hides Start for the
+ * whole station as soon as a single EVSE is busy.
+ */
+export interface TransactionCommandAvailability {
+  canStart: boolean;
+  canStop: boolean;
+}
+
+/**
+ * Structural rather than a Pick of a station DTO: the callers pass ChargingStationDto and
```
devonNewsomDevelop: Unneeded comment

apps/operator-ui/src/lib/client/pages/charging-stations/transaction.command.availability.ts
```diff
@@ -0,0 +1,41 @@
...
+ */
+export interface TransactionCommandAvailabilityInput {
+  transactions?: unknown[] | null;
+  evses?: unknown[] | null;
+}
+
+export function getTransactionCommandAvailability(
+  station: TransactionCommandAvailabilityInput,
+): TransactionCommandAvailability {
+  // The station queries already constrain this relation to `isActive: { _eq: true }`.
```
devonNewsomDevelop: Unneeded comment

devonNewsomDevelop (changes requested):

apps/operator-ui/src/lib/client/components/modals/remote-stop/2.0.1/index.tsx
sjh37: Agreed, and it is gone rather than narrowed. The station detail card's `useOne` already runs with `liveMode: 'auto'`, which is what the earlier round settled on in place of the timeout, so the modal has nothing left to refresh. Both `useInvalidate` calls and their imports are removed.

apps/operator-ui/src/lib/client/components/modals/remote-stop/2.0.1/index.tsx
sjh37: Removed.

apps/operator-ui/src/lib/client/components/modals/remote-start/remote.start.modal.tsx
sjh37: Removed.

apps/operator-ui/src/lib/client/components/modals/shared/evse-selector/evse.option.value.ts
sjh37: Removed.

apps/operator-ui/src/lib/client/pages/charging-stations/detail/evses/evses.list.tsx
sjh37: Removed.

apps/operator-ui/src/lib/client/pages/charging-stations/detail/evses/evses.list.tsx
sjh37: Removed.

apps/operator-ui/src/lib/client/pages/charging-stations/detail/charging.station.detail.card.tsx
sjh37: Removed.

apps/operator-ui/src/lib/client/pages/charging-stations/transaction.command.availability.ts
sjh37: Removed.

apps/operator-ui/src/lib/client/pages/charging-stations/transaction.command.availability.ts
sjh37: Removed.

apps/operator-ui/src/lib/client/pages/charging-stations/transaction.command.availability.ts
sjh37: Removed.

apps/operator-ui/src/lib/client/components/modals/shared/evse-selector/evse.option.value.ts
sjh37: Correction: my "Removed." above was wrong - the block was still there when I said that. My sweep skipped it because it contains a `{@link ...}` reference and I had told the script to leave anything with a JSDoc tag alone. It is gone now, in the head commit: `buildEvseOptionValue` is the two-line lambda and nothing else. Apologies for the noise.

sjh37:
The `operator-ui-e2e-everest` failure here was not this change: `patchEverestNetworkProfile` runs before any spec and lost a race with the EVerest manager for the device-model SQLite write lock, so the lane died with `SQLite patch failed (code 1): Runtime error near line 1: database is locked (5)` and reported it against whichever spec ran first. It hit two unrelated PRs on two different specs within a minute of each other.

Fixed in #943 - `PRAGMA busy_timeout` so the write waits for the lock, plus a bounded retry. I have cherry-picked that commit here so this PR can be reviewed green; it will drop out of the diff when #943 merges.

thanaParis (approved):

PR #861, opened 2026-08-14
fix(deploy): docker build fails on EXPOSE with undefined PORT variable

mrnonaki:
## Problem
`apps/ocpp-server/deploy.Dockerfile` ends with `EXPOSE ${PORT}`, but `PORT` is never declared as an `ARG` (or given a default). The variable expands to an empty string, so a plain

```
docker build -f apps/ocpp-server/deploy.Dockerfile .
```

always fails with:

```
ERROR: Dockerfile parse error: EXPOSE requires at least one argument
```

## Fix
Replace the undefined variable with the concrete ports the server actually listens on — 8080 (REST API) and 8081/8082/8443/8444 (websocket servers, security profiles 0–3) — matching the port mappings already published in `docker-compose.yml`. `EXPOSE` is documentation metadata, so listing the real set is strictly more useful than a single parameterized port.

## Verification
- `docker build -f apps/ocpp-server/deploy.Dockerfile .` fails on `next` before this change and succeeds with it (also verified with podman/buildah, which fails identically on the unset variable).

### DIFF
--- a/apps/ocpp-server/deploy.Dockerfile
+++ b/apps/ocpp-server/deploy.Dockerfile
@@ -25,6 +25,7 @@ WORKDIR /usr/local/apps/citrineos
 
 RUN chmod +x /usr/local/apps/citrineos/apps/ocpp-server/entrypoint.sh
 
-EXPOSE ${PORT}
+# REST API, websocket servers (security profiles 0-3)
+EXPOSE 8080 8081 8082 8443 8444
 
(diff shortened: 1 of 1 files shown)


yuvenus (approved):

thanaParis (changes requested):
This is a no-op for normal docker builds rather than a crash, so I didn't catch this before--are you using Podman?

In any case, I agree it should be fixed. Hard-coding the ports into the image doesn't make sense given they're configurable though. I think the EXPOSE line can be safely removed seeing as it has done nothing for quite a while. Please just get rid of it.

PR #862, opened 2026-08-14
fix(core): add timeout to realtime-auth fetch so a hung endpoint cannot stall Authorize

mrnonaki:
## Problem
`RealTimeAuthorizer.authorize()` calls `fetch(authorization.realTimeAuthUrl, …)` without an abort signal. If the configured realtime-auth endpoint hangs (accepts the TCP connection but never responds), the whole authorization flow blocks indefinitely — and since this runs inside `Authorize` / `TransactionEvent` handling, the charging station's OCPP call stalls well past its own call timeout. One unresponsive external endpoint can effectively wedge authorization for its tokens with no recovery until the socket dies.

## Fix
Add `signal: AbortSignal.timeout(30_000)` to the request. On timeout the abort error flows into the existing `catch` block, which already implements the intended degraded behavior: log the failure and apply the `AllowedOffline` whitelist fallback. No behavior change for endpoints that respond within 30s.

(The existing per-authorization `realTimeAuthTimeout` field governs the last-attempt cache window, not the HTTP request itself, so a module-level constant is used rather than overloading that field.)

## Verification
- `pnpm --filter "@citrineos/core" build` passes.
- The timeout path needs no new handling: `AbortSignal.timeout` rejects the `fetch` promise with a `TimeoutError`, which lands in the function's existing `catch (error)` block (the same path as DNS/connection failures today), so the `AllowedOffline` fallback semantics are unchanged.

### DIFF
--- a/packages/core/src/util/authorizer/RealTimeAuthorizer.ts
+++ b/packages/core/src/util/authorizer/RealTimeAuthorizer.ts
@@ -35,10 +35,6 @@ export interface RealTimeAuthorizationResponse {
   };
 }
 
-// The realtime-auth request runs inside Authorize/TransactionEvent handling, so a
-// hung endpoint would stall the charger's OCPP call past its own timeout.
-const REAL_TIME_AUTH_FETCH_TIMEOUT_MS = 30_000;
-
 export class RealTimeAuthorizer implements IAuthorizer {
   private _locationRepository: ILocationRepository;
   private _config: SystemConfig;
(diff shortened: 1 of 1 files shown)


packages/core/src/util/authorizer/RealTimeAuthorizer.ts
```diff
@@ -35,6 +35,10 @@ export interface RealTimeAuthorizationResponse {
   };
 }
 
+// The realtime-auth request runs inside Authorize/TransactionEvent handling, so a
+// hung endpoint would stall the charger's OCPP call past its own timeout.
+const REAL_TIME_AUTH_FETCH_TIMEOUT_MS = 30_000;
```
yuvenus: Rather than hardcode to 30 seconds, why not use the timeout that's configured by the authorization (and fallsback to what's in the config)? For example, down on line 128/129, it already does this:

```
const timeout = authorization.realTimeAuthTimeout ?? this._config.realTimeAuthDefaultTimeoutSeconds;
```

yuvenus (changes requested):

packages/core/src/util/authorizer/RealTimeAuthorizer.ts
mrnonaki: Good call — updated in 44945078. The fetch now reuses the same pattern as the last-attempt check: `authorization.realTimeAuthTimeout ?? this._config.realTimeAuthDefaultTimeoutSeconds`, converted to ms for `AbortSignal.timeout`. The hardcoded constant is gone.

thanaParis (changes requested):
Looks good! There is a broken test though--I ran it twice under the suspicion that it might just be flaky but it broke twice. It will need to be fixed.

thanaParis:
Closed in favor of https://github.com/citrineos/citrineos-core/pull/1028

PR #867, opened 2026-08-14
Feature: OCPPMessages Partition

lydiazcheng:
### Context
This PR is to implement partition on OCPPMessages table. Details can be found in changes in apps/ocpp-server/README.md

### Tests
1. build and run citrine in docker. The partition is done including: 1 week back, 1 week current and 1 week ahead. Screenshot 2026-08-14 at 2 56 56 PM
2. old data are kept in table OCPPMessages_old. Screenshot 2026-08-14 at 3 40 48 PM
3. test the `apps/ocpp-server/src/scripts/provision-partitions.ts` used in `entrypoint.sh` by set future week to 2. rebuild and start the service, an additional weekly partition table was created. 
4. the partition table contains the correct data, old data are not copied.Screenshot 2026-08-14 at 2 57 10 PM
5. Send a request and get response. The new ocpp message pair is stored in the expected partition table. Screenshot 2026-08-14 at 3 52 37 PM Screenshot 2026-08-14 at 3 39 13 PM


apps/ocpp-server/entrypoint.sh
```diff
@@ -17,6 +17,9 @@ else
     (cd "$SCRIPT_DIR" && pnpm run db:migrate)
 fi
 
+# "OCPPMessages" is partitioned by week; this runs every start.
+node "$SCRIPT_DIR/dist/scripts/provision-partitions.js"
```
thanaParis: This should only run if migrations are enabled. if migrations are disabled ('none') this is a harmful script and should not run

packages/dal/src/models/ocpp-message.ts
```diff
@@ -84,7 +88,9 @@ export class OCPPMessage extends Model implements OCPPMessageDto {
   @BelongsTo(() => ChargingStation, 'stationId')
   declare chargingStation?: ChargingStationDto;
 
-  @ForeignKey(() => OCPPMessage)
+  // "OCPPMessages" is range partitioned on "createdAt", so its only unique key is (id, "createdAt") and a
```
thanaParis: please remove this and other comments that are for explaining these changes rather than helping with legitimately obscure code

thanaParis (changes requested):

apps/ocpp-server/entrypoint.sh
lydiazcheng: Thank you for the review, Thana! I updated codes.

thanaParis (approved):

PR #868, opened 2026-08-17
fix(base): don't reset a session's totalKwh when a meter batch has no energy reading

sjh37:
## Problem

`MeterValueUtils.getTotalKwh` returns the total kWh for a transaction, and every caller writes that return value straight back to `Transaction.totalKwh`, unconditionally:

- `SequelizeTransactionEventRepository.createOrUpdateTransactionByTransactionEventAndStationId` (`TransactionEvent.ts:344`, `:356`)
- `SequelizeTransactionEventRepository.updateTransactionByMeterValues` (`TransactionEvent.ts:638`, `:647`)
- `TransactionService.recalculateTotalKwh` (`TransactionService.ts:85-92`)

When the incoming batch carried no usable energy reading the function returned `0`, so those writes **wiped the accumulated total of a live session** instead of leaving it alone.

Two routine cases hit this:

1. **Clock-aligned OCPP 1.6 MeterValues.** `Sample.Clock` is not in `validContexts`, so `filterValidMeterValues` removes the whole batch and the early return fires.
2. **A batch with no energy measurand.** A `TransactionEvent(Updated)` carrying only `SoC`, or a `MeterValues` carrying only `Current.Import`/`Voltage`, finds nothing in the register, interval or net maps and falls through to the final `return 0`.

Both are common enough that a long session can be zeroed repeatedly, and `totalKwh` is what billing (`CostCalculator.calculateTotalCost`) and the operator UI read.

Reproduced against the built package with a running total of 12.5 kWh:

```
periodic register : 12.5      <- correct
Sample.Clock only : 0         <- wiped
SoC-only batch    : 0         <- wiped
Trigger context   : 0         <- wiped
```

## Fix

Return `currentTotal` rather than `0` on both no-usable-reading paths. Callers already pass `transaction.totalKwh ?? 0`, so an untouched session is left exactly as it was.

The two existing `returns 0 …` tests still hold unchanged, because both pass a `currentTotal` of `0`.

## Tests

Three new cases under `getTotalKwh > Batches with no usable energy reading`, covering the clock-aligned batch, the SoC-only batch and the empty batch. They fail on `next` with `expected +0 to be 12.5` and pass with the fix.

## Verification

- **after:** 1751 passed (+3 new), 2 failed, 4 skipped

The 2 failures are `packages/core/test/dal/e2e/security-event.e2e.test.ts`, which fail identically on unmodified `next` — pre-existing and unrelated.

### DIFF
--- a/apps/operator-ui/tests/e2e/fixtures/everest.ts
+++ b/apps/operator-ui/tests/e2e/fixtures/everest.ts
@@ -48,6 +48,8 @@ const POLL_INTERVAL_MS = 2_000;
 // few seconds after `compose up`; this bounds the wait so a manager that never
 // boots fails fast into awaitStationOnline's diagnostics rather than hanging.
 const DB_READY_TIMEOUT_MS = 60_000;
+const SQLITE_BUSY_TIMEOUT_MS = 30_000;
+const SQLITE_PATCH_ATTEMPTS = 3;
 
 export interface EverestHandle {
   readonly ocppConnectionName: string;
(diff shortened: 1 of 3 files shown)


sjh37:
@thanaParis — I don't have triage access on this repo so I can't request a reviewer myself. Could you assign someone, or point me at who should take these?

Seven bug fixes open against `next`. Each has a failing test committed first and the full suite run before and after.

### Multi-EVSE (open longest, still unreviewed)

- **#859** `fix(core)` — in OCPP 2.0.1 a `connectorId` is scoped to its EVSE, so a two-EVSE station reports `connectorId: 1` twice. `processStatusNotification` writes that EVSE-scoped value straight into the station-wide `Connector.connectorId` column, so both notifications collapse onto one row: every other EVSE ends up owning no connectors and sits at `Unknown` indefinitely. Resolves through `evseTypeConnectorId` instead — the same key the transaction path already uses.
- **#860** `fix(operator-ui)` — the start/stop controls assume one EVSE per station. Both call sites pick between Start and Stop using a single station-wide `hasActiveTransactions`, so one busy EVSE hides Start for the whole station and the remaining EVSEs can't be started from the UI at all. Splits that into independent `canStart`/`canStop` in a shared helper, adds a control per EVSE row, and refreshes the station after either command.

### Transaction totals and cost

- **#868** `fix(base)` — a meter batch with no usable energy reading resets a live session's `totalKwh` to 0. Every caller writes the return value straight back to `Transaction.totalKwh`, so a clock-aligned 1.6 `MeterValues` or an SoC-only `TransactionEvent(Updated)` wipes the running total.
- **#869** `fix(core)` — OCPP 2.x `MeterValues` are never linked to their transaction. `MeterValuesRequestOcpp2Handler` declares `_sendCostUpdatedOnMeterValue` but the constructor never assigns it (and doesn't take `config`), so the guard is permanently false and every reading is stored orphaned.
- **#870** `fix(core)` — `TransactionEvent(Ended)` carrying an idToken skips all cost calculation. The handler branched on whether an authorization response existed, and everything cost-related lived only in the `else`; a card-stopped session takes the other branch.
- **#871** `fix(base)` — an unrecognised meter unit throws from inside `normalizeToKwh` and takes out handling of the whole message.

### Protocol

- **#872** `fix(core)` — boot status and the charger-action blacklist are written under the bare station name, but `MessageRouterImpl` reads them under `createIdentifier(tenantId, ocppConnectionName)`. The keys never match, so B03.FR.08 isn't enforced and the blacklist has no effect.

**#868 is the cheapest place to start** — 9 lines of source.

One ordering note: **#872 should land after #865**, as they touch the same OCPP 1.6 un-blacklist loop. I deliberately left that line alone to avoid conflicting with your fix.

yuvenus (approved):

sjh37:
The `operator-ui-e2e-everest` failure here was not this change: `patchEverestNetworkProfile` runs before any spec and lost a race with the EVerest manager for the device-model SQLite write lock, so the lane died with `SQLite patch failed (code 1): Runtime error near line 1: database is locked (5)` and reported it against whichever spec ran first. It hit two unrelated PRs on two different specs within a minute of each other.

Fixed in #943 - `PRAGMA busy_timeout` so the write waits for the lock, plus a bounded retry. I have cherry-picked that commit here so this PR can be reviewed green; it will drop out of the diff when #943 merges.

PR #869, opened 2026-08-17
fix(core): attach OCPP 2.x MeterValues to their transaction

sjh37:
## Problem

`MeterValuesRequestOcpp2Handler` declares `_sendCostUpdatedOnMeterValue` (`:33`) but **never assigns it** — the constructor does not take `config` at all. Compare `TransactionEventRequestOcpp2Handler:97`, which reads the same setting. Because the field's declared type includes `undefined`, `strictPropertyInitialization` is satisfied and the compiler stays quiet.

So the guard at `:75`

```ts
if (this._sendCostUpdatedOnMeterValue && evseId !== 0) {
```

is permanently false, and **every** OCPP 2.0.1/2.1 `MeterValuesRequest` falls through to `:107`:

```ts
await this._transactionService.createMeterValues(tenantId, meterValues);
```

with no `transactionDatabaseId`, `transactionId` or `tariffId`. The consequences:

- meter values are persisted **orphaned** — never linked to the transaction that produced them;
- `recalculateTotalKwh` never runs from the `MeterValues` path, so `Transaction.totalKwh` only ever moves when readings arrive embedded in a `TransactionEvent`;
- `calculateCostAndNotify` never fires.

This is the default behaviour: both shipped configs (`apps/ocpp-server/src/config/envs/{local,docker}.ts:155/157`) set `costUpdatedInterval`, and the schema refinement at `packages/types/src/config/types.ts:529` forbids setting both.

### Second layer

Assigning the field alone is not sufficient. `TransactionService.createMeterValues:285-298` only threads the transaction ids through when the batch contains a `Sample.Periodic` reading:

```ts
const hasPeriodic = meterValue.sampledValue?.some((s) => s.context === Sample_Periodic);
if (transactionDbId && hasPeriodic) { /* linked */ } else { /* orphaned */ }
```

`Sample.Clock`, `Transaction.Begin`, `Transaction.End` and `Trigger` batches are all silently orphaned. `createMeterValues` has exactly one caller — this handler — so the context test serves no purpose beyond dropping the link.

## Fix

1. Take `config` in the constructor and assign `_sendCostUpdatedOnMeterValue`.
2. Gate only the `CostUpdated` call on that setting. Associating a reading with its transaction and keeping `totalKwh` current are not optional behaviours, and gating them on a cost-notification flag conflates two unrelated concerns. They now happen whenever the request names a real EVSE (`evseId !== 0`) that has an active transaction.
3. Drop the `hasPeriodic` condition in `TransactionService.createMeterValues`.

`config` is already resolvable in this Awilix scope — `TransactionEventRequestOcpp2Handler` sits in the same `TRANSACTIONS_HANDLERS` list and takes it.

## Tests

New `packages/core/test/handlers/requests/2/MeterValuesRequestOcpp2Handler.test.ts`, 7 cases:

- meter values are linked to the EVSE's active transaction (also with `sendCostUpdatedOnMeterValue: false`, the shipped default);
- `totalKwh` is recalculated;
- readings are stored unlinked for `evseId: 0` and when the EVSE has no active transaction;
- `CostUpdated` fires when the setting is on and not when it is off.

5 of the 7 fail on `next` — including *notifies cost when sendCostUpdatedOnMeterValue is enabled*, which demonstrates the setting is inert today.

## Verification

- **after:** 1755 passed (+7 new), 2 failed, 4 skipped

The 2 failures are `packages/core/test/dal/e2e/security-event.e2e.test.ts`, which fail identically on unmodified `next` — pre-existing and unrelated. `eslint` reports 0 errors; the 2 remaining warnings are in `OcppRouter/DataApi.ts` and `Reporting/2/MessageApi.ts`, untouched here.

### DIFF
--- a/packages/core/src/handlers/requests/2/MeterValuesRequestOcpp2Handler.ts
+++ b/packages/core/src/handlers/requests/2/MeterValuesRequestOcpp2Handler.ts
@@ -103,8 +103,6 @@ export class MeterValuesRequestOcpp2Handler extends AbstractHandler {
 
       if (activeTransaction) {
         await this._transactionService.recalculateTotalKwh(activeTransaction, meterValuesCreated);
-        // Only the CostUpdated call is optional; attaching the reading to its transaction
-        // and keeping totalKwh current are not.
         if (this._sendCostUpdatedOnMeterValue) {
           await this._costNotifier.calculateCostAndNotify(
             activeTransaction,
--- a/packages/core/src/handlers/requests/2/MeterValuesRequestOcpp2Handler.ts
+++ b/packages/core/src/handlers/requests/2/MeterValuesRequestOcpp2Handler.ts
@@ -5,6 +5,7 @@ import {
   AbstractHandler,
   type AbstractHandlerDependencies,
   AsRequestHandler,
+  type BootstrapConfig,
   type IMessage,
   type IOcppSender,
   OcppError,
@@ -14,6 +15,7 @@ import {
   type HandlerProperties,
   OCPP_2_VER_LIST,
   OCPP_CallAction,
+  type SystemConfig,
   OCPP2_request_types,
   OCPP2_response_types,
 } from '@citrineos/types';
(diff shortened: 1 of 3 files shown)


yuvenus (approved):

sjh37:

thanaParis (approved):

PR #870, opened 2026-08-17
fix(ocpp): calculate transaction cost even when the event carries an idToken

sjh37:
> **Rebased onto `next`.** `next` has since added the F07 remote-start, C25 QR/DirectPayment, E16 limit-sync and E17 TxProfile blocks to both halves of the branch, and changed `calculateTotalCost` to take the transaction rather than `(connectorId, totalKwh)`; the merge and the two test assertions are redone against that. Most of the diff is still de-indentation - review with [`?w=1`](https://github.com/citrineos/citrineos-core/pull/870/files?w=1), where the real change is **+156 / -68** across both files, and **+23 / -68** in the handler itself.

## Problem

`TransactionEventRequestOcpp2Handler.handle` branched on `if (response)` (`:194`) — i.e. on whether the request carried an `idToken` and therefore produced an authorization response. Everything cost-related lived **only in the `else` branch** (`:383-665`):

- `Ended` `totalCost` calculation (`:554`) and `updateTransactionTotalCostById` (`:532`)
- I02 running total cost on `Updated` (`:406`)
- C20 zero-cost cancellation (`:562`)
- C21 `SettlementByCSMS` (`:596`)
- signed meter value validation (`:631`)
- E17 TxProfile re-send on `TxResumed` (`:480`)
- C23 `LimitSet` maxCost persistence (`:422`)

Per spec a `TransactionEvent(Ended)` carries the idToken that stopped the transaction. So the ordinary **"driver taps their card to stop charging"** path took the authorization branch and skipped all of the above: the session ended with no `totalCost` calculated or stored, and its meter value signatures unchecked.

None of that work depends on an idToken being present — it is a property of the event.

## Fix

Merge the two branches into one linear flow over a single `response` object, seeded from the authorization result when there is one:

```ts
const response: TransactionEventResponse = authorizeResponse ?? {};
```

Behaviour deliberately preserved:

- the **F07 remote-start** and **C25 DirectPayment** blocks keep their own guards (`Started`, `remoteStartId`, `idToken?.type === DirectPayment`), so they are unchanged in effect;
- the **E16 sync** still runs *before* the DirectPayment block, so a DB limit continues to suppress the `PaymentCtrlr.AuthorizationAmount` fallback;
- `persistTransactionLimitToDb` and the CallResult are now issued **once** rather than once per branch;
- the **CostUpdated scheduler** still only fires for an accepted authorized start — `idTokenInfo` is absent when no idToken was sent, so the existing guard is sufficient.

## Tests

New `transaction bookkeeping is independent of the idToken` describe, 6 cases: `totalCost` is calculated, stored and returned for an `Ended` event carrying an idToken; signed meter values are validated on such an event; the authorization result is still returned alongside the cost; and a regression guard that the no-idToken path still works.

5 of the 6 fail on `next`; the one that passes is the no-idToken case, which is precisely the branch that worked.

## Known adjacent issue, not fixed here

`if (response.totalCost && transaction)` at `:532` will not persist a legitimate `0`, which is what the C20 path at `:527` sets. That is a separate pre-existing bug and I have left it alone to keep this diff to the branch merge.

## Verification

On the rebase, `packages/core` (now `packages/ocpp`) in full:

```
 Test Files  101 passed | 1 skipped (102)
      Tests  1574 passed | 4 skipped (1578)
```

`pnpm --filter @citrineos/core run lint` is clean - 0 errors, 0 warnings.

The five with-idToken cases fail on unmodified `next`, and the no-idToken control passes:

```
× calculates totalCost for an Ended event that carries an idToken
× stores totalCost for an Ended event that carries an idToken
× returns totalCost to the charger for an Ended event that carries an idToken
× validates signed meter values on an event that carries an idToken
✓ still calculates and stores totalCost when the Ended event carries no idToken
× still returns the authorization result alongside the cost
```

### DIFF
--- a/packages/ocpp/src/handlers/requests/2/transaction-event-request-ocpp-2-handler.ts
+++ b/packages/ocpp/src/handlers/requests/2/transaction-event-request-ocpp-2-handler.ts
@@ -115,17 +115,18 @@ export class TransactionEventRequestOcpp2Handler extends AbstractHandler {
 
     const transactionEvent = message.payload;
     const transactionId = transactionEvent.transactionInfo.transactionId;
-    let response: OCPP2_response_types.TransactionEventResponse | undefined = undefined;
+    let authorizationResponse: OCPP2_response_types.TransactionEventResponse | undefined =
+      undefined;
     let transaction: Transaction | undefined = undefined;
     if (transactionEvent.idToken) {
       if (isOcpp21) {
-        response = await this._transactionService.authorizeOcpp21IdToken(
+        authorizationResponse = await this._transactionService.authorizeOcpp21IdToken(
           tenantId,
           transactionEvent,
           message.context,
         );
       } else {
-        response = await this._transactionService.authorizeOcpp201IdToken(
+        authorizationResponse = await this._transactionService.authorizeOcpp201IdToken(
           tenantId,
           transactionEvent,
           message.context,
@@ -133,7 +134,7 @@ export class TransactionEventRequestOcpp2Handler extends AbstractHandler {
       }
 
       recordAuthorizeResult({
-        status: response.idTokenInfo?.status,
+        status: authorizationResponse.idTokenInfo?.status,
         ocppVersion: String(message.protocol),
         action: 'TransactionEvent',
       });
@@ -149,7 +150,7 @@ export class TransactionEventRequestOcpp2Handler extends AbstractHandler {
         this._logger.warn(
           `Received ${transactionEvent.eventType} for already-ended transaction ${transactionId} on ${ocppConnectionName}. Ignoring.`,
         );
-        await this._ocppSender.sendCallResultWithMessage(message, response ?? {});
+        await this._ocppSender.sendCallResultWithMessage(message, authorizationResponse ?? {});
         return;
       }
     }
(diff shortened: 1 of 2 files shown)


sjh37:

`TransactionEventRequestOcpp2Handler` moved under it when #893 merged. Rebased; the handler tests are green (31).

sjh37:
Rebased onto `486f1e99` after #990. Clean rebase - the handler is now
`packages/ocpp/src/handlers/requests/2/transaction-event-request-ocpp-2-handler.ts` and its test
alongside it, both tracked as renames, so the diff is unchanged. `?w=1` still shows +23/-68 in the
handler; most of the rest is de-indentation.

Verified on `486f1e99`: build clean, full suite green (141 test files, 2307 passing / 4 skipped),
lint clean.

sjh37:
Rebased onto `c1bd4ca`, clean. The only comment this PR adds was in the test and is removed; the ones the diff appears to add in the handler are upstream's own, moved with the block that came out of the `else` branch, so they are left alone. `?w=1` still shows the real change.

Verified on `c1bd4ca`: build clean, full suite green (325 files, 4374 passing / 8 skipped), lint 0 errors.

thanaParis (approved):

PR #871, opened 2026-08-17
fix(base): ignore non-energy meter units instead of throwing

sjh37:
## Problem

`MeterValueUtils.normalizeToKwh` threw on any unit it did not recognise:

```ts
default:
  throw new Error(`Unknown unit for energy measurement: ${unit}`);
```

It is reached from `getTotalKwh` and `getMeterStart`, which run inside `MeterValuesRequestOcpp2Handler` and `TransactionEventRequestOcpp2Handler` (and, via `updateTransactionByMeterValues`, the 1.6 path). One malformed sampled value therefore aborts handling of the **entire message**: no meter values stored, no total recalculated, and the charger receives a `CallError` rather than a result.

This is easy for a charging station to trigger without meaning to. Per spec a sampled value with **no** `measurand` defaults to `Energy.Active.Import.Register` — and `findMeasurandValue` implements exactly that. So a charger that omits the measurand on a current sample sends something the CSMS reads as an energy register denominated in Amps:

```
MeterValueUtils.getTotalKwh([{ value: 16, unitOfMeasure: { unit: 'A' } }], 12.5)
-> Error: Unknown unit for energy measurement: A
```

Reproduced against the built package.

This is a different defect from #852, which corrects the *measurand* mapping; this is the *unit* switch.

## Fix

Return `null` instead of throwing. Both callers — `findMeasurandValue` and `sumPhasedValues` — already treat `null` as "no reading here", so the unusable sample is skipped and any well-formed readings in the same batch still count.

`sumPhasedValues` needed a matching change. It summed into a zero-initialised accumulator and returned it unconditionally:

```ts
let sum = 0;
for (const value of phaseValues) {
  const normalizedValue = this.normalizeToKwh(value);
  if (normalizedValue !== null) sum += normalizedValue;
}
return sum;
```

A phase set reported entirely in the wrong unit used to throw; without this change it would now yield a genuine-looking **0 kWh register reading**, and a 0 register against a non-zero `meterStart` produces a large negative total. Extracted `sumNormalized`, which returns `null` when nothing normalized. Both phase branches now share it.

## Tests

Four new cases under `getTotalKwh > Unit normalization > Unrecognised units`:

- a non-energy unit mid-batch is ignored and the surrounding readings still total correctly;
- a measurand-less value in Amps is ignored;
- `getMeterStart` skips it too;
- an unrecognised unit among phased values is skipped without poisoning the sum.

All four fail on `next` with `Error: Unknown unit for energy measurement: A`.

## Verification

- **after:** 1752 passed (+4 new), 2 failed, 4 skipped

The 2 failures are `packages/core/test/dal/e2e/security-event.e2e.test.ts`, which fail identically on unmodified `next` — pre-existing and unrelated.

### DIFF
--- a/packages/base/src/util/MeterValueUtils.ts
+++ b/packages/base/src/util/MeterValueUtils.ts
@@ -212,33 +212,36 @@ export class MeterValueUtils {
         return null;
       }
 
-      let sum = 0;
-      for (const value of phaseNeutralValues) {
-        const normalizedValue = this.normalizeToKwh(value);
-        if (normalizedValue !== null) {
-          sum += normalizedValue;
-        }
-      }
-
-      return sum;
+      return this.sumNormalized(phaseNeutralValues);
     }
 
     // Sum all the normalized phase values
+    return this.sumNormalized(phaseValues);
+  }
+
+  /**
+   * Sum a set of same-measurand sampled values in kWh, skipping any whose unit is not an energy
+   * unit. Returns null when none of them normalized, so that a phase set reported entirely in
+   * the wrong unit is treated as "no reading" rather than as a genuine 0 kWh.
+   */
+  private static sumNormalized(sampledValues: SampledValue[]): number | null {
     let sum = 0;
-    for (const value of phaseValues) {
+    let normalizedAny = false;
+    for (const value of sampledValues) {
       const normalizedValue = this.normalizeToKwh(value);
       if (normalizedValue !== null) {
         sum += normalizedValue;
+        normalizedAny = true;
       }
     }
 
-    return sum;
+    return normalizedAny ? sum : null;
   }
 
   /**
    * Convert a sampled value to kWh, applying unit multipliers.
    * @param value A SampledValueType entry.
-   * @returns The converted value in kWh, or null if unit is missing.
+   * @returns The converted value in kWh, or null when the unit is not an energy unit.
    */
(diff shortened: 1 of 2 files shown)


yuvenus (approved):

sjh37:
Closing this - #852 and #857 have between them settled the question the other way, and I think they are right.

This PR made `normalizeToKwh` return `null` for a non-energy unit rather than throw. The case it was written for is in its own comment: a sampled value with no measurand defaults to `Energy.Active.Import.Register`, so a Current reading in Amps arrived under an energy measurand and blew up the total.

**#852 removed that path** - an unmapped measurand is now dropped at the mapper instead of being relabelled as the energy register. With that gone, what is left is a station that explicitly sends `Energy.Active.Import.Register` labelled in Amps, which is out of spec, and **#857 has since made the contract explicit**: `normalizeToKwh` is public, returns `number`, and the doc says it throws for a genuinely unknown energy unit.

Carrying on would mean re-litigating a decision you have now made twice, on an input that should no longer occur. The four tests here go with it.

One thing worth keeping from the review of this area, unrelated to the throw: #868 is the separate fix for `getTotalKwh` returning `0` instead of `currentTotal` when a batch carries no usable energy reading - a TransactionEvent with only SoC currently zeroes the session total. That one is still open and still applies on top of #857.

PR #872, opened 2026-08-17
fix(core): write boot status and the action blacklist where the router reads them

sjh37:
## Problem

`MessageRouterImpl` gates traffic on two cache entries, both namespaced by the **tenant-scoped identifier**:

```ts
// router.ts:855
private _onCallIsAllowed(action, identifier) {
  return this._cache.exists(action, identifier).then((blacklisted) => !blacklisted);
}
// router.ts:917
const status = await this._cache.get<string>(BOOT_STATUS, identifier);
```

where `identifier` is `createIdentifier(tenantId, ocppConnectionName)` — `"1:CS001"`.

**Every writer used the bare station name as the namespace instead**, in both protocol versions:

| | namespace passed | resulting key |
|---|---|---|
| `BootNotificationRequestOcpp16Handler:92,157` | `ocppConnectionName` | `CS001:boot_status` |
| `BootNotificationRequestOcpp2Handler:96,167` | `ocppConnectionName` | `CS001:boot_status` |
| `BootNotificationService:157,168,317,329` | `ocppConnectionName` | `CS001:StatusNotification` |
| **router reads** | `identifier` | `1:CS001:boot_status` / `1:CS001:StatusNotification` |

Both cache implementations key on `` `${namespace}:${key}` `` (`util/cache/memory.ts:56`), and `IDENTIFIER_DELIMITER` is `:` (`identifiers.ts:16`), so the keys can never match.

### Consequences

- **B03.FR.08 is not enforced.** A charging station whose `BootNotification` was `Rejected` keeps receiving Calls from the CSMS, because `_sendCallIsAllowed` never finds the cached `Rejected` status.
- **Charger-originated actions are never blacklisted** during a `Pending` or `Rejected` boot, so a station that has not been accepted can still drive `StatusNotification`, `MeterValues` and `TransactionEvent` handling.
- The entries are **not tenant-scoped**, so two tenants owning a station of the same name share one entry.

## Fix

Pass the identifier from both boot handlers, and rename the parameter on `cacheChargerActionsPermissions` / `cacheOcpp16ChargerActionsPermissions` from `ocppConnectionName` to `identifier` so the contract is explicit rather than implied by the caller. Both handlers already computed `createIdentifier(...)` for their `CacheNamespace.Connections` lookup; that call is now hoisted and reused.

## Tests

New `packages/core/test/handlers/requests/BootNotificationCacheNamespace.test.ts` — 6 cases, 3 per protocol version: the boot status is read and written under the tenant-scoped identifier, and the blacklist call receives it. All 6 fail on `next` with `'station-001'` where `'1:station-001'` was expected.

Plus 6 contract cases in `BootNotificationService.test.ts` pinning the namespace the service writes to, including one asserting a same-named station in another tenant is unaffected.

## Scope notes

- The OCPP 1.6 **un-blacklist** loop has a second, independent defect — it destructures each action name as an array (`.map(async ([action]) => …)`) — which is fixed by **#865**. This change deliberately does not touch that line to avoid a conflict, and the 1.6 tests here cover only the boot status.
- The `GetBaseReport` `requestId` entries (`BootNotificationService:183`, `NotifyReportRequestOcpp2Handler:124,133`) also use the bare station name, but writer and reader agree there, so that pairing works. Left alone; it remains cross-tenant-unsafe and is worth a separate look.

## Verification

- **after:** 1760 passed (+12 new), 2 failed, 4 skipped

The 2 failures are `packages/core/test/dal/e2e/security-event.e2e.test.ts`, which fail identically on unmodified `next` — pre-existing and unrelated. `eslint` reports 0 errors.


yuvenus (approved):
Good catch, approved! 🙏 As noted in the PR description, with these changes, the only thing left tenant-unsafe would be instances such as `createGetBaseReportRequest` in `BootNotificationService`. But it's fine with me to address that in a separate PR to keep this PR focused. 👌

sjh37:
Opened #911 for the `BootNotificationService` gap you mentioned, kept separate as you suggested.

It turned out to be the GetBaseReport progress cache rather than `createGetBaseReportRequest` alone - the same key is written by `NotifyReportRequestOcpp2Handler` and waited on by `confirmGetBaseReportSuccess`, all three under the bare station name. Since the request id is the hard-coded 0, the key is literally `"0"` in namespace `<stationName>`, so two tenants with a station of the same name share one entry and one tenant's NotifyReport can release the other's boot.

I threaded `tenantId` into the two service methods rather than building the identifier in the handler, so it doesn't touch the lines this PR changes in `BootNotificationRequestOcpp2Handler`.

PR #874, opened 2026-08-17
fix(ocpi): scope transaction lookups so CDR/StopSession can't hit the wrong session

vicky-ahmed:


packages/ocpi-base/src/graphql/queries/transaction.queries.ts
```diff
@@ -64,9 +64,95 @@ export const GET_TRANSACTIONS_QUERY = gql`
...
+export const GET_TRANSACTION_BY_ID_QUERY = gql`
+  query GetTransactionById($id: Int!) {
+    Transactions_by_pk(id: $id) {
+      tenant: Tenant {
+        countryCode
+        partyId
+        name
+        isUserTenant
+      }
+      id
```
yuvenus, line 76: EDIT: Oh wait, we're in OCPI 😬 I don't think the consts exist here, right? Nevermind!

Any reason not to use the cool new const-d transaction fields (and by extension the consts for the other nested items also?)

yuvenus (approved):

PR #877, opened 2026-08-18
fix(core): close the OCPP 1.6 authorization bypass for a statusless token

sjh37:
## The bug

An `Authorization` row whose `status` is null is treated as an implicit **allow** on both OCPP 1.6 entry points:

- `TransactionService.authorizeOcpp16IdToken` returns `Accepted` immediately, skipping the `cacheExpiryDateTime` check, the concurrent-transaction check, and every authorizer — including `RealTimeAuthorizer`, which is the operator's only veto over a token at authorize time.
- `AuthorizeRequestOcpp16Handler` does the same for the standalone `Authorize` call.

```ts
// TransactionService.authorizeOcpp16IdToken
if (!authorization.status) {
  response.idTagInfo.status = OCPP1_6.StartTransactionResponseStatus.Accepted;
  return response;                    // no expiry, no concurrency, no authorizers
}
```

## Is a null status actually reachable?

Yes. `Authorizations.status` has no `NOT NULL`, and both the Sequelize and Drizzle models declare it optional. Rows reach that state through any write path that sets only the fields it was given — an OCPP 1.6 `idTag` is a bare string that carries no status of its own.

It fails open on the one protocol where an `idTag` has nothing else authenticating it.

## Precedent

ac4f60e9 (*"fixing authorization loophole for unknown tokens"*) removed exactly this branch from `authorizeOcpp201IdToken`. The two 1.6 copies were left behind; the 2.0.1 `Authorize` handler has never had one. This closes the same hole on 1.6.

## The change

Reject instead of accepting, and log which authorization was rejected so a misconfigured row is findable rather than silently permissive.

Adds `packages/core/test/handlers/requests/1.6/AuthorizeRequestOcpp16Handler.test.ts` — the handler had no test file.

## Verification

```
npx vitest run packages/core/test/handlers/requests/1.6 \
               packages/core/test/modules/Transactions/module/TransactionService.test.ts
 Test Files  4 passed (4)
      Tests  24 passed (24)
```

The three new tests fail on `next` before the source change, each with `expected 'Accepted' to be 'Invalid'`.

### DIFF
--- a/packages/core/src/handlers/requests/1.6/AuthorizeRequestOcpp16Handler.ts
+++ b/packages/core/src/handlers/requests/1.6/AuthorizeRequestOcpp16Handler.ts
@@ -89,7 +89,7 @@ export class AuthorizeRequestOcpp16Handler extends AbstractHandler {
       const authorization = authorizations[0];
 
       if (!authorization.status) {
-        response.idTagInfo.status = OCPP1_6.AuthorizeResponseStatus.Accepted;
+        this._logger.error(`Authorization for idToken ${request.idTag} has no status; rejecting.`);
       } else if (authorization.status === AuthorizationStatusEnum.Accepted) {
         const cacheExpiryDateTime = authorization.cacheExpiryDateTime;
         const groupAuthorizationId = authorization.groupAuthorizationId;
--- a/packages/core/src/modules/Transactions/src/module/TransactionService.ts
+++ b/packages/core/src/modules/Transactions/src/module/TransactionService.ts
@@ -328,7 +328,9 @@ export class TransactionService {
 
       // Check expiration and status
       if (!authorization.status) {
-        response.idTagInfo.status = OCPP1_6.StartTransactionResponseStatus.Accepted;
+        this._logger.error(
+          `Authorization ${authorization.id} for idToken ${idToken} has no status; rejecting.`,
+        );
         return response;
       }
 
--- a/packages/core/src/modules/Transactions/src/module/TransactionService.ts
+++ b/packages/core/src/modules/Transactions/src/module/TransactionService.ts
@@ -424,7 +424,6 @@ export class TransactionService {
     let evseTypeId: number | undefined;
 
     if (typeof evseIdentifier === 'number') {
-      // OCPP 1.6: evseIdentifier is a connector ID — resolve to EVSE type ID
       const connector = await this._locationRepository.readConnectorByStationIdAndOcpp16ConnectorId(
         tenantId,
         ocppConnectionName,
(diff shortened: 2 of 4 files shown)


devonNewsomDevelop (approved):
Looks good, thank you for the work!

devonNewsomDevelop (changes requested):
Ah, actually on a second parse, can we change the comments? They are a commentary of the current bug fix but not actually helpful for the code after the fix. It's basically a historical documentation in comment form which can be distracting for developers who may be seeking current context information. 

Please either remove these comments on packages/core/src/handlers/requests/1.6/AuthorizeRequestOcpp16Handler.ts
and 
packages/core/src/modules/Transactions/src/module/TransactionService.ts

or tweak them if you feel extra information is required in this portion of the code.

sjh37:
@devonNewsomDevelop Removed from both files. Fair point - they were a narrative of the bug rather than context for the code that is there now, and the `logger.error` on the same line already says what is happening and why. Nothing left in the source; the two test files still carry short comments describing the scenario under test, which I hope is the useful kind.

packages/core/src/handlers/requests/1.6/AuthorizeRequestOcpp16Handler.ts
```diff
@@ -89,7 +89,7 @@ export class AuthorizeRequestOcpp16Handler extends AbstractHandler {
       const authorization = authorizations[0];
 
       if (!authorization.status) {
-        response.idTagInfo.status = OCPP1_6.AuthorizeResponseStatus.Accepted;
+        this._logger.error(`Authorization for idToken ${request.idTag} has no status; rejecting.`);
```
sjh37, line 92: > can we change the comments? They are a commentary of the current bug fix but not actually helpful for the code after the fix

Done in `7b9de129` - both files you named carry no added comments at all now, rather than tweaked ones. I agree with the reasoning; the archaeology belongs in the PR body and the commit message, where it stays available without sitting in the way of someone reading for current context.

I have since applied the same rule across every open PR of mine, after @yuvenus asked for the same thing on citrineos/citrineos#222. Three comments survive anywhere, all deliberate, and I am happy to drop any of them on request:

- the `Tenants` one you accepted on #884;
- the OCSP responder guard on #929, where removing it invites someone to simplify the pinned-address request back into a plain `fetch`, which is the vulnerability;
- `/** @type {import('sequelize-cli').Migration} */`, the convention every existing migration carries.

`@param` / `@returns` tags on new parameters I have kept as API documentation rather than commentary - say the word if you would rather those went too.

thanaParis (approved):

PR #878, opened 2026-08-18
fix(core): correct the inverted EVSE restriction checks in the OCPP 2.x Authorize handlers

sjh37:
## The bug

Both restrictions an `Authorization` can carry resolve to the opposite of what they say. Present identically in `AuthorizeRequestOcpp201Handler` and `AuthorizeRequestOcpp21Handler`.

### `allowedConnectorTypes` — off-by-one

```ts
if (authorization.allowedConnectorTypes.indexOf(connectorType.value as string) > 0) {
  evseIds.add(connectorType.evse?.id as number);
}
```

`indexOf` returns `0` for the first entry, and `0 > 0` is false. So:

- A **single-entry** allow list matches no EVSE at all. `evseIds` comes out empty, and every driver holding that token is refused with `NotAllowedTypeEVSE` — the restriction rejects everyone, always.
- A multi-entry list silently ignores its first entry.

### `disallowedEvseIdPrefixes` — inverted

```ts
const evseIdAllowed: boolean = authorization.disallowedEvseIdPrefixes.some(
  (disallowedEvseId: string) => (evseIdAttribute.value as string).startsWith(disallowedEvseId),
);
if (evseIdAllowed && !authorization.allowedConnectorTypes) {
  evseIds.add(evseIdAttribute.evse?.id as number);
} else if (!evseIdAllowed && authorization.allowedConnectorTypes) {
  evseIds.delete(evseIdAttribute.evse?.id as number);
}
```

The `some(...)` answers *"is this EVSE **dis**allowed"*, but the result is bound to `evseIdAllowed` and both branches read it as "allowed". EVSEs matching a disallowed prefix get **added** to the permitted set; EVSEs matching none get **removed**. The restriction admits exactly the locations it was configured to keep the token out of.

### Empty-array inconsistency

The connector-type block tests `allowedConnectorTypes.length > 0`; the prefix block tests `!authorization.allowedConnectorTypes`. An empty array is truthy, so a token carrying `allowedConnectorTypes: []` skips the first block, is treated as restricted by the second, and ends up with an empty permitted set and `NotAtThisLocation`.

## The change

- `includes` instead of `indexOf(...) > 0`.
- Rename to `evseIdDisallowed` and act on it the right way round: a disallowed EVSE is always removed; a permitted one is added only when there was no connector-type filter that already decided membership.
- One hoisted `hasConnectorTypeRestriction` flag so both checks agree on what "restricted" means.

Adds `packages/core/test/handlers/requests/2/AuthorizeRequestOcpp201Handler.test.ts` — the handler had no test file. It covers both restrictions individually, their interaction, and the unrestricted case.

## Verification

```
npx vitest run packages/core/test/handlers
 Test Files  11 passed (11)
      Tests  91 passed (91)
```

Six of the eight new tests fail on `next` before the source change, e.g. `expected 'NotAllowedTypeEVSE' to be 'Accepted'` for a token whose only allowed connector type is present on the station, and `expected [ 1 ] to deeply equal [ 2 ]` where EVSE 1 is the one carrying the disallowed prefix.

### DIFF
--- a/packages/core/src/handlers/requests/2/AuthorizeRequestOcpp201Handler.ts
+++ b/packages/core/src/handlers/requests/2/AuthorizeRequestOcpp201Handler.ts
@@ -228,7 +228,6 @@ export class AuthorizeRequestOcpp201Handler extends AbstractHandler {
                 if (evseIdDisallowed) {
                   evseIds.delete(evseIdAttribute.component?.evse?.id as number);
                 } else if (!hasConnectorTypeRestriction) {
-                  // With no connector-type filter, this list alone decides the permitted set.
                   evseIds.add(evseIdAttribute.component?.evse?.id as number);
                 }
               }
--- a/packages/core/src/handlers/requests/2/AuthorizeRequestOcpp201Handler.ts
+++ b/packages/core/src/handlers/requests/2/AuthorizeRequestOcpp201Handler.ts
@@ -180,8 +180,6 @@ export class AuthorizeRequestOcpp201Handler extends AbstractHandler {
           // Authorization restrictions MUST provide these variable attributes as defined in Physical Component
           // list of Part 2 - Appendices of OCPP 2.0.1
           let evseIds: Set<number> | undefined = undefined;
-          // One flag for both checks below: an empty allowedConnectorTypes array is truthy but
-          // is not a restriction, and reading it as one left the permitted set empty.
           const allowedConnectorTypes = authorization.allowedConnectorTypes ?? [];
           const hasConnectorTypeRestriction = allowedConnectorTypes.length > 0;
           if (hasConnectorTypeRestriction) {
@@ -195,8 +193,6 @@ export class AuthorizeRequestOcpp201Handler extends AbstractHandler {
                 type: AttributeEnum.Actual,
               });
             for (const connectorType of connectorTypes) {
-              // indexOf(...) > 0 dropped the type at index 0, so a single-entry allow list
-              // matched nothing and every driver was refused with NotAllowedTypeEVSE.
               if (allowedConnectorTypes.includes(connectorType.value as string)) {
                 evseIds.add(connectorType.component?.evse?.id as number);
               }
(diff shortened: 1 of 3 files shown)


yuvenus (approved):

yuvenus:
Some overlap with https://github.com/citrineos/citrineos-core/pull/883. Unsure if the component reference is required here as well? 🤔

sjh37:
@yuvenus Good catch, and yes it is required - #883 is right and this PR was wrong.

`constructQuery` (`DeviceModel.ts:586-631`) eager-loads `EvseType` **through** the `Component` include, and never includes the `VariableAttribute.evse` association. Both associations exist on the model (`VariableAttribute.ts:199` and `Component.ts:71`), which is what made the mistake easy, but only the one behind `Component` is populated on rows this query returns. So `connectorType.evse?.id` was `undefined` and `evseIds` was being filled with `undefined` rather than real EVSE ids.

My tests did not catch it because the fixture mocked the shape I assumed rather than the shape the repository actually returns - my error, and a good argument for the fixture mirroring the real include.

I have pushed the same `component?.evse?.id` change here, on all three lookups per handler (the `disallowedEvseIdPrefixes` path reads `EVSE`/`EvseId` attributes through the same include), and corrected the test fixture to nest the EVSE under `component`.

On the overlap: #883 and this PR now make the same change to that line. This one additionally fixes `disallowedEvseIdPrefixes`, where the predicate is bound to a variable named `evseIdAllowed` but holds "is disallowed" and is then acted on as "allowed", and adds test coverage for both restrictions. Happy either way - merge #883 first and I will rebase this onto it, or take this one and close #883. Your call, I do not mind which.

yuvenus (approved):
Thanks for making the changes! Closing https://github.com/citrineos/citrineos-core/pull/883 given this PR ported those changes over and has additional changes that fix a different issue.

Ditto here as well, some of the in-code comments above the changes reference old code that isn't present anymore, so it'd be good to remove to avoid confusion for what they refer to, thanks!

sjh37:
Removed - three comments per handler, all of which described what the old code did rather than what the current code does.

I kept one line, on the `else if (!hasConnectorTypeRestriction)` branch: `// With no connector-type filter, this list alone decides the permitted set.` That branch condition is the least obvious thing left in the function and the comment is about the code as it now stands, not about the bug. Happy to drop it too if you'd rather.

sjh37:
Follow-up: the one line I kept has since gone too, so this PR now adds no comments at all. @yuvenus asked for the same thing across the board on citrineos/citrineos#222 and I have swept every open PR of mine to match.

PR #880, opened 2026-08-18
fix(core): ignore a retried OCPP 1.6 StopTransaction instead of double-counting it

sjh37:
## The bug

`StopTransactionRequestOcpp16Handler` acknowledges the request, then writes a `StopTransaction` row and re-inserts every meter value in `transactionData` — with no check that the transaction is still active.

A charging station that does not receive our `CallResult` retries the `StopTransaction`. That is routine on a link that drops. Each retry adds:

- another `StopTransaction` row for the same session, and
- another copy of the entire `transactionData` meter-value set.

So anything aggregating `MeterValue` rows for a session — charge curve, energy profile, departure planning — counts the same readings several times. `createStopTransaction` has no dedup of its own, so nothing downstream catches it.

The 2.x path already guards this. `TransactionEventRequestOcpp2Handler` drops a non-`Started` event whose transaction is no longer active:

```ts
if (existingTransaction && !existingTransaction.isActive) {
  this._logger.warn(`Received ${eventType} for already-ended transaction ... Ignoring.`);
  await this._ocppSender.sendCallResultWithMessage(message, authorizeResponse ?? {});
  return;
}
```

1.6 has no equivalent.

## The change

Apply the same rule to 1.6. The `CallResult` is already sent before this point, so a retry is still acknowledged and the station stops resending — only the duplicate write is dropped.

**Also**: derive `stoppedReason` once and write the same value to both rows. OCPP 1.6 makes `reason` optional, and `transaction.stoppedReason = request.reason` assigned it straight through, so `Transaction.stoppedReason` was left null for the commonest case while the `StopTransaction` row carried the derived value (`request.reason || (request.idTag ? 'Remote' : 'Local')`) — two rows disagreeing about why the same session ended.

Adds `packages/core/test/handlers/requests/1.6/StopTransactionRequestOcpp16Handler.test.ts` — the handler had none.

## Verification

```
npx vitest run packages/core/test/handlers/requests/1.6/StopTransactionRequestOcpp16Handler.test.ts
 Test Files  1 passed (1)
      Tests  5 passed (5)
```

Two fail on `next` before the source change: the duplicate-stop test (`expected "spy" to not be called at all, but actually been called 1 times`) and the stoppedReason-agreement test.

### DIFF
--- a/packages/core/src/handlers/requests/1.6/StopTransactionRequestOcpp16Handler.ts
+++ b/packages/core/src/handlers/requests/1.6/StopTransactionRequestOcpp16Handler.ts
@@ -118,10 +118,6 @@ export class StopTransactionRequestOcpp16Handler extends AbstractHandler {
     }
 
     if (!transaction.isActive) {
-      // A charger that misses our CallResult retries the StopTransaction. It has already been
-      // acknowledged above, so the retry only needs to be dropped here: writing it again adds a
-      // second StopTransaction row and re-inserts every meter value in transactionData, which
-      // double-counts the session's readings. The 2.x handler already guards this.
       this._logger.warn(
         `Received StopTransaction for already-ended transaction ${request.transactionId} on ${ocppConnectionName}. Ignoring.`,
       );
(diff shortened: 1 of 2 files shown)


yuvenus (approved):
Makes sense to me, approved! Ditto here, I think the in-code comments are unnecessary, at it should make sense why these changes are counted based on the error logs present, so please remove them. 👍

sjh37:
Both removed. The warn line and the test names carry what they need to.

PR #882, opened 2026-08-18
fix(ocpi): map Occupied connectors to EVSE status CHARGING

vicky-ahmed:


packages/ocpi-base/src/mapper/LocationMapper.ts
```diff
@@ -227,11 +227,7 @@ export class EvseMapper {
     return {
       uid: UID_FORMAT(station.ocppConnectionName, evse.id!),
       evse_id: evse.evseId,
-      status: connectors
-        ? EvseMapper.mapEvseStatusFromConnectors(
-            evse.connectors!.filter((c) => connectors.some((con) => con!.id === c.id!.toString())),
-          )
-        : EvseStatus.UNKNOWN,
+      status: EvseMapper.mapEvseStatusFromConnectors(evse.connectors ?? []),
```
yuvenus, line 230: why not just use the already-declared const `connectors` that's up a few lines? 🤔

packages/ocpi-base/src/mapper/LocationMapper.ts
```diff
@@ -262,10 +258,9 @@ export class EvseMapper {
     return {
       evse_id: evse.evseId,
       status:
-        connectors &&
-        EvseMapper.mapEvseStatusFromConnectors(
-          evse.connectors!.filter((c) => connectors.some((con) => con!.id === c.id!.toString())),
-        ),
+        evse.connectors && evse.connectors.length > 0
```
yuvenus, line 261: ditto, why not just use the connectors already defined a few lines above, rather than needing to reference evse.connectors again? 🤔

yuvenus (approved):
Two small comments, otherwise approved!

packages/ocpi-base/src/mapper/LocationMapper.ts
vicky-ahmed: Umm yeah due to the following reasons:
1: The const connectors is from ConnectorMapper.fromGraphql type which doesn't have the status field.
2: onnectors is a filtered list, so using it could drop the connector we need and lead to the original bug by incorrectly returning UNKNOWN again.
3: connectors can also be undefined when empty, while evse.connectors ?? [] safely gives us the raw connectors to determine the actual status.

PR #884, opened 2026-08-18
fix(core): scope readByKey and existsByKey to the tenant

sjh37:
Both methods take a `tenantId` as their first parameter, like every other method on `CrudRepository`, and both ignored it - they resolved through `findByPk`.

That is only safe where the primary key is itself tenant-scoped, and several are not. `"Boots"."id"` is the bare station name (`20250430103000-initial.ts:132-135`), while `ChargingStation` documents `ocppConnectionName` as *"Unique per tenant, but two different tenants may share the same value"* (`Location/ChargingStation.ts:73`). So a station belonging to tenant B read tenant A's boot configuration - heartbeat interval, accepted/rejected status, pending-report flags - on the BootNotification path. `BootNotificationService` calls `readByKey` three times, and `GET /data/configuration/boot` reaches the same path.

Scoped both methods the way `_updateByKey` and `_deleteByKey` already are, with one exception: `Tenants` is the tenancy root and carries no `tenantId` column of its own, so the predicate is only added for models that declare the attribute. `getMaxChargingStationsForTenant` reads that row while deciding whether a station may connect (`container.ts:363`), so adding the predicate unconditionally queries a column that does not exist and refuses every websocket connection - the security-event e2e suite catches exactly this.

Adds `TenantScopedRead.integration.test.ts` next to the existing `TenantScopedDelete.integration.test.ts`. Three of its assertions fail on `next`; the tenancy-root case fails against the naive fix.

### DIFF
--- a/packages/core/src/dal/layers/sequelize/repository/Base.ts
+++ b/packages/core/src/dal/layers/sequelize/repository/Base.ts
@@ -16,10 +16,6 @@ import { type Model, type Sequelize } from 'sequelize-typescript';
 import { type ILogObj, Logger } from 'tslog';
 import { DefaultSequelizeInstance } from '../util.js';
 
-/**
- * Dependencies every Sequelize repository takes. `logger` and `sequelizeInstance`
- * are optional — the base falls back to a default logger and the shared instance.
- */
 export interface SequelizeRepositoryDependencies {
   config: BootstrapConfig;
   logger?: Logger<ILogObj>;
--- a/packages/core/src/dal/layers/sequelize/repository/Base.ts
+++ b/packages/core/src/dal/layers/sequelize/repository/Base.ts
@@ -50,7 +50,13 @@ export class SequelizeRepository<T extends Model<any, any>> extends CrudReposito
     key: string | number,
     namespace: string = this.namespace,
   ): Promise<T | undefined> {
-    return await this.s.models[namespace].findByPk(key).then((row) => row as T);
+    const model = this.s.models[namespace];
+    const where: WhereOptions = { [model.primaryKeyAttribute]: key };
+    // Tenants is the tenancy root and carries no tenantId column of its own.
+    if ('tenantId' in model.getAttributes()) {
+      where.tenantId = tenantId;
+    }
+    return await model.findOne({ where }).then((row) => (row ?? undefined) as T | undefined);
   }
 
   async readAllByQuery(
@@ -95,7 +101,13 @@ export class SequelizeRepository<T extends Model<any, any>> extends CrudReposito
     key: string,
     namespace: string = this.namespace,
   ): Promise<boolean> {
-    return await this.s.models[namespace].findByPk(key).then((row) => row !== null);
+    const model = this.s.models[namespace];
+    const where: WhereOptions = { [model.primaryKeyAttribute]: key };
+    // Tenants is the tenancy root and carries no tenantId column of its own.
+    if ('tenantId' in model.getAttributes()) {
+      where.tenantId = tenantId;
+    }
+    return await model.findOne({ where }).then((row) => row !== null);
   }
 
   async existByQuery(
--- a/packages/core/src/dal/layers/sequelize/repository/Base.ts
+++ b/packages/core/src/dal/layers/sequelize/repository/Base.ts
@@ -51,11 +51,7 @@ export class SequelizeRepository<T extends Model<any, any>> extends CrudReposito
     namespace: string = this.namespace,
   ): Promise<T | undefined> {
     const model = this.s.models[namespace];
-    const where: WhereOptions = { [model.primaryKeyAttribute]: key };
-    // Tenants is the tenancy root and carries no tenantId column of its own.
-    if ('tenantId' in model.getAttributes()) {
-      where.tenantId = tenantId;
-    }
+    const where: WhereOptions = { [model.primaryKeyAttribute]: key, tenantId };
     return await model.findOne({ where }).then((row) => (row ?? undefined) as T | undefined);
   }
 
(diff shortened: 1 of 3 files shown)


packages/core/src/dal/layers/sequelize/repository/Base.ts
```diff
@@ -95,7 +103,12 @@ export class SequelizeRepository<T extends Model<any, any>> extends CrudReposito
     key: string,
     namespace: string = this.namespace,
   ): Promise<boolean> {
-    return await this.s.models[namespace].findByPk(key).then((row) => row !== null);
+    const model = this.s.models[namespace] as ModelStatic<any>;
```
devonNewsomDevelop: We should avoid 'any' casts. This should already be properly typed.

```
const model = this.s.models[namespace];
const where: WhereOptions = { [model.primaryKeyAttribute]: key };
```

packages/core/src/dal/layers/sequelize/repository/Base.ts
```diff
@@ -50,7 +50,15 @@ export class SequelizeRepository<T extends Model<any, any>> extends CrudReposito
     key: string | number,
     namespace: string = this.namespace,
   ): Promise<T | undefined> {
-    return await this.s.models[namespace].findByPk(key).then((row) => row as T);
+    // findByPk ignores the tenant. Several primary keys are not tenant-scoped - Boots is keyed on
+    // the station name alone, which ChargingStation documents as shareable between tenants - so
+    // this returned another tenant's row. Scoped the same way as _updateByKey and _deleteByKey.
+    const model = this.s.models[namespace] as ModelStatic<any>;
+    const where: WhereOptions<any> = { [model.primaryKeyAttribute]: key };
```
devonNewsomDevelop: Same deal here with the ```<any>``` casts.

packages/core/src/dal/layers/sequelize/repository/Base.ts
```diff
@@ -50,7 +50,15 @@ export class SequelizeRepository<T extends Model<any, any>> extends CrudReposito
     key: string | number,
     namespace: string = this.namespace,
   ): Promise<T | undefined> {
-    return await this.s.models[namespace].findByPk(key).then((row) => row as T);
+    // findByPk ignores the tenant. Several primary keys are not tenant-scoped - Boots is keyed on
```
devonNewsomDevelop: This commentary is more historical ledgering rather than useful for future developers. We should remove the comment entirely, the code is already self-documenting.

devonNewsomDevelop (changes requested):

sjh37:
@devonNewsomDevelop All three addressed:

- Dropped the `ModelStatic<any>` and `WhereOptions<any>` casts for `this.s.models[namespace]` and a plain `WhereOptions`, exactly as you wrote it. It compiles without them - the casts were defensive rather than necessary.
- Removed the historical comment on `readByKey`.
- Kept one line on the `getAttributes()` guard: `// Tenants is the tenancy root and carries no tenantId column of its own.` Happy to drop it too if you would rather, but the guard looks like a redundant check without it, and removing it takes out the station connection path - `getMaxChargingStationsForTenant` reads that row, so an unconditional tenantId predicate queries a column that does not exist and every websocket upgrade is refused. The security-event e2e suite catches it, but only after a confusing ten-second timeout.

packages/core/src/dal/layers/sequelize/repository/Base.ts
sjh37: Done, written exactly as you had it:

```ts
const model = this.s.models[namespace];
const where: WhereOptions = { [model.primaryKeyAttribute]: key };
```

It compiles without the casts - `ModelStatic<any>` and `WhereOptions<any>` were both unnecessary.

packages/core/src/dal/layers/sequelize/repository/Base.ts
sjh37: Same here - the `<any>` casts are gone and the plain library types are used. No `as` casts remain anywhere in the diff.

packages/core/src/dal/layers/sequelize/repository/Base.ts
sjh37: Removed. What is left on that branch is one line, in the present tense, stating the constraint rather than the incident:

```ts
// Tenants is the tenancy root and carries no tenantId column of its own.
```

I kept that one because without it the `'tenantId' in model.getAttributes()` guard reads like a defensive check that could be simplified away, and removing it breaks every station connection - `Tenant` genuinely has no such column. Say the word if you would rather it went too.

packages/core/src/dal/layers/sequelize/repository/Base.ts
```diff
@@ -50,7 +50,13 @@ export class SequelizeRepository<T extends Model<any, any>> extends CrudReposito
     key: string | number,
     namespace: string = this.namespace,
   ): Promise<T | undefined> {
-    return await this.s.models[namespace].findByPk(key).then((row) => row as T);
+    const model = this.s.models[namespace];
```
devonNewsomDevelop, line 49: We should not be doing model-specific checks like this in Base. This is against the intention of the design. A better home for this would be in the repository related to the model you are attempting to handle, such as SequelizeTenantRepository. We cannot have model specificity in our base class.

devonNewsomDevelop (changes requested):

packages/core/src/dal/layers/sequelize/repository/Base.ts
sjh37: Agreed - moved in 128653ff.

`readByKey` and `existsByKey` in `Base` now always narrow by `tenantId`, with no model knowledge, and `SequelizeTenantRepository` overrides both to read by key alone:

```ts
/**
 * A tenant is the scope rather than something inside one, so it is read by its own key. Narrowing
 * by tenantId as every other model does would query a column Tenants does not have.
 */
override async readByKey(_tenantId: number, key: string | number): Promise<Tenant | undefined> {
  return (await Tenant.findByPk(key)) ?? undefined;
}
```

Tenants is the only model of the 53 without a `tenantId` column, and `SequelizeTenantRepository` is the only place a repository is constructed over `Tenant.MODEL_NAME`, so the override is always the path taken - including `container.ts:363`, which is the station connection path that reads it.

The existing test for that case already went through `SequelizeTenantRepository`, so it now covers the override rather than the removed branch, and the security-event e2e boots a real server against it.

thanaParis (approved):

PR #885, opened 2026-08-18
fix(core): enforce the configured issuer and audience when verifying a JWT

sjh37:
`OIDCConfig` requires an `issuer` and accepts an `audience` (`OIDCAuthProvider.ts:26,29`, mirrored in the config schema at `packages/types/src/config/types.ts:249-250`). `jwt.verify` was called with neither (`OIDCAuthProvider.ts:109`), so the only thing actually checked was that the token was signed by a key the configured JWKS endpoint would serve.

An identity provider signs every client in a realm with those same keys. A token minted for an unrelated service in the same realm therefore verified here, and `authenticateToken` returned success with whatever `roles` claim the token carried - which `authorizeUser` then feeds straight into the RBAC lookup. The configured issuer and audience were, in effect, documentation.

This passes `issuer` always and `audience` when it is set (it is optional on the interface). No behaviour changes for a correctly-issued token.

Adds `OIDCAuthProvider.test.ts`, which generates an RSA key pair and signs real tokens rather than stubbing the verifier. On `next`, both the wrong-audience and wrong-issuer cases authenticate successfully.

Note this only affects deployments that enable OIDC; none of the checked-in env configs do - `local.ts` and `docker.ts` both select `localByPass`.

### DIFF
--- a/packages/core/src/util/authorization/provider/OIDCAuthProvider.ts
+++ b/packages/core/src/util/authorization/provider/OIDCAuthProvider.ts
@@ -105,10 +105,6 @@ export class OIDCAuthProvider implements IApiAuthProvider {
         throw new Error('Invalid token format');
       }
       const publicKey = await this.fetchPublicKey(decoded.header.kid);
-      // Verify the token with the public key. The issuer and audience are configured but were not
-      // being enforced: an identity provider signs every client in the realm with the same keys, so
-      // a token issued to an unrelated service verified here and authenticated its bearer with
-      // whatever roles it carried.
       const payload = jwt.verify(token, createPublicKey(publicKey), {
         issuer: this._config.issuer,
         ...(this._config.audience ? { audience: this._config.audience } : {}),
--- a/packages/core/src/util/authorization/provider/OIDCAuthProvider.ts
+++ b/packages/core/src/util/authorization/provider/OIDCAuthProvider.ts
@@ -26,7 +26,7 @@ export interface OIDCConfig {
   issuer: string;
 
   // Expected audience of tokens (usually the client ID)
-  audience?: string;
+  audience: string;
 
   // How long to cache keys (ms)
   cacheTime?: number;
@@ -107,7 +107,7 @@ export class OIDCAuthProvider implements IApiAuthProvider {
       const publicKey = await this.fetchPublicKey(decoded.header.kid);
       const payload = jwt.verify(token, createPublicKey(publicKey), {
         issuer: this._config.issuer,
-        ...(this._config.audience ? { audience: this._config.audience } : {}),
+        audience: this._config.audience,
       }) as JwtPayload;
 
       // Extract user info from the decoded token
--- a/packages/core/src/util/authorization/provider/OIDCAuthProvider.ts
+++ b/packages/core/src/util/authorization/provider/OIDCAuthProvider.ts
@@ -105,8 +105,14 @@ export class OIDCAuthProvider implements IApiAuthProvider {
         throw new Error('Invalid token format');
       }
       const publicKey = await this.fetchPublicKey(decoded.header.kid);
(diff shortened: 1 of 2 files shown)


packages/core/src/util/authorization/provider/OIDCAuthProvider.ts
```diff
@@ -105,8 +105,14 @@ export class OIDCAuthProvider implements IApiAuthProvider {
         throw new Error('Invalid token format');
       }
       const publicKey = await this.fetchPublicKey(decoded.header.kid);
-      // Verify the token with the public key
-      const payload = jwt.verify(token, createPublicKey(publicKey)) as JwtPayload;
+      // Verify the token with the public key. The issuer and audience are configured but were not
```
yuvenus: Code is pretty self-explanatory, you can remove the in-code comment. 👍

sjh37:
Removed. The four tests name the behaviour - a token from another issuer is rejected, one for another audience is rejected, audience is only enforced when configured - so the history belongs in the PR body rather than the file.

While I was here: `packages/ocpi-base/src/util/security/oidcAuthMiddleware.ts` already passes `issuer`, `audience` **and** `algorithms: ['RS256']` to `jwt.verify`. This change brings core into line with it. Core is not exploitable by algorithm confusion the way a naive version would be, because it hands `jwt.verify` a `createPublicKey()` KeyObject rather than a PEM string, which already rules out HMAC - so I have not added `algorithms` here. Happy to, if you would rather the two read identically.

packages/core/src/util/authorization/provider/OIDCAuthProvider.ts
```diff
@@ -105,8 +105,14 @@ export class OIDCAuthProvider implements IApiAuthProvider {
...
+      // being enforced: an identity provider signs every client in the realm with the same keys, so
+      // a token issued to an unrelated service verified here and authenticated its bearer with
+      // whatever roles it carried.
+      const payload = jwt.verify(token, createPublicKey(publicKey), {
+        issuer: this._config.issuer,
+        ...(this._config.audience ? { audience: this._config.audience } : {}),
```
yuvenus: Based on `packages/types/src/config` where the `authProvider.oidc` block lives, it seems audience can't be optional, so perhaps the OIDCConfig is incorrect in assuming audience is nullable? 🤨 We know that the config should be coming from that `authProvider.oidc` block in the config, based on the code over in `container.ts` where the `apiAuthProvider` is registered for DI under `registerServices`.

Can you fix OIDCConfig so that audience isn't optional any more? ? Then we don't  this optional check here.

yuvenus (changes requested):

packages/core/src/util/authorization/provider/OIDCAuthProvider.ts
sjh37: Removed. The four tests name the behaviour - a token from another issuer rejected, one for another audience rejected, an expired one rejected - so the history belongs in the PR description rather than the file.

packages/core/src/util/authorization/provider/OIDCAuthProvider.ts
sjh37: You are right, and it is fixed. Both `authProvider.oidc` blocks in `packages/types/src/config/types.ts` declare `audience: z.string()` with no `.optional()`, and the only production construction site is `new OIDCAuthProvider(config.util.authProvider.oidc, logger)` in `container.ts` - so the schema already guarantees it and `OIDCConfig` marking it `audience?: string` was the outlier. That optionality is exactly what the conditional spread was working around.

`audience: string` now, and `jwt.verify` passes it unconditionally. No test changes were needed: every case in the spec already supplies an audience, so there was never one asserting the unconfigured behaviour.

yuvenus (approved):

PR #886, opened 2026-08-18
fix(core): match excluded routes on the path, not the raw url

sjh37:
The global `onRequest` hook is handed `request.url`, which includes the query string, and `isExcludedRoute` compared that against `excludedRoutes` with `url === route || url.startsWith(`${route}/`)` (`ApiAuthPlugin.ts:87-99`).

So `/health` was excluded but `/health?cache-bust=1` matched neither test and was answered 401. Load balancers and liveness probes routinely append a parameter, so the health endpoint failed closed for precisely the callers it exists to serve. `/docs?...` behaves the same way.

Stripping the query string before matching fixes it. The prefix cases (`/health/live?probe=kubelet`) happened to work already, because they still start with `/health/` - the new test pins both.

For anyone wondering about the other direction: a crafted `/health/../data/...` does skip the auth hook, because the raw url starts with `/health/`. It is not exploitable - Fastify's router does not normalise `..`, so every such request 404s rather than reaching a handler. I verified that over a raw socket rather than through `inject()`, which normalises the path itself and would have hidden the behaviour. Worth knowing if the router is ever configured differently.

### DIFF
--- a/packages/core/src/util/authorization/ApiAuthPlugin.ts
+++ b/packages/core/src/util/authorization/ApiAuthPlugin.ts
@@ -85,12 +85,13 @@ const apiAuthPlugin: FastifyPluginAsync<{
 
   // Helper to check if a route is excluded from authentication
   function isExcludedRoute(url: string): boolean {
+    const path = url.split('?', 1)[0];
     // Always exclude health check
-    if (url === '/health') {
+    if (path === '/health') {
       return true;
     }
     const isExcluded = !!options.excludedRoutes?.some(
-      (route) => url === route || url.startsWith(`${route}/`),
+      (route) => path === route || path.startsWith(`${route}/`),
     );
     if (isExcluded && options.debug) {
       _logger.debug(`Skipping authentication for excluded route: ${url}`);
--- a/packages/core/test/util/authorization/ApiAuthPlugin.test.ts
+++ b/packages/core/test/util/authorization/ApiAuthPlugin.test.ts
@@ -1,0 +1,100 @@
+// SPDX-FileCopyrightText: 2026 Contributors to the CitrineOS Project
+//
+// SPDX-License-Identifier: Apache-2.0
+
+import Fastify, { type FastifyInstance } from 'fastify';
+import { afterEach, describe, expect, it, vi } from 'vitest';
+import type { IApiAuthProvider } from '@citrineos/base';
+import { apiAuthPluginFp } from '@util/authorization/ApiAuthPlugin.js';
+
+/**
+ * An auth provider that refuses everything, so that any route reaching the provider is observable as a
+ * 401 and any route skipping it is observable as a 200.
+ */
+function aRefusingProvider() {
+  return {
+    extractToken: vi.fn().mockResolvedValue(undefined),
+    authenticateToken: vi.fn().mockResolvedValue({ isAuthenticated: false, error: 'no token' }),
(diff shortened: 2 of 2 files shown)


packages/core/src/util/authorization/ApiAuthPlugin.ts
```diff
@@ -85,12 +85,16 @@ const apiAuthPlugin: FastifyPluginAsync<{
 
   // Helper to check if a route is excluded from authentication
   function isExcludedRoute(url: string): boolean {
+    // The hook is handed the raw url, query string and all. Comparing that against the configured
+    // routes meant '/health?probe=1' matched neither '/health' nor the '/health/' prefix, so any
+    // liveness probe or documentation request that carried a parameter was answered 401.
```
devonNewsomDevelop: We don't need this commentary here, but I appreciate the time and effort finding these holes and plugging them. Thank you!

packages/core/test/util/authorization/ApiAuthPlugin.test.ts
```diff
@@ -0,0 +1,100 @@
...
+  });
+
+  it('does not attempt authorization once authentication has already answered the request', async () => {
+    const provider = aRefusingProvider();
+    app = await anAuthenticatedServer(provider);
+
+    const response = await app.inject({ method: 'GET', url: '/data/transactions/transaction' });
+
+    // authenticate() replies 401 and returns; running authorize() afterwards sends a second reply on
+    // an already-answered request, which Fastify rejects as FST_ERR_REP_ALREADY_SENT.
```
devonNewsomDevelop: May want to tweak the wording on 95. As far as I understand, Fastify doesn't reject as FST_ERR_REP_ALREADY_SENT but rather logs it as a warning. Minor, minor tweak.

devonNewsomDevelop (changes requested):

sjh37:
@devonNewsomDevelop Both done, and thank you.

The source comment is gone - the diff is four lines now. On the test wording: you are right that it was wrong, and it was wrong in a way worth noting. I checked before writing this PR whether the unconditional `authorize()` after a failed `authenticate()` actually produces a second reply, and it does not surface as an error at all - the reply is already sent, the status stays 401, and nothing is logged in the setup I tried. The assertion never checked that anyway; it checks that the provider is not consulted for a caller who never authenticated. Reworded to say only that.

packages/core/src/util/authorization/ApiAuthPlugin.ts
sjh37: Removed - the source diff is four lines now. The path split and the `/health` early return say what they do without help.

packages/core/test/util/authorization/ApiAuthPlugin.test.ts
sjh37: You are right, and it was wrong in a way worth having caught. I reworded it.

I checked before writing the original: Fastify does not reject when a second send is attempted on an already-sent reply - it logs a warning and drops it, and the route still resolves. So the test could never have observed `FST_ERR_REP_ALREADY_SENT`, and naming it implied a guarantee that does not exist.

The comment now describes what the assertions actually prove: `authenticate()` answers the request and returns, and `authorize()` must not then consult the provider for a caller that never authenticated - which is what `expect(provider.authorizeUser).not.toHaveBeenCalled()` checks.

devonNewsomDevelop (approved):
Thank you for the fix.

PR #888, opened 2026-08-18
fix(ocpi): map the Other idToken type instead of warning that it is unmapped

sjh37:
`mapOcpiTokenTypeToOcppIdTokenType` maps `TokenType.OTHER` to `IdTokenEnum.Other`, but `mapOcppIdTokenTypeToOcpiTokenType` had no `case` for `Other` (`TokensMapper.ts:62-82`). It fell through to `default`, which logs *"Unmapped OCPP idToken type"* - for a value the forward mapping produces itself.

So the warning fired on ordinary token reads, at whatever rate tokens are read, and buried the cases it was presumably added for: `eMAID`, `ISO15693`, `KeyCode` and `MacAddress` genuinely have no representation in OCPI's four-value `TokenType` and collapse to `OTHER` irreversibly. Those are worth a log line; `Other` is not.

Adds `TokensMapper.test.ts` - only the second test file in `packages/ocpi-base` - covering the round trip for all four OCPI types, that `Other` maps without warning, and that `MacAddress` still warns.

Context for anyone integrating: the `MacAddress` collapse is load-bearing if you use autocharge. OCPP 2.0.1 resolves an idToken on `(idToken, type)`, so a vehicle enrolled through OCPI Tokens comes back as `Other` and will not match. That is a protocol limitation rather than a bug, but it means vehicle enrolment has to go through a channel that can set `idTokenType` directly.

### DIFF
--- a/packages/ocpi-base/src/mapper/TokensMapper.ts
+++ b/packages/ocpi-base/src/mapper/TokensMapper.ts
@@ -73,9 +73,6 @@ export class TokensMapper {
       case null:
         return TokenType.OTHER;
       default: {
-        // OCPI has only 4 token types; every other OCPP idTokenType (eMAID, ISO15693, KeyCode,
-        // MacAddress) maps to OTHER rather than throwing. Other is handled above, without a
-        // warning: it is what mapOcpiTokenTypeToOcppIdTokenType produces for TokenType.OTHER.
         Container.get(Logger).warn(
           `Unmapped OCPP idToken type "${type}"; defaulting to OCPI TokenType.OTHER`,
         );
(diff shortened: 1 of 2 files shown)


yuvenus (approved):

sjh37:
Rebased onto `next`. The only conflict was `TokensMapper.test.ts`, which #856 also added - the two sets of tests cover different functions (`mapOcpiTokenToPartialOcppAuthorization` there, `mapOcppIdTokenTypeToOcpiTokenType` here), so I have kept both in the one file rather than dropping either. The source change is unaffected by #856; it is in a different method.

sjh37:

PR #889, opened 2026-08-18
fix(certificates): make GetCertificateStatus OCSP flow fixed

vicky-ahmed:


yuvenus:
I may be a bit unclear, is the only change that the new helper method `createOcspRequest` lowercases the hashAlgorithm, which solves the issue described? Was the original issue that the hashAlgorithm wasn't passed properly in the first place? :O

vicky-ahmed:
@yuvenus The main issue was not that hashAlgorithm was missing, it was passed, but as uppercase SHA256 instead of the lowercase sha256 expected by jsrsasign. The bigger issue was the wrong parameter names like keyhash, namehash, serial, which jsrsasign 11.x doesn’t support, so it used to throw an error before the request could be built.

lydiazcheng (approved):

PR #890, opened 2026-08-19
fix(core): handle FirmwareStatusNotification on OCPP 1.6

sjh37:
A charging station on OCPP 1.6 that reports firmware update progress gets a `CallError` back.

`UpdateFirmware` is exposed as a 1.6 message endpoint (`Configuration/src/module/1.6/MessageApi.ts:246`) and `FirmwareStatusNotification` is listed in the Configuration module's 1.6 `requests`, but no handler declares the `(OCPP1_6, FirmwareStatusNotification, Request)` triple that `AbstractModule` indexes on.

The message still reaches the module - `_declaredActions` is protocol agnostic (its own comment says so, `AbstractModule.ts:176`) and the 2.x handler declares the action - and then the protocol-aware lookup at `AbstractModule.ts:301` misses and throws `NotSupported`, which is returned to the station as a CallError (`:305-312`).

So a 1.6 station reporting `Downloading` / `Downloaded` / `Installing` / `Installed` / `InstallationFailed` receives a protocol error instead of the acknowledgement 1.6 requires, and an operator who started a 1.6 firmware update has no way to learn whether it worked.

The 1.6 `DiagnosticsStatusNotification` sibling already exists and this handler is modelled directly on it; the schema and the action registration were already in place, only the handler was missing.

The test asserts on the decorator metadata rather than on behaviour alone, because "no handler declares this pair" is the actual defect - it reports zero declaring handlers on `next`.

Two related gaps I did **not** fix here, since they log rather than reaching the station: `UpdateFirmware` and `UnlockConnector` have 1.6 message endpoints but no 1.6 *response* handlers, so each successful command logs a handler error. Happy to add those too if wanted.


yuvenus:
Should resolve https://github.com/citrineos/citrineos/issues/216

yuvenus (approved):

vicky-ahmed (approved):

PR #891, opened 2026-08-19
fix(ocpi): skip an unmappable token instead of failing the whole CDR or Session batch

sjh37:
`getTokensForTransactions` (`BaseTransactionMapper.ts:65-92`) loops over a page of transactions and already has a branch for a token it cannot map:

```ts
if (tokenDto) { ... } else { this.logger.debug(`Unmapped token for transaction ${transaction.id}`); }
```

That branch is unreachable. `TokensMapper.toDto` never returns falsy - it throws - so one unmappable authorization aborts the entire page of Sessions or CDRs rather than being skipped.

Two things make it throw:

1. `additionalInfo` is `.nullable().optional()` on `AuthorizationDto` (`authorization.dto.ts:22`) and is unset on every Authorization not created through OCPI, yet `getContractId`, `getVisualNumber` and `getIssuer` reach through it with a non-null assertion. Callers get `TypeError: Cannot read properties of undefined (reading 'find')` instead of the descriptive error written directly below each lookup.
2. Even with `additionalInfo` present, a token carrying no eMAID genuinely has nothing to put in an OCPI `contract_id`, so the descriptive error is the right outcome for that token - just not for the other 99 in the batch.

This relaxes the three assertions so the intended errors are reachable, and catches per transaction so the rest of the batch still maps.

Worth flagging for anyone running a private or depot estate: sessions there are commonly authorised by tokens created directly against CitrineOS rather than received over OCPI, so they have no `additionalInfo` at all. Before this change, a single such session made the Sessions and CDRs endpoints unable to return the page containing it.

### DIFF
--- a/packages/ocpi-base/src/mapper/BaseTransactionMapper.ts
+++ b/packages/ocpi-base/src/mapper/BaseTransactionMapper.ts
@@ -79,15 +79,14 @@ export abstract class BaseTransactionMapper {
         }
       }
       if (transaction.authorization) {
-        // toDto throws for an authorization with no eMAID to use as contract_id; skip that
-        // transaction rather than failing the whole page.
-        try {
-          const tokenDto = await TokensMapper.toDto(transaction.authorization);
+        // An OCPI token carries an eMAID as its contract_id, and an authorization without one
+        // cannot be expressed as a token. That transaction is left out rather than failing the
+        // whole page; anything else that goes wrong here is still an error.
+        if (TokensMapper.findContractId(transaction.authorization)) {
+          const tokenDto = TokensMapper.toDto(transaction.authorization);
           transactionIdToTokenMap.set(transaction.transactionId!, tokenDto);
-        } catch (error) {
-          this.logger.debug(
-            `Unmapped token for transaction ${transaction.id}: ${error instanceof Error ? error.message : error}`,
-          );
+        } else {
+          this.logger.debug(`No contract id for transaction ${transaction.id}; token omitted`);
         }
       } else {
         this.logger.debug(`No token for transaction ${transaction.id}`);
--- a/packages/ocpi-base/src/mapper/BaseTransactionMapper.ts
+++ b/packages/ocpi-base/src/mapper/BaseTransactionMapper.ts
@@ -79,9 +79,6 @@ export abstract class BaseTransactionMapper {
         }
       }
       if (transaction.authorization) {
-        // An OCPI token carries an eMAID as its contract_id, and an authorization without one
-        // cannot be expressed as a token. That transaction is left out rather than failing the
-        // whole page; anything else that goes wrong here is still an error.
         if (TokensMapper.findContractId(transaction.authorization)) {
           const tokenDto = TokensMapper.toDto(transaction.authorization);
           transactionIdToTokenMap.set(transaction.transactionId!, tokenDto);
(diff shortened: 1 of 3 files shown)


packages/ocpi-base/src/mapper/BaseTransactionMapper.ts
```diff
@@ -79,11 +79,17 @@ export abstract class BaseTransactionMapper {
         }
       }
       if (transaction.authorization) {
-        const tokenDto = await TokensMapper.toDto(transaction.authorization);
-        if (tokenDto) {
+        // toDto throws when the authorization cannot be expressed as an OCPI token - it has no
```
devonNewsomDevelop: If we feel this commentary is necessary, can we trim it down to be more concise for a developer?

The rest of this looks great, thank you for the work!

devonNewsomDevelop (changes requested):

sjh37:
@devonNewsomDevelop Trimmed to two lines:

```ts
// toDto throws for an authorization with no eMAID to use as contract_id; skip that
// transaction rather than failing the whole page.
```

Kept that much because the `try` looks like defensive over-caution otherwise - the throw is the normal path for any token not created through OCPI, not an edge case. Thanks for the review.

packages/ocpi-base/src/mapper/BaseTransactionMapper.ts
sjh37: Trimmed to two lines:

```ts
// toDto throws for an authorization with no eMAID to use as contract_id; skip that
// transaction rather than failing the whole page.
```

I kept that much because the `try`/`catch` around a mapper call looks like something to simplify away otherwise - removing it would quietly restore the behaviour the PR fixes. Happy to drop it entirely if you would rather.

packages/ocpi-base/src/mapper/BaseTransactionMapper.ts
```diff
@@ -79,11 +79,15 @@ export abstract class BaseTransactionMapper {
...
+        // toDto throws for an authorization with no eMAID to use as contract_id; skip that
+        // transaction rather than failing the whole page.
+        try {
+          const tokenDto = await TokensMapper.toDto(transaction.authorization);
           transactionIdToTokenMap.set(transaction.transactionId!, tokenDto);
-        } else {
-          this.logger.debug(`Unmapped token for transaction ${transaction.id}`);
+        } catch (error) {
+          this.logger.debug(
+            `Unmapped token for transaction ${transaction.id}: ${error instanceof Error ? error.message : error}`,
```
devonNewsomDevelop: This catch assumes that any error from casting to the dto will always and only be due to an unmapped token, but other errors could still happen and now this silently allows them. Examples: tenantPartner!.countryCode! if tenantPartner is somehow null, errors if there is missing additionalInfo, that kind of thing.

devonNewsomDevelop (changes requested):

packages/ocpi-base/src/mapper/BaseTransactionMapper.ts
sjh37: You are right, and it is gone in 3f8551e7 - there is no catch any more.

An OCPI token carries an eMAID as its `contract_id`, and that is the one thing an authorization can be missing here, so it is checked directly instead:

```ts
if (TokensMapper.findContractId(transaction.authorization)) {
  transactionIdToTokenMap.set(transaction.transactionId!, TokensMapper.toDto(transaction.authorization));
} else {
  this.logger.debug(`No contract id for transaction ${transaction.id}; token omitted`);
}
```

`findContractId` is the lookup `getContractId` already did before throwing, so the two cannot drift.

Your examples now behave as they should - a null `tenantPartner` propagates rather than being counted as an unmapped token. Added a test for exactly that, which fails against the previous revision with `promise resolved "Map{}" instead of rejecting`: the session had silently vanished from the page.

thanaParis (approved):

PR #892, opened 2026-08-19
fix(core): run one contract certificate check, not both

sjh37:
> Rewritten after @lydiazcheng's review. My original premise was wrong - see the note at the bottom.

## Which check decides is currently an accident of ordering

```ts
if (request.iso15118CertificateHashData?.length) {
  response.certificateStatus = await validateCertificateHashData(...);
}
if (request.certificate) {
  response.certificateStatus = await validateCertificateChainPem(request.certificate);
}
if (response.certificateStatus !== Accepted) { /* reject */ }
```

Both write the same field, so when a station sends both, the outcome is decided by whichever
assignment happens to run second rather than by any stated rule.

`validateCertificateChainPem` already runs OCSP over every certificate in the chain, using each
certificate's own AIA responder URL, and additionally checks expiry and resolves the trust root. So
the hash data adds nothing when a chain is present, which is what the spec text @lydiazcheng quoted
says: `iso15118CertificateHashData` is *"Not needed if certificate is provided"*.

There is a second reason not to run both. `validateCertificateHashData` sends its OCSP request to
`reqData.responderURL` - a URL supplied by the charging station:

```ts
await sendOCSPRequest(ocspRequest, reqData.responderURL)
```

When a chain is also present, that outbound request is made and the result is then discarded. The
chain path instead uses the URL from the certificate's own AIA extension, which is not
station-controlled.

The chain is now used when present, the hash data otherwise.

## One deviation from the suggested snippet

The suggestion was:

```ts
if (request.certificate || request.iso15118CertificateHashData?.length) {
```

I kept the outer guard as `if (request.iso15118CertificateHashData || request.certificate)` and made
the inner branch an `else if` on `?.length`. The reason is an edge case: today, an empty
`iso15118CertificateHashData` array with no certificate enters the block, matches neither inner
branch, leaves `certificateStatus` undefined, and is rejected as `Invalid`. Moving the `?.length`
check to the outer guard would skip the block entirely and let the authorization proceed.

`minItems` is 1 on that field so a schema-valid request cannot do this, but the `else if` keeps the
fail-closed behaviour for anything that bypasses validation, at no cost. There is a test for it.
Happy to take the simpler guard if you'd rather.

## Where I was wrong

The original version of this PR made the hash data's refusal win, on the reasoning that the chain
check did not cover revocation. It does - `validateCertificateChainPem` returns `CertificateRevoked`
from its own OCSP loop. So there was no hole of the kind I described, and had my version been
merged it would have preferred the *station-supplied* responder's verdict over the certificate's
own. That was the wrong way round. Thanks for catching it.

The 2.0.1 and 2.1 handlers carry the same block and both are changed.

## Verification

```
npx vitest run test/handlers/requests/2/AuthorizeRequestOcpp201HandlerCertificates.test.ts   # 4 passed
npx prettier --check / npx eslint <changed files>  # clean
```

### DIFF
--- a/packages/core/src/handlers/requests/2/AuthorizeRequestOcpp201Handler.ts
+++ b/packages/core/src/handlers/requests/2/AuthorizeRequestOcpp201Handler.ts
@@ -131,7 +131,14 @@ export class AuthorizeRequestOcpp201Handler extends AbstractHandler {
       // If Charging Station is not able to validate a contract certificate,
       // it SHALL pass the contract certificate chain to the CSMS in certificate attribute (in PEM
       // format) of AuthorizeRequest for validation by CSMS, see C07.FR.06
-      if (request.certificate) {
+      // Only consulted while the hash data has not already refused the certificate: both results
+      // are written to the same field, so an unconditional call let a chain that still verifies
+      // overwrite a CertificateRevoked or CertificateExpired verdict from the OCSP data.
+      if (
+        request.certificate &&
+        (response.certificateStatus === undefined ||
+          response.certificateStatus === AuthorizeCertificateStatusEnum.Accepted)
+      ) {
         response.certificateStatus =
           await this._certificateAuthorityService.validateCertificateChainPem(request.certificate);
       }
--- a/packages/core/src/handlers/requests/2/AuthorizeRequestOcpp201Handler.ts
+++ b/packages/core/src/handlers/requests/2/AuthorizeRequestOcpp201Handler.ts
@@ -124,9 +124,7 @@ export class AuthorizeRequestOcpp201Handler extends AbstractHandler {
       // TODO - implement validation using cached OCSP data described in C07.FR.05
       // If Charging Station is not able to validate a contract certificate,
       // it SHALL pass the contract certificate chain to the CSMS in certificate attribute (in PEM
-      // format) of AuthorizeRequest for validation by CSMS, see C07.FR.06. The chain check runs
-      // OCSP over every certificate in it, so the hash data adds nothing when it is present, and
-      // iso15118CertificateHashData is documented as not needed in that case.
+      // format) of AuthorizeRequest for validation by CSMS, see C07.FR.06
       if (request.certificate) {
         response.certificateStatus =
           await this._certificateAuthorityService.validateCertificateChainPem(request.certificate);
--- a/packages/core/src/handlers/requests/2/AuthorizeRequestOcpp201Handler.ts
+++ b/packages/core/src/handlers/requests/2/AuthorizeRequestOcpp201Handler.ts
@@ -122,25 +122,19 @@ export class AuthorizeRequestOcpp201Handler extends AbstractHandler {
     // Validate Contract Certificates based on OCPP 2.0.1 Part 2 C07
     if (request.iso15118CertificateHashData || request.certificate) {
       // TODO - implement validation using cached OCSP data described in C07.FR.05
-      if (request.iso15118CertificateHashData && request.iso15118CertificateHashData.length > 0) {
+      // If Charging Station is not able to validate a contract certificate,
+      // it SHALL pass the contract certificate chain to the CSMS in certificate attribute (in PEM
+      // format) of AuthorizeRequest for validation by CSMS, see C07.FR.06. The chain check runs
+      // OCSP over every certificate in it, so the hash data adds nothing when it is present, and
+      // iso15118CertificateHashData is documented as not needed in that case.
+      if (request.certificate) {
+        response.certificateStatus =
+          await this._certificateAuthorityService.validateCertificateChainPem(request.certificate);
+      } else if (request.iso15118CertificateHashData?.length) {
         response.certificateStatus =
           await this._certificateAuthorityService.validateCertificateHashData(
             request.iso15118CertificateHashData,
           );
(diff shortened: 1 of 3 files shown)


yuvenus:
Requesting review from @lydiazcheng whenever she gets a chance since she has the most familiarity with the certificates flow. 👍

lydiazcheng:
Hi @sjh37 , thanks for this PR. The result of cert chain always win is based on this description in OCPP 2.0.1, 1.1.1. AuthorizeRequest, iso15118CertificateHashData (same for OCPP 2.1): **Not needed if certificate is provided.**  But I also think an improvement can be done for this part, e.g.,
```typescript
if (request.certificate || request.iso15118CertificateHashData?.length) {
    if (request.certificate) {
        response.certificateStatus = await this._certificateAuthorityService.validateCertificateChainPem(request.certificate);
    } else {
        response.certificateStatus = await this._certificateAuthorityService.validateCertificateHashData(request.iso15118CertificateHashData!);
    }
...
}
```

sjh37:
@lydiazcheng you're right and I've rewritten it to your structure - thank you.

I went and read `validateCertificateChainPem` properly after your comment. It runs OCSP over every certificate in the chain itself and returns `CertificateRevoked` from its own loop, so the gap I claimed doesn't exist. Worse, my version would have preferred the verdict from `reqData.responderURL` - a URL the *charging station* supplies - over the one from the certificate's own AIA extension. Exactly backwards.

The remaining change is the one you suggested: use the chain when it's there, the hash data otherwise. Beyond tidiness it drops an outbound OCSP request to a station-supplied URL whose result was being discarded anyway.

One deviation, flagged in the description: I kept the outer guard and made the inner branch an `else if` on `?.length`, because moving `?.length` into the outer guard changes what happens for an empty `iso15118CertificateHashData` array - today that's rejected as `Invalid`, and with the literal snippet it would skip the block and authorise. `minItems` is 1 so a schema-valid request can't get there, but it costs nothing to keep it fail-closed. Say the word and I'll simplify it.

lydiazcheng (approved):

packages/core/src/handlers/requests/2/AuthorizeRequestOcpp201Handler.ts
```diff
@@ -122,18 +122,17 @@ export class AuthorizeRequestOcpp201Handler extends AbstractHandler {
...
       // it SHALL pass the contract certificate chain to the CSMS in certificate attribute (in PEM
       // format) of AuthorizeRequest for validation by CSMS, see C07.FR.06
       if (request.certificate) {
         response.certificateStatus =
           await this._certificateAuthorityService.validateCertificateChainPem(request.certificate);
+      } else if (request.iso15118CertificateHashData?.length) {
+        response.certificateStatus =
+          await this._certificateAuthorityService.validateCertificateHashData(
+            request.iso15118CertificateHashData,
+          );
```
sjh37, line 135: @lydiazcheng thank you - and thank you for the C07 citation, that is exactly the sentence I was working from.

On the refactor: I agree the `if (a || b) { if (b) ... else if (a?.length) ... }` shape is awkward, and your `else` is cleaner than my `else if`. But I do not want to apply your outer guard silently, because it changes where this fails, and that is your area rather than mine to decide.

**The difference is an empty `iso15118CertificateHashData` array with no certificate.**

- Mine: `request.iso15118CertificateHashData` is `[]`, which is truthy, so the block is entered. Neither inner branch matches, `certificateStatus` stays `undefined`, and `certificateStatus !== Accepted` refuses the authorization. **Fails closed.**
- Yours: `[].length` is `0`, so the block is skipped entirely and the request proceeds as though no contract certificate were involved. **Fails open.**

There is a real argument for yours that I had not weighed: an absent field and an empty array both mean "I am not giving you hash data", and my version treats them differently - absent proceeds, empty refuses. That inconsistency is a fair thing to point at.

The argument for mine is that `iso15118CertificateHashData` has `minItems: 1`, so an empty array is a schema violation that should not reach the handler at all. When it does, something has bypassed validation, and I would rather a contract-certificate check refuse than wave the session through.

I lean to fail-closed, but not strongly, and you own this flow - tell me which you want and I will make it match. Either way the behaviour is pinned by `refuses a request carrying neither a chain nor any hash data to check` in `AuthorizeRequestOcpp201HandlerCertificates.test.ts`, so it cannot drift silently; I will flip that test with the code if you prefer the other direction.

(The same block exists in `AuthorizeRequestOcpp21Handler.ts` and will change with it.)

lydiazcheng: Hi @sjh37, I think the fail-closed version makes more sense. It is good to go with your version.

sjh37: Thank you @lydiazcheng - fail-closed it is, so nothing changes: the branch already behaves that way and the test that pins it (`refuses a request carrying neither a chain nor any hash data to check`) stays as it is. Both handlers, 2.0.1 and 2.1, carry the same shape.

Resolving this one. Ready to merge whenever suits you.

lydiazcheng: Great. Thank you!

PR #893, opened 2026-08-19
fix(core): send the transaction's SetChargingProfile on the station's own protocol

sjh37:
`TransactionEventRequestOcpp2Handler` is registered for `OCPP_2_VER_LIST` but hard-codes `protocol: OCPPVersion.OCPP2_1` on the `SetChargingProfile` it sends after applying a transaction charging profile (`TransactionEventRequestOcpp2Handler.ts:517`). The profile passed in it is built by `OCPP2_0_1_Mapper.ChargingProfileMapper` three lines above.

So a 2.0.1 station's outbound call is recorded and metered under 2.1 and its action mapped through the wrong action table.

This is the same defect as citrineos/citrineos-core#881, which fixed `NotifyEVChargingNeedsRequestOcpp2Handler`, `ClearChargingProfileResponseOcpp2Handler` and `SendLocalListResponseOcpp2Handler`. This is a fourth site that pass missed - I found it by checking every `OCPP_2_VER_LIST` handler that calls `sendCall`, and this is now the only remaining one; every other such handler already passes `message.protocol`.

`OCPPVersion` stays imported - it is still used at `:115` for `isOcpp21`.

### DIFF
--- a/packages/core/src/handlers/requests/2/TransactionEventRequestOcpp2Handler.ts
+++ b/packages/core/src/handlers/requests/2/TransactionEventRequestOcpp2Handler.ts
@@ -514,7 +514,12 @@ export class TransactionEventRequestOcpp2Handler extends AbstractHandler {
                     await this._ocppSender.sendCall({
                       ocppConnectionName,
                       tenantId,
-                      protocol: OCPPVersion.OCPP2_1,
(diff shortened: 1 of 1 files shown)


yuvenus (approved):

sjh37:

PR #894, opened 2026-08-19
fix(smart-charging): convert TxProfile schedule duration to seconds

vicky-ahmed:


packages/core/src/modules/SmartCharging/src/module/smartCharging/InternalSmartCharging.ts
```diff
@@ -119,7 +119,9 @@ export class InternalSmartCharging implements ISmartCharging {
       ? new Date(chargingNeeds.departureTime)
       : undefined;
     const currentTime = new Date();
-    const duration = departureTime ? departureTime.getTime() - currentTime.getTime() : undefined;
+    const duration = departureTime
+      ? Math.floor((departureTime.getTime() - currentTime.getTime()) / 1000)
+      : undefined;
```
yuvenus, line 124: ah okay interesting, is this conversion needed bc departureTime and currentTime are in milliseconds, so / 1000 will be in seconds?

vicky-ahmed: Yep, that's exactly the fix the ticket is about. The value was being sent in milliseconds, so it was like 1000x too large. But now it fits as OCPP requires.

yuvenus (approved):

PR #898, opened 2026-08-20
fix(ocpi): correct the transaction filters behind Sessions and CDRs

sjh37:
Two defects in the same `where` clause, both shaping what a polling client actually receives.

**`date_to` was inclusive.** Both services applied it as `_lte`. OCPI defines `date_from` as inclusive and `date_to` as exclusive, so a record whose `last_updated` lands exactly on the boundary is returned by the poll that ends there *and* by the poll that starts there. For a client that pages by advancing its cursor to the previous `date_to`, that is a duplicated CDR. Worth noting the implementation these services replaced had it right - the commented-out `TransactionQueryBuilder.addDateFilters` uses `Op.lt`.

**"Ended" was inferred from delivered energy.** `CdrsService` did not filter on completion at all and relied on `CdrMapper` dropping active transactions *after* the query (`CdrMapper.ts:222`). So `Transactions_aggregate` - which feeds `X-Total-Count` - counted sessions that never become CDRs, and every page came back short of its limit. `SessionsService` used `totalKwh > 0` for `endedOnly`, which is not a completion signal in either direction: a session still running has delivered energy, and a session that ended without delivering any has still ended. `isActive` is the actual flag and is already what `CdrMapper` uses.

No records were being skipped by the paging itself - the offset advances over transactions - but the total is wrong and pages are short.

Also adds `_lt`, `isActive` and `Boolean_Comparison_Exp` to the hand-maintained Hasura type stubs in `operations.ts`, and types `dateFilters` as `Timestamptz_Comparison_Exp`. It was declared `any`, which is how an undeclared `_lte` compiled in the first place. `columns: '*'` on the Transactions select permission confirms Hasura will accept filtering on `isActive`.

Adds `packages/ocpi-base/test/services/` - the third test file in this package.

### DIFF
--- a/packages/ocpi-base/src/services/CdrsService.ts
+++ b/packages/ocpi-base/src/services/CdrsService.ts
@@ -48,14 +48,10 @@ export class CdrsService {
           partyId: { _eq: fromPartyId },
         },
       },
-      // A CDR only exists for a finished session, and one that lost its active flag without ever
-      // being stopped has no end_date_time to carry. CdrMapper drops both, so the aggregate count
-      // behind X-Total-Count has to drop them as well.
       isActive: { _eq: false },
       endTime: { _is_null: false },
     };
     const dateFilters: Timestamptz_Comparison_Exp = {};
-    // OCPI defines date_from as inclusive and date_to as exclusive.
     if (dateFrom) dateFilters._gte = dateFrom.toISOString();
     if (dateTo) dateFilters._lt = dateTo.toISOString();
     if (Object.keys(dateFilters).length > 0) {
--- a/packages/ocpi-base/src/services/CdrsService.ts
+++ b/packages/ocpi-base/src/services/CdrsService.ts
@@ -12,6 +12,7 @@ import { OcpiResponseStatusCode } from '../model/OcpiResponse.js';
 import type {
   GetTransactionsQueryResult,
   GetTransactionsQueryVariables,
+  Timestamptz_Comparison_Exp,
   Transactions_Bool_Exp,
 } from '../graphql/index.js';
 import { GET_TRANSACTIONS_QUERY, OcpiGraphqlClient } from '../graphql/index.js';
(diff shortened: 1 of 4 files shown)


packages/ocpi-base/src/services/SessionsService.ts
```diff
@@ -49,15 +50,21 @@ export class SessionsService {
...
     if (dateFrom) dateFilters._gte = dateFrom.toISOString();
-    if (dateTo) dateFilters._lte = dateTo.toISOString();
+    if (dateTo) dateFilters._lt = dateTo.toISOString();
     if (Object.keys(dateFilters).length > 0) {
       where.updatedAt = dateFilters;
     }
 
     if (endedOnly) {
-      where.totalKwh = { _gt: 0 };
+      // Delivered energy is not a completion signal: a session still running has some, and one
```
yuvenus: Since this comment will reference old code/logic that will stop existing, please remove it to avoid confusion 👍

packages/ocpi-base/src/services/CdrsService.ts
```diff
@@ -47,10 +48,17 @@ export class CdrsService {
...
         },
       },
+      // A CDR only exists for a finished session. CdrMapper already drops active transactions
+      // after the fact, but the query and its aggregate count did not, so X-Total-Count counted
+      // sessions that never become CDRs and every page came back short of its limit.
+      isActive: { _eq: false },
     };
-    const dateFilters: any = {};
+    const dateFilters: Timestamptz_Comparison_Exp = {};
+    // OCPI defines date_from as inclusive and date_to as exclusive. _lte made the upper bound
```
yuvenus: Ditto, the reference to _lte when it'll stop existing may be confusing for people; I think it makes sense to leave the comment to just `// OCPI defines date_from as inclusive and date_to as exclusive` for folks that may be wondering why the range is like this. 👍

yuvenus (approved):

sjh37:
Pushed one more filter to this branch, for the same reason as the `isActive` one already here.

`deactivateActiveTransactionsByStationIdAndEvseId` clears `isActive` on a stale transaction when a new one starts on the same EVSE, and sets no `endTime`:

```ts
return await this.transaction.updateAllByQuery(
  tenantId,
  { isActive: false },
  { where: { id: { [Op.in]: ids } } },
);
```

`CdrMapper` cannot build a CDR without an `end_date_time`, so those transactions are dropped by the mapper in #904 - which puts `total` out of step with `data` unless the aggregate drops them too. Hence `endTime: { _is_null: false }` alongside the existing filter, with a test for it.

sjh37:
Done, and thanks - agreed on both.

- `SessionsService` `endedOnly`: comment removed entirely.
- Both date comments trimmed to exactly what you suggested: `// OCPI defines date_from as inclusive and date_to as exclusive.`
- I gave the `isActive`/`endTime` comment the same treatment while I was in there, so it now says why those rows are excluded rather than what the query used to do.

packages/ocpi-base/src/services/SessionsService.ts
sjh37: Removed entirely - the `endedOnly` branch is one line and the test name carries the reasoning.

packages/ocpi-base/src/services/CdrsService.ts
sjh37: Trimmed to exactly what you suggested:

```ts
// OCPI defines date_from as inclusive and date_to as exclusive.
```

I gave the `isActive` / `endTime` comment the same treatment while I was in the file, so it now says why those rows are excluded rather than what the query used to do.

One thing worth knowing since it came out of this review: the same `_lte` was still on `LocationsService` and `TariffsService`, which this PR did not touch. That is #922, and it types both filter objects as `Timestamptz_Comparison_Exp` instead of `any` - the `any` is what let `_lte` and `_lt` diverge across the four senders without the compiler noticing.

PR #899, opened 2026-08-20
fix(ocpi): resolve the OCPP 1.6 connector for StartSession instead of using the OCPI id

sjh37:
OCPI addresses a connector by its database id. OCPP 1.6 addresses it by `Connector.connectorId` — which the model itself documents as *"the serial int starting at 1 used in OCPP 1.6 to refer to the connector, unique per Charging Station"*. They are different numbers.

`sendUnlockConnectorCommand` already resolves one to the other:

```ts
const ocpp1_6ConnectorId = chargingStation.connectors.find(
  (c) => c.id === Number(unlockConnector.connector_id),
)?.connectorId;
```

`sendStartSessionCommand` did not — it passed the OCPI value straight through:

```ts
connectorId: Number(startSession.connector_id),
```

So a `RemoteStartTransaction` is aimed at whatever connector happens to carry that number, or at one the station does not have. The two agree only for a connector whose row id equals its position on the station, which is the first connector in a fresh database and increasingly unlikely after that.

This resolves the connector the same way unlock does, and reports `FAILED` on the command callback rather than sending a guess when the connector is not on the station. Both commands now share the lookup and the failure report.

The 2.0.1 handler is unaffected — it maps `evse_uid` through `EXTRACT_EVSE_ID`. So two of the three sites were already correct and this brings the third into line.

One note on the test, which may be useful beyond this PR: loading `OCPP1_6_CommandHandler` under vitest needs the package barrel, the graphql client and typedi itself stubbed, because `BaseClientApi` and `OCPPCommandHandler` use bare `@Inject()` property decorators that resolve their type from `design:type` metadata — and esbuild, which vitest uses to transform, does not emit it. The decorators throw as the class is defined, before any test runs. That is, I think, a large part of why `packages/ocpi-base` has so few tests, and it would be worth fixing centrally (a SWC transform, or explicit `@Inject(Token)` on those properties) rather than each test working around it as this one does.


packages/ocpi-base/src/util/ocppCommandHandlers/OCPP1_6_CommandHandler.ts
```diff
@@ -40,8 +40,19 @@ export class OCPP1_6_CommandHandler extends OCPPCommandHandler {
       this.config.commands.ocpiBaseUrl +
       `/2.2.1/commands/callback/${tenantPartner.id}/${this.supportedVersion}/${CommandType.START_SESSION}/${commandId}`;
     options.queryParameters = queryParameters;
+    const connectorId = this.resolveOcpp16ConnectorId(chargingStation, startSession.connector_id);
+    if (connectorId === undefined) {
```
lydiazcheng: connectorId is optional in start session command. Undefined connectorId should not trigger the not found result if it is not given in original StartSession command.

lydiazcheng (changes requested):

packages/ocpi-base/src/util/ocppCommandHandlers/OCPP1_6_CommandHandler.ts
sjh37: Good catch, thank you - fixed in 90be5637.

`connector_id` is `.nullable().optional()` on `StartSessionSchema`, and `connectorId` is `number | null | undefined` on `RemoteStartTransactionRequest`, so an omitted connector is valid at both ends and the station is the right place to choose. It is now passed through as omitted, and only a connector that was named and is not on the station reports FAILED.

`UnlockConnector` is unchanged, since OCPI requires `connector_id` there.

Added two cases covering `undefined` and `null`, asserting the request goes out as `{ idTag }` with no `connectorId` and no command result is posted.

lydiazcheng (approved):

PR #902, opened 2026-08-20
fix(ocpi): read the real charging current, not the neutral conductor

sjh37:
> Rebased onto #857, which fixed the energy half of this independently. What is left is the
> `CURRENT` dimension, which #857 did not touch.

## CURRENT was reading the neutral conductor

```ts
case MeasurandEnum['Current.Import']:
  if (sampledValue.phase === 'N') {
    cdrDimensions.push({ type: CdrDimensionType.CURRENT, volume: Number(sampledValue.value) });
  }
```

Phase `N` is the neutral. On a balanced three-phase supply it carries the imbalance, which is near
zero, and OCPI's `CURRENT` is the current over all phases - not the neutral. So an AC session
reports roughly zero amps.

DC carries no phase at all, so the condition never matches and **a DC session never produces a
`CURRENT` dimension**.

It now reads the unphased total, which is exactly what the energy measurand in the same `switch`
was already doing.

A station that reports only per-phase currents and no total still yields no `CURRENT` dimension.
Summing those is a separate change and is not attempted here.

## Tests

`SessionMapperChargingPeriods.test.ts` is new and is the only coverage of `getChargingPeriods`. It
covers the `CURRENT` fix and, since #857 added no tests, the energy dimensions that PR fixed:

- `ENERGY_IMPORT` in kWh for a meter reporting Wh
- an absent `unitOfMeasure` treated as Wh, the OCPP default
- a kWh reading passed through unchanged
- `unitOfMeasure.multiplier` applied
- the per-period `ENERGY` delta in kWh
- the periods summing to `MeterValueUtils.getTotalKwh` for the same meter values, which is the
  invariant that matters: `cdr.total_energy` comes from that, so a CDR has to reconcile against its
  own charging periods
- `CURRENT` present for a DC sample
- `CURRENT` not taken from the neutral

The two `CURRENT` cases were confirmed failing against the rebased branch with the condition put
back:

```
× reports the charging current of a DC session, which carries no phase
  → expected undefined to be 350
× does not report the neutral conductor current as the charging current
  → expected 0.4 not to be 0.4
```

## Verification

```
npx vitest run test/mapper/SessionMapperChargingPeriods.test.ts   # 8 passed
npx tsc --noEmit -p packages/ocpi-base/tsconfig.json              # clean
npx prettier --check                                              # clean
```

### DIFF
--- a/packages/ocpi-base/src/mapper/SessionMapper.ts
+++ b/packages/ocpi-base/src/mapper/SessionMapper.ts
@@ -375,7 +375,7 @@ export class SessionMapper extends BaseTransactionMapper {
     for (const sampledValue of meterValue.sampledValue) {
       switch (sampledValue.measurand) {
         case MeasurandEnum['Current.Import']:
-          if (sampledValue.phase === 'N') {
+          if (!sampledValue.phase) {
             cdrDimensions.push({
               type: CdrDimensionType.CURRENT,
               volume: Number(sampledValue.value),
--- a/packages/ocpi-base/test/mapper/SessionMapperChargingPeriods.test.ts
+++ b/packages/ocpi-base/test/mapper/SessionMapperChargingPeriods.test.ts
@@ -1,0 +1,142 @@
+// SPDX-FileCopyrightText: 2026 Contributors to the CitrineOS Project
+//
+// SPDX-License-Identifier: Apache-2.0
+import 'reflect-metadata';
+import type { MeterValueDto, SampledValue } from '@citrineos/types';
+import { MeasurandEnum } from '@citrineos/types';
+import { MeterValueUtils } from '@citrineos/base';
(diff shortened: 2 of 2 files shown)


sjh37:
Rebased onto `next` now that #857 has landed.

#857 fixed the energy dimensions the same way this PR did - `MeterValueUtils.normalizeToKwh` made public and applied to both `ENERGY_IMPORT` and the delta - so that half is dropped and only the `CURRENT` fix remains. #857 left `if (sampledValue.phase === 'N')` in place, and that is the part that matters most to us: our estate is DC, which carries no phase, so no session of ours has ever produced a CURRENT dimension.

I have kept the test file covering both, since #857 added none. If you would rather the energy cases lived with #857's change instead, say so and I will strip them back to the two CURRENT ones.

thanaParis (approved):

PR #904, opened 2026-08-20
fix(ocpi): stop CdrMapper emitting CDRs that fail its own schema

sjh37:
Three defects in `CdrMapper`, each producing a CDR that `CdrSchema` - and OCPI 2.2.1 - rejects.
`CdrMapper.test.ts` is new and is the first end-to-end coverage of this mapper; every case below
was confirmed failing against the unpatched code first.

## 1. An inactive transaction is not necessarily a finished one

```ts
private getCompletedTransactions(transactions: TransactionDto[]): TransactionDto[] {
  return transactions.filter((transaction) => !transaction.isActive);
}
```

`SequelizeTransactionEventRepository.deactivateActiveTransactionsByStationIdAndEvseId` clears
`isActive` on any stale transaction still open on an EVSE when a new one starts there:

```ts
return await this.transaction.updateAllByQuery(
  tenantId,
  { isActive: false },
  { where: { id: { [Op.in]: ids } } },
);
```

No `endTime` - unlike the stop paths, which set `{ endTime: timestamp, isActive: false }` together.
So an abandoned session reaches the mapper looking finished and produces `end_date_time: null` with
`total_time: 0`, while still charging for the energy delivered.

`getCompletedTransactions` now requires an end time as well. `CdrBroadcaster.broadcastPostCdr`
already logs `No CDRs generated for Transaction` and returns when the mapper yields nothing, so the
abandoned session surfaces in the log instead of going to a partner malformed.

## 2. charging_periods could be empty

`CdrSchema` has `charging_periods: z.array(ChargingPeriodSchema).min(1)`, but the mapper did
`session.charging_periods || []`. A station can deliver a whole session without sending a meter
value - `MeterValueSampleInterval` at 0 and no `transactionData` on the 1.6 stop - and then there
are no periods at all.

Rather than emit none, the session is described as one period carrying the totals the CDR already
states: `ENERGY` = `total_energy`, `TIME` = `total_time`. Nothing is invented; it restates the same
figures at the granularity OCPI requires.

## 3. tariffs could be [null]

```ts
tariffs: [ocpiTariff],
```

`mapSessionsToCDRs` filters on `transactionIdToTariffMap` but reads the value out of
`transactionIdToOcpiTariffMap` - a different map, filled by a second `GET_TARIFF_BY_KEY_QUERY` that
can return nothing while the first succeeded from the tariff already on the transaction row. The
`!` then produced `tariffs: [undefined]`, serialised as `[null]`, which no `Tariff` consumer can
parse. `tariffs` is optional, so it is omitted instead.

## Note on the aggregate count

`CdrsService` counts transactions with a `Transactions_aggregate` and maps the page separately, so
anything the mapper drops makes `total` disagree with `data`. The matching query-side predicate for
case 1 belongs with the `isActive` filter in #898 and I have pushed it there rather than opening the
same object literal in two PRs.

## Verification

```
npx vitest run                 # packages/ocpi-base: 2 files, 12 tests passed
npx tsc --noEmit -p packages/ocpi-base/tsconfig.json    # clean
npx prettier --check / npx eslint <changed files>       # clean
```

### DIFF
--- a/packages/ocpi-base/src/mapper/CdrMapper.ts
+++ b/packages/ocpi-base/src/mapper/CdrMapper.ts
@@ -218,7 +218,12 @@ export class CdrMapper extends BaseTransactionMapper {
     return undefined;
   }
 
+  /**
+   * A CDR requires an end_date_time, so a transaction that lost its active flag without ever being
+   * stopped is not billable. That happens when a new transaction starts on an EVSE that still had
+   * a stale one open.
+   */
   private getCompletedTransactions(transactions: TransactionDto[]): TransactionDto[] {
-    return transactions.filter((transaction) => !transaction.isActive);
+    return transactions.filter((transaction) => !transaction.isActive && !!transaction.endTime);
   }
 }
--- a/packages/ocpi-base/src/mapper/CdrMapper.ts
+++ b/packages/ocpi-base/src/mapper/CdrMapper.ts
@@ -78,9 +78,6 @@ export class CdrMapper extends BaseTransactionMapper {
   ): Promise<Cdr[]> {
     return Promise.all(
       sessions
-        // A CDR needs a tariff and at least one charging period. A session missing either cannot
-        // be described faithfully, so it is left out rather than filled in; the caller reports
-        // which sessions were dropped.
         .filter(
           (session): session is Session & { charging_periods: ChargingPeriod[] } =>
             transactionIdToTariffMap.has(session.id) && !!session.charging_periods?.length,
@@ -226,11 +223,6 @@ export class CdrMapper extends BaseTransactionMapper {
     return undefined;
   }
 
-  /**
-   * A CDR requires an end_date_time, so a transaction that lost its active flag without ever being
-   * stopped is not billable. That happens when a new transaction starts on an EVSE that still had
-   * a stale one open.
-   */
   private getCompletedTransactions(transactions: TransactionDto[]): TransactionDto[] {
     return transactions.filter((transaction) => !transaction.isActive && !!transaction.endTime);
   }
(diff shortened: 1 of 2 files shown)


packages/ocpi-base/src/mapper/CdrMapper.ts
```diff
@@ -131,6 +133,27 @@ export class CdrMapper extends BaseTransactionMapper {
...
+  /**
+   * A CDR needs at least one charging period, and a station can deliver a whole session without
+   * ever sending a meter value. The session is then described as one period carrying the totals
+   * the CDR already states, rather than as none at all.
+   */
+  private getChargingPeriods(session: Session, tariff: TariffDto): ChargingPeriod[] {
+    if (session.charging_periods?.length) {
+      return session.charging_periods;
+    }
+    return [
```
lydiazcheng: I don't think creating an additional charging period manually is a correct behavior here 🤔 , mapping CDR should fails in this case. CC: @thanaParis

sjh37: Agreed, and changed in 442ac4f3 - the synthesis is gone and the session is dropped instead.

You are right that it reads as measured detail that was never measured. The period I was building restated `total_energy` and `total_time`, so it added no information, but a partner receiving it cannot tell that apart from a real single-period session - and a CDR that looks complete is worse than a missing one.

Dropping it also fits what the mapper already does: `mapSessionsToCDRs` filters out sessions with no tariff, so a session with no charging periods is now filtered on the same line for the same reason. With #906 the caller reports which sessions were dropped and why, so this is visible rather than silent.

The rest of the PR is unchanged: `getCompletedTransactions` still requires an `endTime` so no CDR goes out with a null `end_date_time`, and `tariffs` is omitted rather than set to `[null]`.

thanaParis (approved):

PR #906, opened 2026-08-20
fix(ocpi): say which transactions were dropped from a Sessions or CDRs page

sjh37:
## A short page with no explanation

`mapTransactionsToSessionsHelper` drops any transaction whose location, token or tariff is missing:

```ts
if (location && token && tariff) {
  result.push(this.mapTransactionWithContextToSession(transaction, location, token, tariff));
} else {
  this.logger.debug(`Skipped transaction ${transaction.transactionId}`);
}
```

`CdrsService` and `SessionsService` do not count what the mapper returns - `total` comes from a
separate `Transactions_aggregate` over the same `where`:

```ts
const response = buildOcpiPaginatedResponse(
  OcpiResponseStatusCode.GenericSuccessCode,
  result.Transactions_aggregate?.aggregate?.count ?? 0,
  limit, offset, mappedCdr,
);
```

So a client paging by `offset` against `total` gets fewer records than it was told exist and has no
way to discover which. For a CDR poll that is missing billing data, arriving as silence.

The most likely cause in practice is the token: `TokensMapper.getContractId` throws for any
authorization without an eMAID in `additionalInfo`, which is every token not created through OCPI.

One `warn` per page now names the transactions and what each was missing, instead of a `debug` line
per transaction that named neither:

```
Skipped 2 of 50 transactions: tx-41 (no token); tx-49 (no location, tariff)
```

It stays quiet when everything maps.

## Verification

```
npx vitest run                 # packages/ocpi-base: 2 files, 10 tests passed
npx tsc --noEmit -p packages/ocpi-base/tsconfig.json    # clean
npx prettier --check / npx eslint <changed files>       # clean
```

Confirmed failing first with `expected "warn" to be called once, but got 0 times`.

### DIFF
--- a/packages/ocpi-base/src/mapper/SessionMapper.ts
+++ b/packages/ocpi-base/src/mapper/SessionMapper.ts
@@ -119,6 +119,7 @@ export class SessionMapper extends BaseTransactionMapper {
     transactionIdToTariffMap: Map<string, TariffDto>,
   ): Promise<Session[]> {
     const result: Session[] = [];
+    const skipped: string[] = [];
     for (const transaction of transactions) {
       const location = transactionIdToLocationMap.get(transaction.transactionId!);
       const token = transactionIdToTokenMap.get(transaction.transactionId!);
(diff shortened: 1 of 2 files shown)


lydiazcheng (approved):

vicky-ahmed (approved):
LGTM

PR #907, opened 2026-08-20
fix(core): map stored non-Accepted status in OCPP 1.6 Authorize

vicky-ahmed:


packages/core/src/dal/layers/sequelize/mapper/1.6/AuthorizationMapper.ts
```diff
@@ -15,12 +15,14 @@ export class AuthorizationMapper {
...
         return OCPP1_6.AuthorizeResponseStatus.Blocked;
+      case AuthorizationStatusEnum.ConcurrentTx:
+        return OCPP1_6.AuthorizeResponseStatus.ConcurrentTx;
       case AuthorizationStatusEnum.Expired:
         return OCPP1_6.AuthorizeResponseStatus.Expired;
       case AuthorizationStatusEnum.Invalid:
         return OCPP1_6.AuthorizeResponseStatus.Invalid;
       default:
-        throw new Error('Unknown IdTagInfoStatus status');
+        return OCPP1_6.AuthorizeResponseStatus.Invalid;
```
lydiazcheng, line 25: Why do we replace this with Invalid? In this case the error will be ignored, dose it fit for all existing use cases?

lydiazcheng (commented):
The changes looks good. Only one question.

packages/core/src/dal/layers/sequelize/mapper/1.6/AuthorizationMapper.ts
vicky-ahmed: Invalid is correct because unsupported authorization statuses have no OCPP 1.6 Authorize.conf representation. Its only caller is AuthorizeRequestOcpp16Handler, which wraps in try/catch already. So this works for all existing use cases.

vicky-ahmed: @thanaParis @masonseaman thoughts?

yuvenus:
Should resolve https://github.com/citrineos/citrineos/issues/215

packages/core/src/dal/layers/sequelize/mapper/1.6/AuthorizationMapper.ts
thanaParis: I think invalid is fine, except we need a warn or error log here to surface this

thanaParis (changes requested):

thanaParis (approved):

PR #908, opened 2026-08-20
OCPI awilix conversion

devonNewsomDevelop:
Converts OCPI to use Awilix for DI
Moves the container for OCPP out of app server and into Core


apps/ocpi-server/package.json
```diff
@@ -46,6 +46,7 @@
     "@citrineos/types": "workspace:*",
     "@directus/extensions": "~1.0.2",
     "@fastify/type-provider-json-schema-to-ts": "4.0.1",
+    "awilix": "13.0.3",
```
yuvenus, line 49: Perhaps worth moving awilix into the pnpm-workspace and then referencing it with "catalog:" in all package.jsons so that the versions are synced up 👀

yuvenus (approved):
Builds and runs on my local, looking pretty good so far 🤞🔥

apps/ocpi-server/package.json
devonNewsomDevelop: Yeah that's a great idea

PR #911, opened 2026-08-20
fix(core): scope the GetBaseReport progress cache to the tenant

sjh37:
Follow-up to #872, which @yuvenus suggested keeping as a separate PR.

## One cache entry shared by every tenant

`createGetBaseReportRequest` marks the request in progress:

```ts
const requestId = 0;
await this._cache.set(requestId.toString(), 'ongoing', ocppConnectionName, maxCachingSeconds);
```

`confirmGetBaseReportSuccess` then blocks on `onChange(requestId, ..., ocppConnectionName)` until
`NotifyReportRequestOcpp2Handler` writes `'complete'` to the same key, also under
`ocppConnectionName`.

The request id is a hard-coded 0 - the OCTT workaround noted in the comment there - so the key is
literally `"0"` in namespace `<stationName>`, with no tenant anywhere in it. Two tenants that each
have a station called `CP001` share that single entry.

The failure is a boot that completes on someone else's report: tenant A is waiting, tenant B's
station sends its `NotifyReport`, `'complete'` lands on the shared key, and A's
`confirmGetBaseReportSuccess` returns successfully. The boot handler then sets
`getBaseReportOnPending = false` and carries on with a device model it never actually received.

All four operations now use `createIdentifier(tenantId, ocppConnectionName)`, the namespace the rest
of the boot path already uses. `tenantId` is threaded into the two service methods rather than
resolved in the handler, so `BootNotificationRequestOcpp2Handler` only gains an argument on two
calls and does not collide with #872's changes in the same file.

## Verification

```
npx vitest run test/modules/Configuration test/handlers/requests/2   # 6 files, 58 tests passed
npx prettier --check / npx eslint <changed files>  # clean
```

Both new tests were confirmed failing first, with the parameter in place but the namespace left as
the bare station name, so the failure is about the namespace and nothing else:

```
× marks the request ongoing under the tenant-scoped identifier
  → expected "spy" to be called with arguments: [ '0', 'ongoing', '1:Station01', 10 ]
× waits on the tenant-scoped identifier
  → expected "spy" to be called with arguments: [ '1', 10, '1:Station01' ]
```

### DIFF
--- a/packages/core/src/handlers/requests/2/BootNotificationRequestOcpp2Handler.ts
+++ b/packages/core/src/handlers/requests/2/BootNotificationRequestOcpp2Handler.ts
@@ -192,6 +192,7 @@ export class BootNotificationRequestOcpp2Handler extends AbstractHandler {
       await this._cache.remove(OCPP_CallAction.NotifyReport, identifier);
 
       const getBaseReportRequest = await this._bootService.createGetBaseReportRequest(
+        tenantId,
         ocppConnectionName,
         this._config.maxCachingSeconds,
       );
@@ -206,6 +207,7 @@ export class BootNotificationRequestOcpp2Handler extends AbstractHandler {
       });
 
       await this._bootService.confirmGetBaseReportSuccess(
+        tenantId,
         ocppConnectionName,
         getBaseReportRequest.requestId.toString(),
         getBaseReportConfirmation,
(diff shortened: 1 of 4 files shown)


sjh37:

This branch carried a `Merge branch 'next'` commit, which was what tripped GitGuardian (it re-introduced the repo's own docker-compose files into the diff) and dragged a stale test tree in with it. Rebased into a single commit on current `next` - the merge commit is gone, the prettier error in `BootNotificationService.test.ts` with it, and `packages/core/test/modules/Configuration` is green (27).

thanaParis (approved):

PR #912, opened 2026-08-20
fixing autocommissioning of multi-evse stations

thanaParis:


lydiazcheng (approved):
Changes look good. Only few tests fail.

PR #913, opened 2026-08-21
fix(ocpi): make packages/ocpi-base loadable, so it can be tested at all

sjh37:
One test file covers 310 source files (20.5k lines) in this package. That is not a decision about
priorities - the package could not be imported.

```
CannotInjectValueError: Cannot inject value into "BaseClientApi.logger".
  ❯ typedi/src/decorators/inject.decorator.ts:23:13
  ❯ __decorateClass src/trigger/BaseClientApi.ts:22:24
  ❯ src/trigger/SessionsClientApi.ts:5:1
```

That is `await import('../src/mapper/CdrMapper.js')` in a test with no mocking. There are two
independent causes.

## 1. A parameterless @Inject() needs metadata the test transform does not emit

```ts
// typedi/utils/resolve-to-type-wrapper.util.js
if (!typeOrIdentifier && propertyName) {
  const identifier = Reflect.getMetadata('design:type', target, propertyName);
  typeWrapper = { eagerType: identifier, lazyType: () => identifier };
}
```

`tsconfig.build.json` sets `emitDecoratorMetadata`, so `tsc` emits `design:type` and the build is
fine. Vitest transforms with **esbuild**, which does not implement it. `eagerType` is `undefined`
and the decorator throws.

Decorators run when the class is **defined**, so this fails on import, not on resolution. Every
module that transitively reaches such a class becomes unloadable - **56 modules** from the
`BaseClientApi.logger` site alone, measured by reverting it and counting.

`packages/core` is unaffected because it never relies on inferred types: `@Column(DataType.INTEGER)`
rather than `@Column()`.

Fifteen bare `@Inject()` across six files now name their type:

```ts
-  @Inject()
+  @Inject(() => Logger)
   protected logger!: Logger<ILogObj>;
```

This is the form the package already uses (`@Inject(OcpiConfigToken)`) and the one typedi documents
for cyclic dependencies. It needs no metadata, so it behaves the same under tsc, esbuild, swc or a
bundler - which also removes a latent production risk, since the same failure appears the moment
this package is built with anything but `tsc`.

The arrow matters. typedi *calls* a function argument:

```ts
if (typeOrIdentifier && typeof typeOrIdentifier === 'function') {
  typeWrapper = { eagerType: null, lazyType: () => typeOrIdentifier() };
}
```

so `@Inject(Logger)` would throw "Class constructor cannot be invoked without 'new'" at resolution.
It has to be `@Inject(() => Logger)`.

Adding an SWC or Babel transform to vitest instead would also work, but it needs a new dependency,
keeps the source dependent on a compiler feature it does not need, and helps nobody bundling this
package. Happy to go that way if maintainers prefer.

## 2. Three modules import a barrel that imports them back

With the decorators fixed, a second failure appears:

```
TypeError: Class extends value undefined is not a constructor or null
  ❯ src/mapper/CdrMapper.ts:31:32   export class CdrMapper extends BaseTransactionMapper
  ❯ src/mapper/index.ts:5:1
```

`SessionMapper` imports `LocationsService`, which imports the **mapper barrel** for three mappers
that all live in `LocationMapper.js`. The barrel re-exports `CdrMapper`, which extends a
`BaseTransactionMapper` that is still initialising.

Same shape in two more places:

- `util/ocppCommandHandlers/base.ts` imports the **package root barrel** for four values and four
  types. That reaches `OCPP2_0_1_CommandHandler`, which extends `base.ts` itself.
- `broadcaster/LocationsBroadcaster.ts` imports the mapper barrel, like `LocationsService`.

Each now imports from the module that declares what it needs. That is three import statements; the
package has 60 internal barrel imports in total and untangling them all is a separate piece of work.

Note this only throws for the module the graph is entered from, which is why importing every module
in sequence passes while a test naming one of them as its first import fails.

## Tests

- **`moduleLoading.test.ts`** imports the barrel, then every source module, then each likely entry
  point on its own with a reset module registry. Reverting one decorator fails it with 56 unloadable
  modules; reverting one barrel import fails the `SessionMapper` entry-point case.
- **`containerWiring.test.ts`** resolves a service through the container and checks each injection
  landed. A lazy token naming the wrong class still *loads*, so import coverage alone would not catch
  a mistake in part 1.
- **`services/CommandsService.test.ts`** - 14 tests over the `START_SESSION` and `STOP_SESSION`
  guards. `CommandsService` held a bare `@Inject()`, so it could not be tested before: with the
  decorator reverted the file reports `no tests` because it never collects. It also confirms
  `RESERVE_NOW` and `CANCEL_RESERVATION`, whose handler bodies are commented out, report
  `NOT_SUPPORTED` rather than success.

## Known remaining gap

`LocationsBroadcaster` still cannot be the entry point of the graph: it builds a JSON schema from a
zod schema as it loads, and a further cycle leaves one referenced schema `undefined`. Nothing
imports it first today, so it is not in `ENTRY_POINTS` and is left for its own change.

## Verification

```
packages/ocpi-base: 4 files, 33 tests passed
full suite:         96 files passed / 1 skipped, 2040 passed / 5 skipped
prettier --check / eslint                # 0 errors (11 pre-existing warnings)
```

`packages/core/test/dal/e2e/security-event.e2e.test.ts` fails on this branch, and fails identically
on unmodified `next` with these changes stashed - both server instances bind port 8080, which is
#873.

On our own branch, where the tests from #891, #898, #901-#906 are already applied, this change lets
every one of them drop its `vi.mock` workaround, including the file that had to stub `typedi`
itself: 15 files, 75 tests, no mocking of the package's own modules anywhere.

### DIFF
--- a/packages/ocpi-base/src/broadcaster/LocationsBroadcaster.ts
+++ b/packages/ocpi-base/src/broadcaster/LocationsBroadcaster.ts
@@ -22,7 +22,7 @@ import {
   type TenantDto,
   HttpMethod,
 } from '@citrineos/types';
-import { ConnectorMapper, EvseMapper, LocationMapper } from '../mapper/index.js';
+import { ConnectorMapper, EvseMapper, LocationMapper } from '../mapper/LocationMapper.js';
 import { OcpiEmptyResponseSchema } from '../model/OcpiEmptyResponse.js';
 
 @Service()
--- a/packages/ocpi-base/src/modules/Commands/module/CommandsModuleApi.ts
+++ b/packages/ocpi-base/src/modules/Commands/module/CommandsModuleApi.ts
@@ -89,7 +89,7 @@ for (const [commandType, { schema, name }] of Object.entries(COMMAND_REQUEST_SCH
 @JsonController(`/:${versionIdParam}/${ModuleId.Commands}`)
 @Service()
 export class CommandsModuleApi extends BaseController implements ICommandsModuleApi {
-  @Inject()
+  @Inject(() => CommandExecutor)
   private commandsExecutor!: CommandExecutor;
 
   constructor(readonly commandsService: CommandsService) {
--- a/packages/ocpi-base/src/services/CommandsService.ts
+++ b/packages/ocpi-base/src/services/CommandsService.ts
@@ -34,13 +34,13 @@ import { EXTRACT_STATION_ID } from '../model/DTO/EvseDTO.js';
 
 @Service()
 export class CommandsService {
-  @Inject()
+  @Inject(() => Logger)
   protected logger!: Logger<ILogObj>;
 
-  @Inject()
+  @Inject(() => OcpiGraphqlClient)
   protected ocpiGraphqlClient!: OcpiGraphqlClient;
 
-  @Inject()
+  @Inject(() => CommandExecutor)
   protected commandExecutor!: CommandExecutor;
 
   @Inject(OcpiConfigToken) readonly config!: OcpiConfig;
--- a/packages/ocpi-base/src/services/LocationsService.ts
+++ b/packages/ocpi-base/src/services/LocationsService.ts
@@ -36,7 +36,7 @@ import {
   GET_LOCATIONS_QUERY,
   OcpiGraphqlClient,
 } from '../graphql/index.js';
-import { ConnectorMapper, EvseMapper, LocationMapper } from '../mapper/index.js';
+import { ConnectorMapper, EvseMapper, LocationMapper } from '../mapper/LocationMapper.js';
 import type { ChargingStationDto, ConnectorDto, EvseDto, LocationDto } from '@citrineos/types';
 
 @Service()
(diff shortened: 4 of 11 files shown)


sjh37:
Superseded by #908 - closing.

This PR gave every `@Inject` in `packages/ocpi-base` an explicit token and broke the import cycles
that stopped the package loading. #908 removed typedi from the package altogether and replaced it
with explicit constructor dependency objects (`OcpiDependencies`, `OcppCommandHandlerDependencies`,
`CommandsServiceDependencies`), which solves the same problem better - there are no `@Inject`
decorators left in `packages/ocpi-base/src` to give tokens to.

The consequence I opened this for is gone with it. `packages/ocpi-base/test` on `next` now holds 15
test files against the one it had when I wrote this, so the package is importable by a test and the
premise no longer stands.

Nothing here is worth salvaging on top of that: my two commits conflict across six files, and every
conflict is this PR adding a decorator where `next` has already deleted the decorator. #942 was
stacked on this one and has been rebased off it - it is now a single commit on `next`.

PR #914, opened 2026-08-21
fix(core): enforce the OCPP 1.6 local auth list size limits

sjh37:
## The 1.6 path forwards a list the station cannot hold

`LocalAuthListService` validates a `SendLocalList` before forwarding it. On 2.0.1 that includes two
size checks:

- `LocalAuthListCtrlr.Entries` `maxLimit` - the most the station can store (D01.FR.12)
- `LocalAuthListCtrlr.ItemsPerMessage` - the most it accepts in one message (D01.FR.11)

The 1.6 method checks the version number and duplicate idTags, and then forwards whatever it was
given. Both size checks are absent, so an oversized list goes to the charger and is refused there -
or silently truncated, depending on the vendor.

1.6 has both equivalents, as configuration keys rather than device model variables:

| 2.0.1 | 1.6 |
|---|---|
| `LocalAuthListCtrlr.Entries` maxLimit | `LocalAuthListMaxLength` |
| `LocalAuthListCtrlr.ItemsPerMessage` | `SendLocalListMaxLength` |

`GetConfigurationResponseOcpp16Handler` already persists every reported key as a
`ChangeConfiguration` row, so the values are there. Nothing read them. Before this change the only
occurrence of either key name in the repository was a doc comment on this very method:

```ts
 * Mirrors the 2.0.1 path but uses flat idTag and 1.6 update enum. Item-count limits for 1.6
 * come from the chargepoint via `LocalAuthListMaxLength` configuration; absence is non-fatal.
```

which describes a read that has never existed. That comment is corrected too.

## What changed

Both keys are now read from `ChangeConfiguration` and enforced. A key the station has not reported
is not enforced - the same posture the 2.0.1 path takes for a missing `Entries` maxLimit, with the
same warning.

Two details worth calling out:

**The checks run before persisting.** The 2.0.1 path calls
`createSendLocalListFromStationIdAndRequestAndCurrentVersion` first and then throws on a size
violation, which leaves a `SendLocalList` row behind for a request that was rejected. The 1.6 path
validates first so that does not happen here. I have not changed the 2.0.1 ordering in this PR -
happy to, in a separate one, if you agree it is a problem.

**The maximum counts the resulting list, not the request.** For `Full` that is the request length.
For `Differential` it is what the station already holds, plus the additions, less the deletes - a
1.6 entry carrying no `idTagInfo` removes that idTag, which `createSendLocalListFromRequestData16`
already treats as a tombstone.

## Tests

`LocalAuthListService16.test.ts` is new; the existing file has 12 specs and every one of them
exercises 2.0.1, so the 1.6 method had no coverage at all. Eight cases:

- a list longer than `SendLocalListMaxLength` is refused, and nothing is persisted
- a list exactly at the limit is accepted
- an unreported `SendLocalListMaxLength` is not enforced
- a `Full` list longer than `LocalAuthListMaxLength` is refused, and nothing is persisted
- a `Differential` update is counted against what the station already holds
- a `Differential` entry replacing one already held does not count as a new slot
- a `Differential` delete frees a slot
- an unreported `LocalAuthListMaxLength` is not enforced

The three refusal cases were confirmed failing first with `promise resolved "{}" instead of
rejecting`.

`LocalAuthListService` gains a `changeConfigurationRepository` dependency, already registered in the
container, so the existing 2.0.1 spec file supplies a stub for it.

## Verification

```
npx vitest run test/modules/EVDriver     # 2 files, 20 tests passed
full suite:                              # 98 files passed / 1 skipped, 2050 passed / 4 skipped
prettier --check / eslint                # clean
```

### DIFF
--- a/packages/core/src/modules/EVDriver/src/module/LocalAuthListService.ts
+++ b/packages/core/src/modules/EVDriver/src/module/LocalAuthListService.ts
@@ -6,6 +6,7 @@ import type { ILogObj } from 'tslog';
 import { Logger } from 'tslog';
 import { v4 as uuidv4 } from 'uuid';
 import type {
+  IChangeConfigurationRepository,
   IDeviceModelRepository,
   ILocalAuthListRepository,
 } from '@dal/interfaces/repositories.js';
@@ -20,19 +21,23 @@ import {
 export class LocalAuthListService {
   protected _localAuthListRepository: ILocalAuthListRepository;
   protected _deviceModelRepository: IDeviceModelRepository;
+  protected _changeConfigurationRepository: IChangeConfigurationRepository;
   protected _logger: Logger<ILogObj>;
 
   constructor({
     localAuthListRepository,
     deviceModelRepository,
+    changeConfigurationRepository,
     logger,
   }: {
     localAuthListRepository: ILocalAuthListRepository;
     deviceModelRepository: IDeviceModelRepository;
+    changeConfigurationRepository: IChangeConfigurationRepository;
     logger: Logger<ILogObj>;
   }) {
     this._localAuthListRepository = localAuthListRepository;
     this._deviceModelRepository = deviceModelRepository;
+    this._changeConfigurationRepository = changeConfigurationRepository;
     this._logger = logger
       ? logger.getSubLogger({ name: this.constructor.name })
       : new Logger<ILogObj>({ name: this.constructor.name });
@@ -221,8 +226,9 @@ export class LocalAuthListService {
 
   /**
    * OCPP 1.6 variant: validate and persist a SendLocalListRequest, returning the persisted row.
-   * Mirrors the 2.0.1 path but uses flat idTag and 1.6 update enum. Item-count limits for 1.6
-   * come from the chargepoint via `LocalAuthListMaxLength` configuration; absence is non-fatal.
+   * Mirrors the 2.0.1 path but uses flat idTag and the 1.6 update enum. The size limits come from
+   * the SendLocalListMaxLength and LocalAuthListMaxLength configuration keys rather than from the
+   * device model, and a key the station has not reported is not enforced.
    */
   async persistSendLocalListForStationIdAndCorrelationIdAndSendLocalListRequest16(
     tenantId: number,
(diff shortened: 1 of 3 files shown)


packages/core/src/modules/EVDriver/src/module/LocalAuthListService.ts
```diff
@@ -255,6 +255,39 @@ export class LocalAuthListService {
       }
     }
 
+    const sendLocalListMaxLength = await this.getOcpp16ConfigurationNumber(
```
yuvenus, line 258: Fascinating! Can you point the exact part of the spec that specifies the following information:

1. That these configuration keys are the one a station should be using to report this information
2. What the behavior should be if the max length doesn't exist
3. If the behavior in countUpdatedAuthList16 aligns with how the spec specifies it

sjh37: Good questions - taking them in order, and flagging where I am citing the spec from outside the repo rather than something you can check here.

**1. That these are the keys a station reports this on**

OCPP 1.6 Edition 2, §9.1 *Standard Configuration Key Names & Values*, under the Local Auth List Management profile:

| Key | Access | Required | Description |
|---|---|---|---|
| `LocalAuthListEnabled` | RW | Required | Whether the Local Authorization List is enabled |
| `LocalAuthListMaxLength` | RO | Required | Maximum number of identifications that can be stored in the Local Authorization List |
| `SendLocalListMaxLength` | RO | Required | Maximum number of identifications that can be sent in a single SendLocalList.req |

`SendLocalListMaxLength` is also named from the message definition itself - §6.39 `SendLocalList.req`, on the `localAuthorizationList` field: *"Maximum number of AuthorizationData elements is available in the configuration key: SendLocalListMaxLength"*. That is the strongest form of the answer for that one: the spec points at the key from the field it constrains.

Both are Required for the profile, so a station that supports Local Auth List Management reports them. In this repo they arrive via `GetConfigurationResponseOcpp16Handler`, which stores every returned key as a `ChangeConfiguration` row - that is the table `getOcpp16ConfigurationNumber` reads. A `GetConfiguration.req` with no `key` list returns all keys, which is what the boot flow sends when `getConfigurationsOnPending` is set.

There is no 1.6 configuration-key enum in the repo to bind these to, so they are string literals. Happy to add one if you would prefer that.

**2. Behaviour when the max length is absent**

Absent means *the CSMS has not been told*, not *the station has no limit* - the row only exists once a `GetConfiguration` response has been stored, so a station added without a boot fetch has neither key. The CSMS must not invent a limit, and it does not need to: the station is the authority and answers `SendLocalList.conf` with `Failed`. So the check is skipped and the request is forwarded.

You caught an inconsistency in doing so: `LocalAuthListMaxLength` logged that the check had not run, `SendLocalListMaxLength` skipped silently. Fixed in 0635bbbd - both now warn and forward, so an operator whose 5000-entry list comes back `Failed` can see from the CSMS logs that no length check ran.

**3. Whether `countUpdatedAuthList16` matches the spec**

§6.4 `AuthorizationData`, on `idTagInfo`: *"Optional. (Required when UpdateType is Full) ... For a Differential update the following applies: If this element is present, then this entry SHALL be added or updated in the Local Authorization List. If this element is absent, then this entry SHALL be deleted from the Local Authorization List."*

So the resulting list size is:

- **Full** - exactly `localAuthorizationList.length`, since that list *is* the new list and `idTagInfo` is required on every entry.
- **Differential** - start from what the station already holds, then per entry: `idTagInfo` present adds or replaces, absent deletes. Replacing an existing entry does not grow the list; deleting frees a slot.

That is what the method does, and it is the same predicate `createSendLocalListFromRequestData16` already uses to decide whether an entry is a tombstone, so the count matches what is actually persisted. The `if (!entry.idTag) continue` mirrors that method skipping such entries, so a malformed entry is not counted against a limit it will never occupy. Each of those cases has a test in `LocalAuthListService16.test.ts`.

One asymmetry I did **not** change here, since it is outside this fix: the 2.0.1 sibling `countUpdatedAuthListFromRequestAndCurrentVersion` unions the request and the stored list by `authorizationId` and never removes anything, so a differential delete counts as an addition there. 2.0.1 §`SendLocalListRequest` has the same absent-`idTokenInfo`-means-delete rule, so I think that one is wrong in the same way. Say the word and I will raise it separately.

thanaParis (approved):

PR #919, opened 2026-08-21
fix(core): apply the BytesPerMessage limit in SetVariables and GetMonitoringReport

sjh37:
## The defect

Both endpoints batch by `ItemsPerMessage` but never check `BytesPerMessage`, so a request under the item limit still goes out when it is over the station's byte limit. OCPP defines both constraints; only one was applied.

`ClearVariableMonitoringEndpoint` and `GetVariablesEndpoint` already check both, so this brings the two that did not into line with them.

## The fix

Each endpoint reads `BytesPerMessage` from the same component and instance its `ItemsPerMessage` read already used:

| Endpoint | Component | Instance |
| --- | --- | --- |
| `SetVariables` | `DeviceDataCtrlr` | `SetVariables` |
| `GetMonitoringReport` | `DeviceDataCtrlr` | `GetReport` |

`GetMonitoringReport` sharing the `GetReport` variable is per spec, not an oversight: `ItemsPerMessageGetReport` (2.1.17) and `BytesPerMessageGetReport` (2.1.19) are each defined as constraining `GetReportRequest` **or** `GetMonitoringReportRequest`.

An unset limit means unlimited, as elsewhere, so a station that does not report the variable is unaffected.

## Changes since review

@yuvenus caught that an earlier revision of this PR pointed `GetMonitoringReport` at `MonitoringCtrlr`/`GetMonitoringReport`. That was wrong - thank you. Reverted in 8137ec8f, and the test that asserted it now asserts the spec behaviour instead.

## Verification

```
 ✓ packages/core/test/modules/Api/endpoints/ocpp/reportingTransactionsEndpoints.test.ts (17 tests)
 ✓ packages/core/test/modules/Api/endpoints/ocpp/monitoringEndpoints.test.ts
```

### DIFF
--- a/packages/core/src/modules/Api/src/module/endpoints/ocpp/2/reporting/GetMonitoringReportEndpoint.ts
+++ b/packages/core/src/modules/Api/src/module/endpoints/ocpp/2/reporting/GetMonitoringReportEndpoint.ts
@@ -18,7 +18,7 @@ import {
 } from '@citrineos/types';
 import { getBatches, getSizeOfRequest } from '@util/index.js';
 import type { DeviceModelService } from '@util/deviceModel/DeviceModelService.js';
-import { COMPONENT_MONITORING_CTRLR } from '../components.js';
+import { COMPONENT_DEVICE_DATA_CTRLR } from '../components.js';
 import { OCPP2_PROTOCOLS, ocpp2Schema } from '../schemas.js';
 
 interface Dependencies extends AbstractMessageEndpointDependencies {
@@ -64,8 +64,8 @@ export class GetMonitoringReportEndpoint extends AbstractMessageEndpoint {
     for (const ocppConnectionName of identifiers) {
       const maxBytes =
         await this._deviceModelService.getBytesPerMessageByComponentAndVariableInstanceAndStationId(
-          COMPONENT_MONITORING_CTRLR,
-          OCPP_CallAction.GetMonitoringReport,
+          COMPONENT_DEVICE_DATA_CTRLR,
+          OCPP_CallAction.GetReport,
           tenantId,
           ocppConnectionName,
         );
@@ -79,8 +79,8 @@ export class GetMonitoringReportEndpoint extends AbstractMessageEndpoint {
 
       const itemsPerMessageGetReport =
         (await this._deviceModelService.getItemsPerMessageByComponentAndVariableInstanceAndStationId(
-          COMPONENT_MONITORING_CTRLR,
-          OCPP_CallAction.GetMonitoringReport,
+          COMPONENT_DEVICE_DATA_CTRLR,
+          OCPP_CallAction.GetReport,
           tenantId,
           ocppConnectionName,
         )) ?? componentVariable.length;
--- a/packages/core/src/modules/Api/src/module/endpoints/ocpp/2/reporting/GetMonitoringReportEndpoint.ts
+++ b/packages/core/src/modules/Api/src/module/endpoints/ocpp/2/reporting/GetMonitoringReportEndpoint.ts
@@ -16,9 +16,9 @@ import {
   type OCPPVersion,
   type OCPP2_request_types,
 } from '@citrineos/types';
-import { getBatches } from '@util/index.js';
+import { getBatches, getSizeOfRequest } from '@util/index.js';
 import type { DeviceModelService } from '@util/deviceModel/DeviceModelService.js';
-import { COMPONENT_DEVICE_DATA_CTRLR } from '../components.js';
+import { COMPONENT_MONITORING_CTRLR } from '../components.js';
 import { OCPP2_PROTOCOLS, ocpp2Schema } from '../schemas.js';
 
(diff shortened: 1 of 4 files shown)


packages/core/src/modules/Api/src/module/endpoints/ocpp/2/reporting/GetMonitoringReportEndpoint.ts
```diff
@@ -62,10 +62,25 @@ export class GetMonitoringReportEndpoint extends AbstractMessageEndpoint {
 
     const confirmations: IMessageConfirmation[] = [];
     for (const ocppConnectionName of identifiers) {
+      const maxBytes =
+        await this._deviceModelService.getBytesPerMessageByComponentAndVariableInstanceAndStationId(
+          COMPONENT_MONITORING_CTRLR,
+          OCPP_CallAction.GetMonitoringReport,
```
yuvenus: As per the OCPP 2.0.1 spec (referring to OCPP-2.1_edition2_part2_specification.pdf, section 2.1.19. BytesPerMessageGetReport):

Image

If the image isn't loading: 

**Component:** DeviceDataCtrlr
**VariableName**: BytesPerMessage
**VariableInstance**: GetReport
**Description**: Message Size (in bytes) - puts constraint on GetReportRequest or
GetMonitoringReportRequest message size.

So the component is in fact `COMPONENT_DEVICE_DATA_CTRLR` and the call action should be `OCPP_CallAction.GetReport` -- please change to the correct component and variable names.

packages/core/src/modules/Api/src/module/endpoints/ocpp/2/reporting/GetMonitoringReportEndpoint.ts
```diff
@@ -62,10 +62,25 @@ export class GetMonitoringReportEndpoint extends AbstractMessageEndpoint {
...
+        confirmations.push({
+          success: false,
+          payload: `${ocppConnectionName}: the request size exceeds the limit of ${maxBytes} bytes.`,
+        });
+        continue;
+      }
+
       const itemsPerMessageGetReport =
         (await this._deviceModelService.getItemsPerMessageByComponentAndVariableInstanceAndStationId(
-          COMPONENT_DEVICE_DATA_CTRLR,
```
yuvenus, line 67: As per the OCPP 2.0.1 spec (referring to OCPP-2.1_edition2_part2_specification.pdf, section 2.1.17. ItemsPerMessageGetReport):

Image

If the image isn't loading: 

**Component:** DeviceDataCtrlr
**VariableName**: ItemsPerMessage
**VariableInstance**: GetReport
**Description**: Maximum number of ComponentVariable entries that can be sent in one GetReportRequest or GetMonitoringReportRequest message.

So the component is in fact `COMPONENT_DEVICE_DATA_CTRLR` and the call action should be `OCPP_CallAction.GetReport` -- please change back to the correct component and variable names.

yuvenus (changes requested):

packages/core/src/modules/Api/src/module/endpoints/ocpp/2/reporting/GetMonitoringReportEndpoint.ts
sjh37: You are right, and thank you for the citation - I had this backwards. ItemsPerMessageGetReport and BytesPerMessageGetReport are both defined as constraining GetReportRequest **or** GetMonitoringReportRequest, so the two messages share the DeviceDataCtrlr/GetReport variable rather than GetMonitoringReport having its own on MonitoringCtrlr.

Reverted in 8137ec8f: both reads are back to `COMPONENT_DEVICE_DATA_CTRLR` and `OCPP_CallAction.GetReport`, which is what the ItemsPerMessage read already used. The test that asserted the MonitoringCtrlr behaviour now asserts that the byte limit and the item limit come from the same variable.

What is left in the PR is only the missing BytesPerMessage check itself, in both SetVariables and GetMonitoringReport.

packages/core/src/modules/Api/src/module/endpoints/ocpp/2/reporting/GetMonitoringReportEndpoint.ts
sjh37: You are right, and thank you for the citation - I had this backwards. ItemsPerMessageGetReport and BytesPerMessageGetReport are both defined as constraining GetReportRequest **or** GetMonitoringReportRequest, so the two messages share the DeviceDataCtrlr/GetReport variable rather than GetMonitoringReport having its own on MonitoringCtrlr.

Reverted in 8137ec8f: both reads are back to `COMPONENT_DEVICE_DATA_CTRLR` and `OCPP_CallAction.GetReport`, which is what the ItemsPerMessage read already used. The test that asserted the MonitoringCtrlr behaviour now asserts that the byte limit and the item limit come from the same variable.

What is left in the PR is only the missing BytesPerMessage check itself, in both SetVariables and GetMonitoringReport.

yuvenus (approved):

PR #921, opened 2026-08-21
fix(core): populate the relations the Drizzle authorization repository promises

sjh37:
## Two implementations of one interface, answering differently

`SequelizeAuthorizationRepository` and `DrizzleAuthorizationRepository` both implement
`IAuthorizationRepository`, and `CITRINEOS_USE_DRIZZLE=true` swaps one for the other in the
container. The Drizzle mapper is explicit about what it does not do:

```ts
// Relation, not a scalar column.
groupAuthorization: undefined,
...
// Relation, not a scalar column.
tariff: undefined,
```

and nothing fills them in afterwards. The Sequelize repository loads both.

## groupIdToken is never populated under Drizzle

```ts
// SequelizeAuthorizationRepository._constructQuery
// Eager-load the group Authorization so IdTokenInfo.groupIdToken can be surfaced.
include: [{ model: Authorization, as: 'groupAuthorization' }],
```

The comment names the consumer, and it is real - `TransactionService`:

```ts
groupIdToken: dto.groupAuthorization
  ? ({ idToken: dto.groupAuthorization?.idToken ?? '', ... })
  : null,
```

The Drizzle version does a plain `select().from(authorizationTable)`, so `groupAuthorization` is
always undefined and every authorization response carries `groupIdToken: null`. A charging station
never learns which group a token belongs to, which is the whole point of the field.

## Every driver tariff is dropped, silently

```ts
// Sequelize
include: [{ model: Tariff, as: 'tariff', required: true }]

// Drizzle
.where(and(eq(authorizationTable.tenantId, tenantId), isNotNull(authorizationTable.tariffId)))
```

The Drizzle version filters on `tariffId` being set but attaches no tariff, despite the method name.
`GetTariffsRequestOcpp21Handler` then does:

```ts
const tariff = authorization.tariff;
if (tariff && tariff.tariffId) {
```

so under Drizzle the loop body never runs and every `DriverTariff` is missing from the
`GetTariffs` response. No error, no log - the list is just empty.

## The fix

A self-join through `groupAuthorizationId` for the group, and an inner join to `Tariffs` for the
tariff. The tariff join is inner rather than left for the same reason `required: true` is set on the
Sequelize side: an authorization whose `tariffId` no longer resolves must not come back as though it
had a tariff.

## Tests

`packages/core/test/dal/layers/drizzle/` is new - the Drizzle repositories had no coverage at all
beyond one e2e that flips the env var and checks the server boots, which is how two missing joins
went unnoticed.

Three cases against a real database, the schema created with Sequelize since both layers describe
the same one. Two were confirmed failing first:

```
× surfaces the group authorization    → expected undefined to be 'FLEET-PARENT'
× returns the tariff alongside each authorization that has one
                                      → expected undefined to be 'driver-tariff-1'
```

The third - that `groupAuthorization` stays unset for a token in no group - passed before and after,
and is there so the left join cannot quietly become an inner one.

## Verification

```
npx vitest run    # packages/core: 1390 passed / 4 skipped
prettier --check / eslint                          # 0 errors
```

### DIFF
--- a/packages/core/src/dal/layers/drizzle/repository/Authorization.ts
+++ b/packages/core/src/dal/layers/drizzle/repository/Authorization.ts
@@ -17,6 +17,12 @@ import { type Explicit } from '../types.js';
 import { DrizzleRepository } from './Base.js';
 import type { AuthorizationQuerystring, IAuthorizationRepository } from '@/dal/index.js';
 import { and, eq, isNotNull } from 'drizzle-orm';
+import { alias } from 'drizzle-orm/pg-core';
+import { type TariffEntity, tariffTable } from '../schema/Tariff.js';
+import { toTariffDto } from './Tariff.js';
+
+// Self-join target for the group Authorization a token belongs to.
+const groupAuthorizationTable = alias(authorizationTable, 'groupAuthorization');
 
 // ─── Mapper ──────────────────────────────────────────────────────────────────
 // Maps a Drizzle entity (DB row) to the external AuthorizationDto contract.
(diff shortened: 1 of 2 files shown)


yuvenus (approved):
Should resolve https://github.com/citrineos/citrineos/issues/217

PR #923, opened 2026-08-21
fix(ocpi): scope the Locations by-id reads to the party that asked

sjh37:
## The list is scoped; the by-id reads are not

`LocationsService.getLocations` filters on the CPO named in the OCPI headers:

```ts
const where: Locations_Bool_Exp = {
  Tenant: {
    countryCode: { _eq: ocpiHeaders.toCountryCode },
    partyId: { _eq: ocpiHeaders.toPartyId },
  },
};
```

The three by-id reads beside it take no headers at all:

```ts
async getLocationById(locationId: string): Promise<LocationResponse>
async getEvseById(locationId: string, stationId: string, evseId: number): Promise<EvseResponse>
async getConnectorById(locationId: string, stationId: string, evseId: number, connectorId: number)
```

and their queries filter on the id alone:

```graphql
query GetLocationById($id: Int!) {
  Locations(where: { id: { _eq: $id } }) {
```

The routes do not request the headers either - `getLocations` declares
`@FunctionalEndpointParams() ocpiHeaders: OcpiHeaders`, the other three do not.

So on a CSMS hosting more than one CPO, an authenticated partner can read any CPO's location, EVSE
or connector by walking numeric ids, while `GET /locations` correctly returns only their
counterparty's. Location ids are sequential integers, so there is nothing to guess.

## The fix

All three queries carry the same party predicate the list uses, and the routes pass the headers
through. Both call sites already map an empty result to `ClientUnknownLocation`, so an id outside
the requesting party's estate is indistinguishable from one that does not exist - it fails closed
without disclosing that the id is real.

`GET_LOCATION_BY_ID_QUERY` has a second caller: `BaseTransactionMapper`, when a transaction arrives
without an eager-loaded location. That lookup now passes the transaction's own tenant. It was
already correct by construction - the transaction was fetched under that tenant - so this only
states it.

## Tests

Four cases. The three scope tests assert both halves, because they fail differently: the variables
must carry the country and party, **and** the query document must actually use them. Passing
variables a query ignores filters nothing, so the document is asserted directly.

Confirmed by removing only the predicate and leaving the plumbing in place - the three fail with:

```
→ expected '\n  query GetLocationById($id: Int!, …' to contain 'countryCode: { _eq: $countryCode }'
```

The fourth checks the response code for a location the party cannot see is `2003`
ClientUnknownLocation rather than a generic error.

## Verification

```
full suite:  2044 passed / 6 skipped
prettier --check / eslint                # 0 errors
```

### DIFF
--- a/packages/ocpi-base/src/graphql/operations.ts
+++ b/packages/ocpi-base/src/graphql/operations.ts
@@ -254,6 +254,8 @@ export type GetLocationsQueryResult = {
 
 export type GetLocationByIdQueryVariables = Exact<{
   id: Scalars['Int']['input'];
+  countryCode: Scalars['String']['input'];
+  partyId: Scalars['String']['input'];
 }>;
 
 export type GetLocationByIdQueryResult = {
@@ -340,6 +342,8 @@ export type GetEvseByIdQueryVariables = Exact<{
   locationId: Scalars['Int']['input'];
   stationId: Scalars['String']['input'];
   evseId: Scalars['Int']['input'];
+  countryCode: Scalars['String']['input'];
+  partyId: Scalars['String']['input'];
 }>;
 
 export type GetEvseByIdQueryResult = {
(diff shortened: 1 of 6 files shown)


vicky-ahmed (approved):
LGTM

PR #924, opened 2026-08-21
test(ocpp): cover the authorization token round trip end to end

sjh37:
## What

Adds one testcontainers-backed suite covering the full life of an `Authorization` row: enrolled once, presented by a charger over each OCPP version, and joined back to the transaction it paid for.

Test only. No production code changes.

## Why

An `Authorization` is resolved three different ways, and no test exercised more than one of them against the same row:

| Leg | Resolves on | Source |
| --- | --- | --- |
| OCPP 2.0.1 `Authorize` | `(idToken, idTokenType)` | `authorize-request-ocpp-201-handler.ts:154` |
| OCPP 1.6 `Authorize` | `idToken` alone, refusing when more than one row matches | `authorize-request-ocpp-16-handler.ts:65`, `:81` |
| Transaction attribution | `(idToken, idTokenType)` | `TransactionEvent.ts:172`, `:252` |

`SequelizeAuthorizationRepository._constructQuery` applies the `idTokenType` predicate only when the caller supplies a type, which is what lets one row serve both protocols - and also what makes a second row for the same token invisible to 2.0.1 but fatal to 1.6.

Those rules interact in ways that are not evident from reading any one of them, so the suite pins the combined behaviour rather than each part separately.

## What it covers

- A single row accepts over both 2.0.1 and 1.6.
- 2.0.1 given the right token under the wrong `idTokenType` returns `Unknown` - the request is refused, not merely left unattributed.
- `idToken` is `CITEXT`, so case does not matter.
- Separators do: `AA:BB:...`, `AA-BB-...` and `AABB...` are three distinct tokens. Case-insensitivity might reasonably be read as implying more leniency than there is, and all three spellings pass `validateIdentifierStringIdToken`, so nothing rejects them earlier.
- Enrolling one token a second time under a second type is permitted by the composite unique, satisfies 2.0.1, and makes 1.6 refuse the token outright.
- `Transaction.authorizationId` points at the enrolled row after a `TransactionEvent`.

## Verification

```
$ npx vitest run packages/ocpp/test/handlers/requests/authorization-token-round-trip-integration.test.ts

 ✓ packages/ocpp/test/handlers/requests/authorization-token-round-trip-integration.test.ts (7 tests) 10642ms
   ✓ authorises over 2.0.1 when the charger presents the MAC it was enrolled under 593ms
   ✓ authorises over 1.6, where the same row is found by the idTag alone 607ms
   ✓ refuses a 2.0.1 request that names the right MAC under the wrong type 615ms
   ✓ matches the MAC irrespective of case, because idToken is CITEXT 618ms
   ✓ does not match the same MAC written with different separators 603ms
   ✓ breaks 1.6 if the same MAC is enrolled a second time under another type 625ms
   ✓ joins the energy back to the enrolled vehicle through Transaction.authorizationId 705ms

 Test Files  1 passed (1)
      Tests  7 passed (7)
```

Requires a container runtime, as the other `*-integration.test.ts` suites do.

### DIFF
--- a/packages/ocpp/test/handlers/requests/authorization-token-round-trip-integration.test.ts
+++ b/packages/ocpp/test/handlers/requests/authorization-token-round-trip-integration.test.ts
@@ -1,0 +1,290 @@
+// SPDX-FileCopyrightText: 2026 Contributors to the CitrineOS Project
+//
+// SPDX-License-Identifier: Apache-2.0
+
+import { afterAll, beforeAll, beforeEach, describe, expect, it } from 'vitest';
+import { GenericContainer, type StartedTestContainer, Wait } from 'testcontainers';
+import type { Sequelize } from 'sequelize-typescript';
+import { type BootstrapConfig, DEFAULT_TENANT_ID, type IMessage } from '@citrineos/base';
+import {
+  AuthorizationStatusEnum,
+  EventGroup,
+  IdTokenEnum,
+  type IdTokenEnumType,
+  MessageOrigin,
+  MessageState,
+  OCPP1_6,
+  OCPP2_0_1,
+  OCPP_CallAction,
+  type OcppRequest,
+  OCPPVersion,
+} from '@citrineos/types';
+import {
+  Authorization,
+  ChargingStation,
+  DefaultSequelizeInstance,
+  SequelizeAuthorizationRepository,
+  SequelizeTransactionEventRepository,
+  Tenant,
+  Transaction,
+} from '@citrineos/dal';
+import { AuthorizeRequestOcpp16Handler, AuthorizeRequestOcpp201Handler } from '@handlers/index.js';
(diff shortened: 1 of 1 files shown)


sjh37:
Rebased onto `486f1e99` after #990. This one needed a home rather than a rename: the test was in
`packages/core/test/integration/`, and `core` is gone. It now sits at
`packages/ocpp/test/handlers/requests/authorization-token-round-trip-integration.test.ts` - it
drives `AuthorizeRequestOcpp16Handler` and `AuthorizeRequestOcpp201Handler`, so it belongs with the
request handlers rather than in `dal`. Imports repointed to `@citrineos/dal` and
`@test/test-container.js`.

Verified on `486f1e99`: build clean, full suite green (142 test files, 2308 passing / 4 skipped),
lint clean.

thanaParis (approved):

PR #925, opened 2026-08-21
fix(ocpp-server): make the Authorizations token uniqueness actually hold

sjh37:
## The defect

Uniqueness on `Authorizations` is a unique **index** over `(tenantId, idToken, idTokenType)`, created by `20260413000000-fix-authorization-unique-index-add-tenant.ts:53`. `idTokenType` is nullable (`Authorization.ts:51-55`).

Postgres treats NULLs in a unique index as distinct from one another, so the index constrains nothing at all for rows enrolled without a token type. The same token can be inserted any number of times.

That matters because OCPP 1.6 has no token type to match on. `authorize-request-ocpp-16-handler.ts:65` looks the `idTag` up by string alone and `:81` refuses when more than one row comes back:

```ts
if (authorizations.length > 1) {
  this._logger.error(`Too many authorizations found for idToken: ${request.idTag}`);
  response.idTagInfo.status = OCPP1_6.AuthorizeResponseStatus.Invalid;
```

So one accidental duplicate deauthorises that token permanently, on every station, with nothing surfaced beyond a log line. `SequelizeAuthorizationRepository._constructQuery:60` applies the type predicate only when a caller supplies a type, so 2.0.1 keeps working and only 1.6 breaks.

## The second half

`addIndex(..., { unique: true })` creates an index, not a constraint. Hasura resolves `on_conflict` against unique **constraints**, so an upsert naming this target fails at runtime. The operator-ui e2e fixture already documents working around exactly this (`apps/operator-ui/tests/e2e/fixtures/everest.ts:374`):

> Idempotency: the Authorizations table has a unique INDEX on (idToken, idTokenType) but no matching unique CONSTRAINT, so Hasura cannot resolve `on_conflict: { constraint: idToken_type }` at runtime even though its enum lists the name. Query-then-insert keeps the helper idempotent without depending on a constraint Hasura cannot apply.

This is load-bearing because **no CitrineOS code path creates an `Authorization`** - there is no `Authorization.create` anywhere in `packages/ocpp`, `packages/dal` or `packages/ocpi-base`, and no REST endpoint for it. Hasura is the only way a token is enrolled, so query-then-insert is currently the only idempotent option, and it races.

Both halves have one cause: uniqueness expressed as an index over a nullable column.

## The fix

One migration replacing the index with a `UNIQUE NULLS NOT DISTINCT` constraint of the same name. It:

- refuses to run over pre-existing duplicates and lists them, following the pattern in `20260413000000`;
- drops uniqueness under either the migrated name or the one `sync()` derives, so it applies to a database built either way.

Note `sync()` and the migrations had already diverged: a `sync()`-built database gets a constraint named `Authorizations_idToken_idTokenType_tenantId_key`, a migrated one an index named `idToken_type`. Both are handled.

`NULLS NOT DISTINCT` requires Postgres 15; the shipped compose is `postgis/postgis:16-3.5`.

Sequelize has no `NULLS NOT DISTINCT` option, so a database built by `sync()` alone keeps the looser constraint until it is migrated. Happy to add an `@AfterSync` hook if you would rather the two paths converge now.

## Verification

First run is with `up()` doing what `20260413000000` did, so the failures are the defect itself:

```
 ✓ is not enforced by the pre-migration index when idTokenType is null 447ms
 × rejects the duplicate once migrated 459ms
 ✓ still allows one token under two different types 442ms
 ✓ still allows the same untyped token in two tenants 440ms
 × leaves uniqueness as a named constraint, which is what on_conflict resolves against 351ms
   → expected [] to include 'idToken_type'
 × refuses to migrate over existing duplicates, naming them 433ms
 ✓ restores the unique index on down 351ms

AssertionError: promise resolved "Authorization{ …(6) }" instead of rejecting
AssertionError: expected [] to include 'idToken_type'
AssertionError: promise resolved "undefined" instead of rejecting

 Tests  3 failed | 4 passed (7)
```

`expected [] to include 'idToken_type'` is the Hasura half: there is no unique constraint on the table at all.

With the migration as committed:

```
 ✓ apps/ocpp-server/test/migrations/authorization-unique-constraint-integration.test.ts (7 tests) 9204ms

 Test Files  1 passed (1)
      Tests  7 passed (7)
```

The suite starts from the shape `20260413000000` leaves a database in, so the defect is demonstrated on the schema the migration then repairs.

### DIFF
--- a/apps/ocpp-server/migrations/20260821120000-authorization-unique-constraint-nulls-not-distinct.ts
+++ b/apps/ocpp-server/migrations/20260821120000-authorization-unique-constraint-nulls-not-distinct.ts
@@ -1,0 +1,100 @@
+// SPDX-FileCopyrightText: 2026 Contributors to the CitrineOS Project
+//
+// SPDX-License-Identifier: Apache-2.0
+'use strict';
+
+/** @type {import('sequelize-cli').Migration} */
+import { QueryInterface, QueryTypes } from 'sequelize';
+
+const TABLE_NAME = 'Authorizations';
+const NAME = 'idToken_type';
+const COLUMNS = '"tenantId", "idToken", "idTokenType"';
+
+// sync() names the constraint it derives from the model's `unique: 'idToken_type'` group after the
+// table and columns, so a database built that way carries a different name from a migrated one.
+const SYNC_DERIVED_NAME = 'Authorizations_idToken_idTokenType_tenantId_key';
+
+const constraintExists = async (queryInterface: QueryInterface, name: string): Promise<boolean> => {
+  const rows = await queryInterface.sequelize.query(
+    `SELECT 1 FROM information_schema.table_constraints
+       WHERE table_schema = 'public' AND table_name = :table AND constraint_name = :name`,
+    { replacements: { table: TABLE_NAME, name }, type: QueryTypes.SELECT },
+  );
+  return rows.length > 0;
+};
+
+const dropUniqueness = async (queryInterface: QueryInterface) => {
+  for (const name of [NAME, SYNC_DERIVED_NAME]) {
+    if (await constraintExists(queryInterface, name)) {
+      await queryInterface.sequelize.query(
+        `ALTER TABLE "${TABLE_NAME}" DROP CONSTRAINT "${name}"`,
+        { type: QueryTypes.RAW },
+      );
+    }
+  }
+  // A plain unique index shares the namespace but is not a constraint, so it needs its own drop.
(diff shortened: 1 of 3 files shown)


sjh37:
Rebased onto `486f1e99` after #990. The migration is untouched. Its test is renamed to
`apps/ocpp-server/test/migrations/authorization-unique-constraint-integration.test.ts` and its
import moved from `@dal/index.js` to `@citrineos/dal`.

Verified on `486f1e99`: build clean, full suite green (142 test files, 2308 passing / 4 skipped),
lint clean.

thanaParis (approved):

PR #928, opened 2026-08-21
fix(core): record every installed certificate a station reports

sjh37:
## The defect

`GetInstalledCertificateIds` is the reconciliation call - the station answers with `certificateHashDataChain`, a list of every certificate it holds. The handler walks that list, but looks each entry up by station and type alone (`GetInstalledCertificateIdsResponseOcpp2Handler.ts:104-110`):

```ts
let existingInstalledCertificate =
  await this._installedCertificateRepository.readOnlyOneByQuery(tenantId, {
    where: {
      ocppConnectionName: ocppConnectionName,
      certificateType: certificateType,
    },
  });
```

That cannot tell two certificates of one type apart, and a station holding several trusted V2G roots is the ordinary case rather than an edge one.

**Reported together**, the row created for the first root is found by the lookup for the second and overwritten in place, so only the last of each type survives. **Already recorded**, two rows of a type make `readOnlyOneByQuery` throw (`packages/base/src/interfaces/repository.ts:41`) and the reconciliation aborts partway through the list, leaving the record in whatever state it had reached.

Either way the CSMS ends up believing a station trusts fewer roots than it does - and this is the call whose entire purpose is to find that out.

## The fix

Match on the hash data that identifies the certificate, so each reported certificate gets its own record.

The update branch goes with it: every field it assigned - `hashAlgorithm`, `issuerNameHash`, `issuerKeyHash`, `serialNumber` - is now part of the match, so it could never change anything.

Certificates the station no longer reports are still left behind. Only the `NotFound` branch clears records, and that is unchanged here; narrowing this lookup does not make the staleness any worse.

## Verification

Before:

```
 × records every certificate the station reports, not only the last of each type 676ms
   → expected [ 'serial-2' ] to deeply equal [ 'serial-1', 'serial-2' ]
 × reconciles a station that already has two certificates of one type recorded 647ms
   → More than one value found for query: {"where":{"ocppConnectionName":"CP-CERT-2","certificateType":"V2GRootCertificate"}}
 ✓ does not duplicate a certificate that is reported again 632ms

 Tests  2 failed | 1 passed (3)
```

The first failure is the overwrite - one root reported, two sent. The third test is the control: it passes before and after, so re-reporting a single certificate still does not duplicate it.

After:

```
 ✓ packages/core/test/handlers/responses/2/GetInstalledCertificateIdsResponseOcpp2Handler.integration.test.ts (3 tests) 8515ms
 Test Files  1 passed (1)
      Tests  3 passed (3)
```

Full suite:

```
```

One earlier run of the full suite showed a transient failure in an unrelated file, with the run taking 126s against the usual 78s and two extra suites skipped - a testcontainers startup timeout under load rather than a result. The re-run above is clean.

### DIFF
--- a/packages/core/src/handlers/responses/2/GetInstalledCertificateIdsResponseOcpp2Handler.ts
+++ b/packages/core/src/handlers/responses/2/GetInstalledCertificateIdsResponseOcpp2Handler.ts
@@ -104,33 +104,35 @@ export class GetInstalledCertificateIdsResponseOcpp2Handler extends AbstractHand
         const certificateHashData = certificateHashDataWrap.certificateHashData;
         const certificateType =
           certificateHashDataWrap.certificateType as unknown as CertificateUseEnumType;
-        let existingInstalledCertificate =
+        // A station holds more than one certificate of a type - several trusted V2G roots is the
+        // ordinary case - so each reported certificate is matched on the hash data that identifies
+        // it. Matching on the type alone finds a sibling, and then one root overwrites the other.
+        const existingInstalledCertificate =
           await this._installedCertificateRepository.readOnlyOneByQuery(tenantId, {
             where: {
               ocppConnectionName: ocppConnectionName,
               certificateType: certificateType,
+              hashAlgorithm: certificateHashData.hashAlgorithm,
+              issuerNameHash: certificateHashData.issuerNameHash,
+              issuerKeyHash: certificateHashData.issuerKeyHash,
+              serialNumber: certificateHashData.serialNumber,
             },
           });
         if (existingInstalledCertificate) {
-          existingInstalledCertificate.hashAlgorithm = certificateHashData.hashAlgorithm;
-          existingInstalledCertificate.issuerNameHash = certificateHashData.issuerNameHash;
-          existingInstalledCertificate.issuerKeyHash = certificateHashData.issuerKeyHash;
-          existingInstalledCertificate.serialNumber = certificateHashData.serialNumber;
-          await existingInstalledCertificate.save();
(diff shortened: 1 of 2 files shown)


vicky-ahmed (approved):

thanaParis (changes requested):
This PR solves one problem and replaces it with another: the current behavior results in too few installed certificates, the new behavior results in too many. It is a bug that the logic does not allow more than one certificate per type, however the purpose of this handler is to make the db reflective of the current state of the charger, i.e. the certs in the db are the only the ones the CSMS thinks are _currently_ on the charging station. this new logic would leave certs no longer on the station in the db forever.

sjh37:
@thanaParis Agreed - that is a fair description of what I did, and the handler's job is to make the
database reflect what is on the station now. Reworked.

**What it does now**

The response carries the station's current set, so the handler treats it as one:

- a reported certificate that is not recorded is inserted (the original fix - the lookup matched on
  `certificateType` alone, so a station with two certificates of one type kept only the last);
- a recorded certificate for that station that the station did **not** report is removed.

Removal is scoped to the `certificateType` the originating request asked for, so a
`GetInstalledCertificateIds` for one type does not clear the others. That is the same request lookup
the `NotFound` branch beside it already does, so I hoisted it into a single
`readRequestedCertificateType` and both branches use it.

**Tests** - three added to the three that were there:

- `removes a certificate the station no longer reports`
- `removes every certificate when the station reports none of that type`
- `leaves a certificate of a type the request did not ask about`

Six green. Neither too few nor too many, and the scoping case is now covered.

thanaParis (approved):

PR #929, opened 2026-08-21
fix(ocpp): check the OCSP responder URL before the CSMS reaches it

sjh37:
@yuvenus flagging you directly on this one - it is a security fix rather than an ordinary correctness one, and I could not find a private channel to raise it through first. Issues are disabled on this repo, `GET /repos/citrineos/citrineos-core/private-vulnerability-reporting` returns `{"enabled":false}`, and there is no `SECURITY.md` in either `citrineos-core` or `citrineos`. If you would rather these went somewhere non-public in future, say where and I will use it. Enabling GitHub's private vulnerability reporting on the repo would give you that for free.

## The defect

`sendOCSPRequest` passed its `responderURL` straight to `fetch` (`certificate-util.ts:313`):

```ts
export async function sendOCSPRequest(ocspRequest, responderURL): Promise<string> {
  const response = await fetch(responderURL, { method: 'POST', ... });
```

Every URL that reaches it is chosen by the charging station:

| Call site | Where the URL comes from | Response body returned to the station? |
| --- | --- | --- |
| `get-certificate-status-request-ocpp-2-handler.ts:50` | `ocspRequestData.responderURL` | **Yes** |
| `CertificateAuthority.ts:241` (Authorize) | `iso15118CertificateHashData[].responderURL` | No |
| `CertificateAuthority.ts:205` (chain) | AIA extension of a station-supplied certificate | No |

So a station names a URL, the CSMS issues a POST to it from inside its own network, and on the first path hands the response body back over the OCPP connection. A CSMS is reachable by anything that can complete a websocket handshake, and `GetCertificateStatus` needs no certificate - just a well-formed message - so link-local and private-range addresses that are otherwise unreachable become readable from the far end of a station connection.

## The fix

All three paths meet in `sendOCSPRequest`, so the check goes there.

A responder must be `http` or `https`, and must not resolve to a loopback, private, link-local, carrier-grade NAT or unique-local address. **Every** address a name resolves to is tested, so a name answering with one public and one private address is refused rather than half-accepted.

Checking a name and then handing the URL to a client that resolves it again would leave the check meaningless - the second answer decides where the request goes, and whoever runs the name's DNS chooses both answers. The request is therefore sent to the address that was checked. That is why this moves off `fetch` and onto `node:http(s)`, which accepts the address directly; `fetch` would need an `undici` dispatcher with a pinned `lookup`, and `undici` is not a dependency here. The original host travels in the `Host` header and in `servername`, so the responder still sees the name it expects and TLS is still verified against the name rather than the address. Redirects are refused rather than followed, since a public responder that redirects would otherwise reach exactly what the check excluded, and the request now has a timeout.

## Notes for review

- `resolvePublicOcspResponder` and `isPrivateAddress` are exported so the ranges are testable in isolation.
- An operator running a private OCSP responder would now be refused. I have assumed that is not a case worth supporting, since a responder for publicly-trusted V2G certificates is itself public - happy to put it behind config if you would rather it were allowed.
- The existing `sendOCSPRequest` tests mocked `global.fetch`; they now mock the sender, and the transport is covered in the new file.

## Verification

```
 ✓ packages/ocpp/test/services/certificate/ocsp-responder-url.test.ts (32 tests) 9ms
 ✓ packages/ocpp/test/services/certificate/certificate-util.test.ts (9 tests) 63ms
```

The test that matters most asserts the request goes where the check said, not where the name points:

```
 ✓ sends the request to the address that was checked, not to the name
     host: '203.0.113.10', servername: 'ocsp.example.com', headers.Host: 'ocsp.example.com'
 ✓ refuses a name that answers with one public and one private address
 ✓ does not issue a request at all when the responder is private
```

Full suite:

```
      Tests  2075 passed | 4 skipped (2079)
```

### DIFF
--- a/packages/ocpp/src/services/certificate/certificate-util.ts
+++ b/packages/ocpp/src/services/certificate/certificate-util.ts
@@ -14,6 +14,7 @@ import moment from 'moment';
 import type { ILogObj } from 'tslog';
 import { Logger } from 'tslog';
 import KJUR = jsrsasign.KJUR;
+import { sendToPublicOcspResponder } from './ocsp-responder-url.js';
 import OCSPRequest = jsrsasign.KJUR.asn1.ocsp.OCSPRequest;
 import X509 = jsrsasign.X509;
 import KEYUTIL = jsrsasign.KEYUTIL;
@@ -304,26 +305,25 @@ export function createOcspRequest(
   });
 }
 
+const OCSP_REQUEST_TIMEOUT_MS = 10_000;
+
 export async function sendOCSPRequest(
   ocspRequest: OCSPRequest,
   responderURL: string,
 ): Promise<string> {
-  const response = await fetch(responderURL, {
-    method: 'POST',
-    headers: {
-      'Content-Type': 'application/ocsp-request',
-      Accept: 'application/ocsp-response',
-    },
-    body: Uint8Array.from(Buffer.from(ocspRequest.getEncodedHex(), 'hex')),
-  });
+  const response = await sendToPublicOcspResponder(
+    responderURL,
+    Buffer.from(ocspRequest.getEncodedHex(), 'hex'),
+    OCSP_REQUEST_TIMEOUT_MS,
+  );
 
-  if (!response.ok) {
+  if (response.status < 200 || response.status > 299) {
     throw new Error(
-      `Failed to fetch OCSP response from ${responderURL}: ${response.status} with error: ${await response.text()}`,
+      `Failed to fetch OCSP response from ${responderURL}: ${response.status} with error: ${response.body}`,
     );
   }
 
-  return Buffer.from(await response.arrayBuffer()).toString('hex');
+  return response.body.toString('hex');
 }
 
 export function parseCSRForVerification(csrPem: string): CertificationRequest {
--- a/packages/ocpp/src/services/certificate/ocsp-responder-url.ts
+++ b/packages/ocpp/src/services/certificate/ocsp-responder-url.ts
@@ -7,9 +7,6 @@ import { lookup } from 'node:dns/promises';
 import { request as httpRequest } from 'node:http';
 import { type RequestOptions, request as httpsRequest } from 'node:https';
 
-// The request is sent to the address that was checked, with the name in Host and SNI, so a second
-// DNS answer cannot send it somewhere the check would have refused.
-
 const ALLOWED_PROTOCOLS = ['http:', 'https:'];
 
 function isPrivateIPv4(address: string): boolean {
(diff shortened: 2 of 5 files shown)


sjh37:

`CertificateUtil` moved under it. Rebased; the certificate suites are green (83 passed, 4 skipped). Note the RFC 5280 comment near `getValidityTimeString` is now upstream's, so this PR no longer touches it.

sjh37:

- the guard is now `packages/ocpp/src/services/certificate/ocsp-responder-url.ts`, imported by
  `certificate-util.ts` beside it
- its test is `packages/ocpp/test/services/certificate/ocsp-responder-url.test.ts`, and the
  `CertificateUtil` suite's `vi.mock` target moved with it, to
  `@/services/certificate/ocsp-responder-url.js`

Verified on `486f1e99`: build clean, full suite green (142 test files, 2334 passing / 4 skipped),
lint clean.

sjh37:
Rebased onto `c1bd4ca` - the conflict was the `@services/index.js` import rename in the CertificateUtil suite. Comments cut back to the four-line note on the guard and the CIDR labels, and the follow-up "drop the in-code commentary" commit is squashed away, so this is one commit again.

Verified on `c1bd4ca`: build clean, full suite green (327 files, 4409 passing / 8 skipped), lint 0 errors.

sjh37:
Removed the last in-code comments: the address-pinning note at the top of `ocsp-responder-url.ts` and the labels on the five range checks. The pinning rationale is in the description above ("The fix", second paragraph). Code unchanged otherwise.

Verified on this head: build clean, full suite green (327 files, 4409 passing / 8 skipped), lint 0 errors.

sjh37:
Heads-up for whoever merges: this PR and #958 both edit `packages/ocpp/src/services/certificate/certificate-util.ts` (the URL guard here, `sendOCSPRequest` there), and git reports a conflict between the two branches. Merge in either order - whichever lands second, I will rebase the same day so it is one click again.

sjh37:
Rebased onto `0fd4e1f` now that #958 has landed.

#958 made `sendOCSPRequest` post the DER of an `OCSPRequest` and read the response as bytes, so the responder guard now carries bytes both ways: `sendToPublicOcspResponder` takes the request as a `Uint8Array` and returns the body as a `Buffer`, and `sendOCSPRequest` hex-encodes it exactly as #958 does. The `Content-Type` and `Accept` headers #958 added were already set by the guard.

#958's GetCertificateStatus handler test stubbed the global `fetch`, which the guard does not use, so it now stubs `sendToPublicOcspResponder` instead - same four assertions, including that the body is the DER of an `OCSPRequest` and that `ocspResult` is the responder's DER in base64.

Verified on `0fd4e1f`: build clean, full suite green (333 files, 4463 passing / 8 skipped), lint 0 errors.

thanaParis (changes requested):
I don't think these checks earn their complexity here — resolve-then-connect is rebinding-prone, and a redirect to an internal address bypasses all of them.

I would merge an optional config list of permitted responder hosts, empty by default meaning no restriction, plus disabling redirect-following on this client and adding a timeout and a response size cap. 

Note: this is also solvable at the infrastructure layer by restricting egress, so the config is a convenience.

sjh37:
Hi @thanaParis 

Reworked to your design, thanks. The argument against resolve-then-connect is right, especially the redirect hole.

Gone: the DNS resolution, the private-range checks and the address pinning. In their place `integrations.ocsp.allowedResponderHosts`, empty by default so there's no restriction unless an operator opts in, and the request itself is now bounded - back on plain `fetch` with `redirect: 'error'`, a 10s `AbortSignal.timeout`, and the response body capped at 64 KiB. The cap checks `content-length` and also counts bytes while streaming, so a responder that omits the header can't stream unbounded.

I agree that egress restriction is the stronger control, and this is a convenience on top of it.

thanaParis (approved):

PR #930, opened 2026-08-21
(Sequelize) Schema Drift Checker

yuvenus:
# Why? 

Currently, not all Sequelize models declared in the code match with the way the database is setup via the database migrations. This happened because we stopped using the Sequelize models to determine the database schemas, preferring to use migrations instead. As a result, you'd only find out something was amiss between the models and database the moment you tried to insert data into the database.

Note that any column in the Sequelize models annotated with Column decorator will count towards this check (since that's the only way Sequelize knows about those columns).

This PR aims to solve the drift by not allowing CitrineOS to start if any "schema drift" is detected, and also actually fixes the current existing drift.

## Tariff Alt Text?

Somewhat relatedly, the UI for Tariff Alt Text input was fixed as well (since I was testing to make sure the schema drift fixes actually worked!), hence the Operator UI changes in this PR.

# What counts as drift?

For this PR, the following things count as "schema drift":

1. If an entire table is missing
2. If a column is missing
3. If the column type is different
4. If the nullability of a column is different
5. If the "width" of the column is different (i.e. for varchars)

There are a few allowances though for some of these differences, depending on how severe they are:

1. If an entire table is missing...
  * No allowances, it should fail the diff check ❌ 
2. If a column is missing...
  * If the column is in the database, but not the model, that's a warning ⚠️ 
  * If the column is in the model, but not the database, that's a failure ❌ 
3. If the column type is different...
  * No allowances, it should fail the diff check ❌ 
4. If the nullability of a column is different...
  * If the column is null in the database, but not the model, that's a failure ❌  (bc this would cause a type error)
  * If the column is null in the model, but not the database, that's a warning ⚠️ 
5. If the "width" of the column is different...
  * If the column is wider in the database than the model, that's a warning ⚠️ 
  * If the column is wider in the model than the database, that's a failure ❌  (since inserts from the code could fail if you exceed the width)

## What were the actual drifts? 

[sequelize-drift-checker-logs.txt](https://github.com/user-attachments/files/31699506/sequelize-drift-checker-logs.txt)

Attached is the text dump of the logs at the time prior to the schema being fixed.
 
# Configuration Updates

Three new database configurations were added to the config to provide flexibility for the validation:

1. `schema` -> Defaults to 'public'. What the actual database schema should be. Not an immediate thing we worry about now, but for folks who have a non-public schema, this should help.
2. `validateSchema` -> Defaults to true. If true, will run the schema validation. If false, will not run it.
3. `validateSchemaSeverity` -> Either 'error' or 'warn' -- defaults to 'error'. If error, schema validation will block startup if there's a failure. If warn, schema validation will not block startup if there's a failure.

# Reconciling The Drift

For any errors on the database side, a new migration was added <date>-reconcile-erroring-schema-differences.ts to do this reconciliation. Otherwise, the affected Sequelize models were updated in this PR (luckily, all of them were changes to the nullability).

Warnings were left alone for the most part to keep this PR focused on making sure CitrineOS can continue to run.

# What about Drizzle?

Since we're still in the middle of the Drizzle migration and will likely be doing it for a while, the Drizzle-version of the schema drift checker will come at a later time. This PR keeps things focused to Sequelize so we don't go crazy. 😄


thanaParis (changes requested):
please fix conflicts

yuvenus:
@thanaParis Conflicts resolved and tests are all good again 😎👍

thanaParis (approved):

PR #931, opened 2026-08-21
fix(dal): store meter values against the connector the transaction is on

sjh37:
## The defect

`MeterValue.connectorId` is a foreign key to `Connector.id` (`meter-value.ts:68-73`), and the OCPP 1.6 handler writes the connector number out of the `MeterValues` message into it (`meter-values-request-ocpp-16-handler.ts:63`):

```ts
meterValueEntity.connectorId = connectorId;   // connectorId = message.payload.connectorId
```

Those are different numbers that share a type. The message carries the station's own connector number - 1, 2, 3, restarting at 1 on every station - while `Connector.id` is a database-wide surrogate key.

Two outcomes, both silent:

- **A connector holds that id.** The reading is recorded against it. Since ids are handed out across all stations, that is usually a connector on a different station.
- **No connector holds that id.** The insert violates `MeterValues_connectorId_fkey`, and `meter-values-request-ocpp-16-handler.ts:76` catches it and logs `Failed to process MeterValues`. The reading is lost.

The second case is the common one on any station commissioned after the first few, since its connector numbers are small while the ids in use are not.

## The fix

`updateTransactionByMeterValues` already loads the transaction the readings belong to, and `Transaction.connectorId` is a correct reference to the connector. The reference is taken from there, and the handler no longer sets it at all.

## Verification

Before. The fixture commissions a neighbouring station first, so database id 2 belongs to it while this station calls its own connector 2:

```
 × stores the meter value against the connector the transaction is on
   → expected 2 to be 4
 × does not attribute the meter value to another station
   → expected 2 not to be 2
 × still stores the meter value when no connector carries that number as its id
   → expected +0 to be 1

 Tests  3 failed (3)
```

`expected 2 to be 4` is the misattribution - the reading landed on the other station's connector. `expected +0 to be 1` is the loss.

Worth noting the first two pass on a database where ids happen to line up with connector numbers, which is what a freshly seeded test database looks like. The fixture deliberately breaks that alignment.

After:

```
 ✓ packages/ocpp/test/handlers/requests/1.6/meter-values-connector-reference-integration.test.ts (3 tests) 14951ms
 Test Files  1 passed (1)
      Tests  3 passed (3)
```

Full suite:

```
```

## Related

`StatusNotifications.connectorId` and `EvseTypes.connectorId` are also foreign keys to `Connector.id` that get written with OCPP-level connector numbers. I will raise those separately - the fix there is to the associations rather than to a caller.

---

**Correction, 2 September.** The second outcome above assumes `MeterValues_connectorId_fkey`. That foreign key is declared in the Sequelize model only; a migration-built database has no foreign key on `MeterValues.connectorId` (checked in `pg_constraint` after applying every migration on `next`), so on a migrated database the reading is not lost - it is stored, with the station's connector number in a column the model reads as a `Connectors.id`. The first outcome, the misattribution, is the one that applies in production, and it is what the fix addresses. Nothing here depends on how the station was commissioned.

### DIFF
--- a/packages/dal/src/repositories/sequelize/transaction-event.ts
+++ b/packages/dal/src/repositories/sequelize/transaction-event.ts
@@ -628,6 +628,10 @@ export class SequelizeTransactionEventRepository
         meterValue.transactionDatabaseId = transaction.id;
         meterValue.transactionId = transaction.transactionId;
         meterValue.tariffId = transaction.tariffId;
+        // MeterValue.connectorId references Connector.id, while a MeterValues message names the
+        // station's own connector number. They are different numbers of the same type, so the
+        // reference comes from the transaction rather than from the message.
+        meterValue.connectorId = transaction.connectorId ?? undefined;
         const createdMeterValue = await MeterValue.create(meterValue);
         this.meterValue.emit('created', [createdMeterValue]);
       }),
--- a/packages/dal/src/repositories/sequelize/transaction-event.ts
+++ b/packages/dal/src/repositories/sequelize/transaction-event.ts
@@ -628,9 +628,6 @@ export class SequelizeTransactionEventRepository
         meterValue.transactionDatabaseId = transaction.id;
         meterValue.transactionId = transaction.transactionId;
         meterValue.tariffId = transaction.tariffId;
-        // MeterValue.connectorId references Connector.id, while a MeterValues message names the
-        // station's own connector number. They are different numbers of the same type, so the
-        // reference comes from the transaction rather than from the message.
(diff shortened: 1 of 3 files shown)


sjh37:
Rebased onto `486f1e99` after #990. The two one-line changes are now in
`packages/dal/src/repositories/sequelize/transaction-event.ts` and
`packages/ocpp/src/handlers/requests/1.6/meter-values-request-ocpp-16-handler.ts`, both tracked as
renames; the integration test is
`packages/ocpp/test/handlers/requests/1.6/meter-values-connector-reference-integration.test.ts`.

lint clean.

thanaParis (approved):

PR #932, opened 2026-08-21
fix(core): stop treating a reported connector number as a Connector reference

sjh37:
## The defect

`StatusNotifications.connectorId` and `EvseTypes.connectorId` both hold a connector number a charging station reported - the station's own numbering, restarting at 1 on every station, and within every EVSE for the 2.0.1 one. Neither is a reference to a `Connector` row.

Three declarations made them foreign keys to `Connector.id` anyway:

- `StatusNotification.ts:72` - `@BelongsTo(() => Connector, 'connectorId')`
- `Connector.ts:158` - `@HasMany(() => StatusNotification, 'connectorId')`
- `Connector.ts:143` - `@HasMany(() => EvseType, 'connectorId')`

Confirmed against a live schema:

```
StatusNotifications_connectorId_fkey   connectorId  -> Connectors.id
EvseTypes_connectorId_fkey             connectorId  -> Connectors.id
```

Two outcomes:

- **A Connector holds that id.** The row resolves to it. Ids are handed out across every station, so connector 1 on one station means whichever connector was created first anywhere.
- **None does.** The insert fails. An OCPP 2.0.1 `StatusNotification` for a station whose connectors are not yet recorded cannot be stored at all - `insert or update on table "StatusNotifications" violates foreign key constraint "StatusNotifications_connectorId_fkey"`.

The 1.6 path avoids the second only by upserting the Connector first, and says so at `StatusNotificationService.ts:205`:

> We upsert the Connector BEFORE saving the StatusNotification because StatusNotifications.connectorId has an FK to Connectors.connectorId.

The constraint is actually to `Connectors.id`, and the 2.0.1 path saves the StatusNotification first.

## The fix

Neither navigation is read anywhere - `connector.statusNotifications`, `connector.evseTypes` and `statusNotification.connector` have no callers - so the three declarations go, and a migration drops the constraints they left on already-migrated databases.

`Connectors.evseTypeConnectorId -> EvseTypes.databaseId` is a genuine reference and is untouched.

## Verification

Before, on a database where the ids do not happen to line up with the connector numbers:

```
 ✓ records the connector number the station sent
 × leaves the number alone rather than forcing it to name a connector row
 × does not constrain the EvseType connector number to a connector row either
```

After:

```
 ✓ packages/core/test/modules/Transactions/module/StatusNotificationConnectorNumber.integration.test.ts (3 tests)
 Test Files  1 passed (1)
      Tests  3 passed (3)
```

Full suite:

```
```

0 lint errors, the 2 pre-existing warnings.

## Still open, not fixed here

With the constraint gone, an OCPP 2.0.1 `StatusNotification` from a station whose connectors are not yet recorded fails differently: `Connector.evseTypeConnectorId cannot be null`. `StatusNotificationService.ts:96` takes it from a matching connector, and there is none to match yet.

The 1.6 path handles exactly this with `commissionEvseForOcpp16Connector`; 2.0.1 has no equivalent, so its connectors only ever come into existence through the transaction path. That looked like a design decision rather than an oversight, so I have left it alone - happy to open something if it is not.

### DIFF
--- a/apps/ocpp-server/migrations/20260821140000-drop-connector-number-foreign-keys.ts
+++ b/apps/ocpp-server/migrations/20260821140000-drop-connector-number-foreign-keys.ts
@@ -1,0 +1,58 @@
+// SPDX-FileCopyrightText: 2026 Contributors to the CitrineOS Project
+//
+// SPDX-License-Identifier: Apache-2.0
+'use strict';
+
+/** @type {import('sequelize-cli').Migration} */
+import { QueryInterface, QueryTypes } from 'sequelize';
+
+// Both columns hold a connector number reported by a charging station, not a reference to a
+// Connector row. The associations that made them foreign keys have been removed; these drop the
+// constraints they left behind.
+const CONSTRAINTS: { table: string; name: string; columns: string }[] = [
+  {
+    table: 'StatusNotifications',
+    name: 'StatusNotifications_connectorId_fkey',
+    columns: '"connectorId"',
+  },
+  { table: 'EvseTypes', name: 'EvseTypes_connectorId_fkey', columns: '"connectorId"' },
+];
+
+const constraintExists = async (
+  queryInterface: QueryInterface,
+  table: string,
+  name: string,
+): Promise<boolean> => {
+  const rows = await queryInterface.sequelize.query(
+    `SELECT 1 FROM information_schema.table_constraints
+       WHERE table_schema = 'public' AND table_name = :table AND constraint_name = :name`,
+    { replacements: { table, name }, type: QueryTypes.SELECT },
+  );
+  return rows.length > 0;
+};
+
+export default {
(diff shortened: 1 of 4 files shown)


sjh37:
Closing per citrineos/citrineos#222, for two reasons.

The scenario is a station whose connectors are not yet recorded, which with commissioning done by a person does not arise.

And the constraint this drops does not exist on a migration-built database. After applying every migration on `next` to a scratch Postgres, `pg_constraint` shows `StatusNotifications` with foreign keys on `stationId` and `tenantId` only, and `EvseTypes` with `tenantId` only. The `@BelongsTo(() => Connector, 'connectorId')` on `StatusNotification.ts:72` is model-only and reaches a database through `database.sync`. So the failed insert was a sync-built-database artefact, not a production one - and the integration test in this PR, which builds its schema with `sequelize.sync`, was exercising a constraint the migrations never created.

PR #933, opened 2026-08-21
fix(core): store a composite schedule against the EVSE number the station reported

sjh37:
## What

`CompositeSchedule.evseId` was a foreign key to `Evse.id`, but the value written into it comes straight from a `GetCompositeScheduleResponse`:

```ts
// GetCompositeScheduleResponseOcpp201Handler.ts
await this._chargingProfileRepository.createCompositeSchedule(
  tenantId,
  OCPP2_0_1_Mapper.ChargingProfileMapper.fromCompositeScheduleType(response.schedule),
  message.context.ocppConnectionName,
);
```

`CompositeScheduleType.evseId` is the EVSE number the station calls it. That number restarts at 1 on every station, and `0` names the whole grid connection rather than an EVSE. `Evse.id` is an auto-increment key over every station in the tenant. They are different numbers of the same type.

## Why it matters

Three outcomes, none of them the intended one:

- **evseId 0** - the whole-station limit, which is the one a site with a shared grid connection actually asks for - can never satisfy the constraint. No `Evse` row carries id 0. The insert throws and the schedule is lost.
- **An EVSE number no `Evse` row carries as its id** - a station with six EVSEs on a database whose ids have moved past 6 - also throws.
- **An EVSE number that *does* happen to be some row's id** - normally a row belonging to a different station - inserts cleanly and files the schedule against that other station's EVSE.

The failure is invisible on a freshly seeded database, where ids and EVSE numbers still coincide.

## Fix

Drop the association. The column holds an OCPP-level EVSE number, so it is not a reference:

- `@ForeignKey(() => Evse)` and `@BelongsTo(() => Evse, 'evseId')` removed from `CompositeSchedule`, along with the matching `@HasMany` on `Evse`.
- Migration drops `CompositeSchedules_evseId_fkey`.

Nothing reads the navigation - `CompositeScheduleDto` has no `evse` property and no query includes it - and the Drizzle schema already declares `evseId` as a plain integer with no reference, so this brings the two layers into agreement.

## Test

`packages/core/test/handlers/responses/2/CompositeScheduleEvseReference.integration.test.ts` - testcontainers-backed, runs the real handler against a real Postgres. A neighbouring station is commissioned first so that database ids no longer line up with EVSE numbers.

Against the unpatched model:

```
× stores the schedule for the whole charging station
× stores the schedule for an EVSE no other station has a row for
× does not tie the reported EVSE number to an EVSE row
  → SequelizeForeignKeyConstraintError: insert or update on table "CompositeSchedules"
    violates foreign key constraint "CompositeSchedules_evseId_fkey"
```

After: 3 passed.


thanaParis (changes requested):
this does not 'fix' the issue, it codifies it. the issue is that the evse is not being correctly associated with the composite schedule, when it can be (as noted, when evseId is 0, there's no relevant evse).

sjh37:
@thanaParis You are right, and I have reworked it rather than argued.

Dropping the foreign key recorded the ambiguity instead of resolving it. The schedule does belong to
an EVSE whenever the station names one, and the CSMS has everything it needs to work out which.

**What it does now**

`CompositeSchedule.evseId` stays a foreign key to `Evse` and becomes nullable. `createCompositeSchedule`
resolves the number the station reported against that station's own rows -
`ocppConnectionName` plus `evseTypeId` - so the schedule is associated with the EVSE it actually
belongs to rather than whichever tenant-wide row happens to hold that id. `evseId` 0 is the grid
connection, so it stores null, and so does a number the station has not reported (logged as a
warning rather than dropped).

Both handlers go through that path, so 1.6 and 2.x get it.

**Migration** makes the column nullable, backfills existing rows by resolving `ocppConnectionName`
plus `evseTypeId`, then clears any remaining value that does not resolve to a real row.

**Tests** rewritten to assert the association rather than the raw number:

- `stores the schedule for the whole charging station without an EVSE` - 0 resolves to null
- `associates the schedule with the EVSE the station named`
- `does not associate the schedule with another station EVSE carrying that number` - the case that
  motivated the PR, now asserted positively
- `stores the schedule without an EVSE when the station names one it has not reported`

Four green. `CompositeScheduleDto.evseId` is `nullish()` to match.

The in-code commentary is gone from this one too, on the same basis as #969.

thanaParis (approved):

PR #935, opened 2026-08-21
fix(core): identify a device model component by its EVSE

sjh37:
## What

`Components` was unique on `(tenantId, name, instance)`. An OCPP 2.0.1 station names the same component once per EVSE and tells them apart by `ComponentType.evse`, so the EVSE belongs in the key. Without it every EVSE collapses onto one row:

```ts
// findOrCreateEvseAndComponent, before
const [component, componentCreated] = await this.component.readOrCreateByQuery(tenantId, {
  where: { tenantId, name: componentType.name, instance: componentType.instance ?? null },
});
// Note: this permits changing the evse related to the component
if (component.evseDatabaseId !== evse?.databaseId && evse) {
  await this.component.updateByKey(tenantId, { evseDatabaseId: evse.databaseId }, component.get('id'));
}
```

The second EVSE to report finds the first EVSE's row and rewrites its `evseDatabaseId`.

## Why it matters

`VariableAttribute` is keyed on `(stationId, type, variableId, componentId)`. With one component per name, that key no longer distinguishes EVSEs either - a station with six EVSEs keeps **one** value per variable name for the whole station, and each `NotifyReport` overwrites the last. Reading back with `component_evse_id` joins through `Component.evseDatabaseId`, so it matches only whichever EVSE reported most recently.

The code downstream is already written for the correct shape. `StatusNotificationService` looks a connector component up by both EVSE and connector number:

```ts
let components = await this._componentRepository.readAllByQuery(tenantId, {
  where: { tenantId, name: 'Connector' },
  include: [
    { model: EvseType, where: { id: statusNotificationRequest.evseId,
                                connectorId: statusNotificationRequest.connectorId } },
    { model: Variable, where: { name: 'AvailabilityState' } },
  ],
});
...
if (components.length === 0) {
  this._logger.warn('Missing component or variable for status notification. ' +
                    'Status notification cannot be assigned to device model.');
}
```

On a multi-EVSE station that query can only ever match one EVSE, so the rest log that warning and never reach the device model.

## Fix

- `Component` identity becomes `(tenantId, name, instance, evseDatabaseId)`. Both `instance` and `evseDatabaseId` are optional and Postgres treats nulls in a unique constraint as distinct, so the combinations that leave one out get partial indexes, following the pattern already used on `EvseType`.
- `findOrCreateEvseAndComponent` includes the EVSE in the lookup, and the "permits changing the evse" mutation goes with it.
- `findComponentAndVariable` narrows by the component's EVSE, since more than one component can now share a name and `readOnlyOneByQuery` throws on more than one match.

`VariableAttribute` needs no change: once `componentId` differs per EVSE, its existing key is correct.

Widening a unique key can only reduce collisions, so no existing row can violate the new one and the migration cannot fail on live data. Rows written before it keep whichever EVSE reported last; a fresh `GetBaseReport` repopulates them one per EVSE.

## Test

`packages/core/test/dal/layers/sequelize/repository/DeviceModelComponentPerEvse.integration.test.ts` - testcontainers-backed, a real station reporting `EVSE.AvailabilityState` for EVSE 1 and EVSE 2.

Against the unpatched repository:

```
× keeps one component per EVSE
  → expected [ Component{ …(8) } ] to have a length of 2 but got 1
× keeps the state each EVSE reported
  → expected undefined to be 'Available'
× resolves the component and variable for one EVSE
  → expected undefined to be 2
```

After: 3 passed, and the full `packages/core` suite is green (75 files, 1390 tests).

### DIFF
--- a/apps/ocpp-server/migrations/20260822100000-component-identity-includes-evse.ts
+++ b/apps/ocpp-server/migrations/20260822100000-component-identity-includes-evse.ts
@@ -6,14 +6,8 @@
 /** @type {import('sequelize-cli').Migration} */
 import { QueryInterface, QueryTypes } from 'sequelize';
 
-// A charging station names the same component once per EVSE, so the EVSE belongs in a component's
-// identity. Widening the key can only reduce collisions, so no existing row can violate it. Rows
-// written before this migration keep whichever EVSE reported last; a fresh GetBaseReport repopulates
-// them one per EVSE.
 const TABLE_NAME = 'Components';
 
-// Both instance and evseDatabaseId are optional, and Postgres treats nulls in a unique constraint as
-// distinct, so each combination that leaves one out needs its own partial index.
 const OLD_INDEXES = ['components_tenantId_name'];
 const OLD_CONSTRAINTS = ['Components_tenantId_name_instance_key', 'Components_name_instance_key'];
 
@@ -81,8 +75,6 @@ export default {
     }
   },
 
-  // Narrowing the key back fails wherever a station has reported the same component for more than
-  // one EVSE, which is the state it was wrong about in the first place.
   down: async (queryInterface: QueryInterface) => {
     for (const index of NEW_INDEXES) {
       await queryInterface.sequelize.query(`DROP INDEX IF EXISTS "${index.name}"`, {
--- a/apps/ocpp-server/migrations/20260822100000-component-identity-includes-evse.ts
+++ b/apps/ocpp-server/migrations/20260822100000-component-identity-includes-evse.ts
@@ -1,0 +1,110 @@
+// SPDX-FileCopyrightText: 2026 Contributors to the CitrineOS Project
+//
+// SPDX-License-Identifier: Apache-2.0
+'use strict';
+
+/** @type {import('sequelize-cli').Migration} */
+import { QueryInterface, QueryTypes } from 'sequelize';
+
+// A charging station names the same component once per EVSE, so the EVSE belongs in a component's
+// identity. Widening the key can only reduce collisions, so no existing row can violate it. Rows
+// written before this migration keep whichever EVSE reported last; a fresh GetBaseReport repopulates
+// them one per EVSE.
+const TABLE_NAME = 'Components';
+
+// Both instance and evseDatabaseId are optional, and Postgres treats nulls in a unique constraint as
+// distinct, so each combination that leaves one out needs its own partial index.
+const OLD_INDEXES = ['components_tenantId_name'];
+const OLD_CONSTRAINTS = ['Components_tenantId_name_instance_key', 'Components_name_instance_key'];
+
+const NEW_CONSTRAINT = {
+  name: 'Components_tenantId_name_instance_evseDatabaseId_key',
+  columns: '"name", "instance", "evseDatabaseId", "tenantId"',
+};
+
+const NEW_INDEXES = [
+  {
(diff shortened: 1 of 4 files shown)


thanaParis:
> An OCPP 2.0.1 station names the same component once per EVSE and tells them apart by ComponentType.evse, so the EVSE belongs in the key.

This is incorrect. Component.instance is used to differentiate between components at different levels.

> By default, all components are located at the ChargingStation tier, but individual instances of any component can be associated with a specific EVSE, or a specific Connector (on a specific EVSE) by including EVSE or EVSE and Connector identification numbers as part of a component addressing reference.

A given component instance is only associated with up to one evse type.

sjh37:
@thanaParis Thanks for reading it. I think the objection describes what this code does today rather
than what the spec asks for, so let me put the evidence on the record and then leave it with you.

**`instance` is not what carries the EVSE here.** `Component` in this repo is keyed on
`(tenantId, name, instance)`, with a partial unique index for the common case where `instance` is
null - `packages/core/src/dal/layers/sequelize/model/DeviceModel/Component.ts`:

```ts
@Table({
  indexes: [
    {
      unique: true,
      name: 'components_tenantId_name',
      fields: ['tenantId', 'name'],
      where: { instance: null },
    },
  ],
})
```

Note the scope: that index is `tenantId` + `name`. Not per station, and not per EVSE.

**"A given component instance is only associated with up to one evse type" is the bug, not the
rule.** It holds because `findOrCreateEvseAndComponent` looks the row up without the EVSE and then
overwrites the association -
`packages/core/src/dal/layers/sequelize/repository/DeviceModel.ts:234`:

```ts
const [component, componentCreated] = await this.component.readOrCreateByQuery(tenantId, {
  where: {
    tenantId,
    name: componentType.name,
    instance: componentType.instance ? componentType.instance : null,
  },
});
// Note: this permits changing the evse related to the component
if (component.evseDatabaseId !== evse?.databaseId && evse) {
  await this.component.updateByKey(tenantId, { evseDatabaseId: evse.databaseId }, component.get('id'));
}
```

So a six-EVSE station reporting `Connector` for each EVSE with no `instance` produces **one** row,
whose `evseDatabaseId` is whichever EVSE reported last. `VariableAttribute` is keyed on
`componentId`, so it collapses with it: one value per variable name for the whole station. Because
the index is tenant-wide rather than station-wide, two different stations in the same tenant share
that row too.

**On the passage you quoted** - I read it the same way you do, and I think it argues for the change:

> individual instances of any component can be associated with a specific EVSE, or a specific
> Connector (on a specific EVSE) **by including EVSE or EVSE and Connector identification numbers as
> part of a component addressing reference**

If the EVSE number is part of the addressing reference, it belongs in the key. `instance` is
"Name of instance in case the component exists as multiple instances" - a station is free to leave it
null and distinguish by `evse`, which is what the standardised component list does for `Connector`
and `EVSE`, and libocpp does in practice.

**The corroborating symptom is already in this repo.** `StatusNotificationService` looks a `Connector`
component up by EVSE *and* connector number, and logs "Status notification cannot be assigned to
device model" when it finds nothing - which is what happens on any station past the first EVSE,
because only one row can ever match.

If I have misread the addressing model I would genuinely like to know where, and I will drop it. If
it holds, I am happy for you to reopen this, or to re-cut it as a smaller PR that only changes the
lookup key and leaves the migration for a follow-up - your call, and no hard feelings either way.

PR #936, opened 2026-08-21
fix(core): treat a connector's EVSE-scoped number as a number, not a row reference

sjh37:
## What

`Connector.evseTypeConnectorId` is documented as *"the serial int starting at 1 used in OCPP 2.0.1 to refer to the connector, unique per EVSE"*, and its unique key `(evseId, evseTypeConnectorId)` says the same. It was also `@ForeignKey(() => EvseType)`, making it a reference to `EvseTypes.databaseId` - a key over the tenant-wide device-model catalogue.

Every OCPP 2.0.1 path already treats it as a connector number:

```ts
// SequelizeLocationRepository.readConnectorByStationIdAndOcpp201EvseType
where: { tenantId, ocppConnectionName, evseTypeConnectorId: ocpp201EvseType.connectorId }

// SequelizeTransactionEventRepository.createOrUpdateTransactionByTransactionEventAndStationId
where: { tenantId, ocppConnectionName, evseId: evse.id, evseTypeConnectorId: value.evse.connectorId }

// StatusNotificationService.processStatusNotification
matchingEvse?.connectors?.find((c) => c.evseTypeConnectorId === statusNotificationRequest.connectorId)
```

Only `commissionEvseForOcpp16Connector` wrote a database key, and it did so to satisfy the constraint.

## Why it matters

A 2.0.1 station that has not been commissioned cannot record a connector correctly. Where the tenant-wide catalogue holds no row with that `databaseId` - a new tenant, which is what the test sets up - its first `TransactionEventRequest` reaches:

```
SequelizeForeignKeyConstraintError: insert or update on table "Connectors"
violates foreign key constraint "Connectors_evseTypeConnectorId_fkey"
```

`createOrUpdateTransactionByTransactionEventAndStationId` runs inside a Sequelize transaction, so the whole event is rolled back.

Where the connector number *does* happen to be an `EvseTypes.databaseId`, the insert succeeds and the connector points at an unrelated catalogue row - typically one belonging to another station, since `EvseTypes` is tenant-wide.

## Fix

- `@ForeignKey(() => EvseType)` and `@BelongsTo(() => EvseType, 'evseTypeConnectorId')` removed. Nothing read the `evseTypeByConnector` navigation.
- `commissionEvseForOcpp16Connector` returns `1` instead of `EvseType.databaseId`. A 1.6 connector becomes its own single-connector EVSE, so its number within that EVSE is 1.
- Migration drops `Connectors_evseTypeConnectorId_fkey`.

`commissionEvseForOcpp16Connector` runs only for a station with no `Evse` rows, so rows written before this keep their old value; they are read only by the 2.0.1 paths, which never see a 1.6 station.

## Test

`packages/core/test/dal/layers/sequelize/repository/ConnectorEvseScopedNumber.integration.test.ts` - testcontainers-backed, real repositories.

Against the unpatched model:

```
× records the connector a 2.0.1 transaction names on an uncommissioned station
  → SequelizeForeignKeyConstraintError: insert or update on table "Connectors"
    violates foreign key constraint "Connectors_evseTypeConnectorId_fkey"
× commissions an OCPP 1.6 connector with a connector number, not a catalogue key
  → expected 3 to be 1
× does not tie the connector number to a device model catalogue row
```

After: 3 passed, and the full `packages/core` suite is green (75 files, 1390 tests).

### DIFF
--- a/apps/ocpp-server/migrations/20260822110000-drop-connector-evse-type-foreign-key.ts
+++ b/apps/ocpp-server/migrations/20260822110000-drop-connector-evse-type-foreign-key.ts
@@ -6,9 +6,6 @@
 /** @type {import('sequelize-cli').Migration} */
 import { QueryInterface, QueryTypes } from 'sequelize';
 
-// Connectors.evseTypeConnectorId holds the connector's number within its EVSE, which is what an OCPP
-// 2.0.1 station sends. It is not a reference to an EvseType row, and the association that made it one
-// has been removed; this drops the constraint it left behind.
 const TABLE_NAME = 'Connectors';
 const CONSTRAINT_NAME = 'Connectors_evseTypeConnectorId_fkey';
 
@@ -31,8 +28,6 @@ export default {
     }
   },
 
-  // Restoring the constraint fails wherever a stored connector number does not happen to be an
-  // EvseType key, which is the state it was wrong about in the first place.
   down: async (queryInterface: QueryInterface) => {
     if (!(await constraintExists(queryInterface))) {
       await queryInterface.sequelize.query(
--- a/apps/ocpp-server/migrations/20260822110000-drop-connector-evse-type-foreign-key.ts
+++ b/apps/ocpp-server/migrations/20260822110000-drop-connector-evse-type-foreign-key.ts
@@ -1,0 +1,46 @@
+// SPDX-FileCopyrightText: 2026 Contributors to the CitrineOS Project
+//
+// SPDX-License-Identifier: Apache-2.0
+'use strict';
+
+/** @type {import('sequelize-cli').Migration} */
+import { QueryInterface, QueryTypes } from 'sequelize';
+
+// Connectors.evseTypeConnectorId holds the connector's number within its EVSE, which is what an OCPP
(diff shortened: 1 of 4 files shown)


sjh37:
Closing per citrineos/citrineos#222.

The premise was that `Connectors.evseTypeConnectorId` is a foreign key to `EvseTypes.databaseId`. It is in the Sequelize model (`Connector.ts:76`), and only there: on a migration-built database `Connectors` carries foreign keys on `evseId`, `stationId`, `tariffId` and `tenantId`, and its only unique index besides the key is `(stationId, connectorId)`. Neither the `@ForeignKey(() => EvseType)` nor the `evseId_evseTypeConnectorId` unique has a migration behind it, so only a `database.sync` database enforces them.

On a migration-built database, then, an operator entering `1` in the Operator UI field - described in the UI as "the serial integers starting at 1 used in OCPP 2.0.1 to refer to the connector, unique per EVSE" - stores exactly that, and the five reads that compare the column to a reported number are right. What I found is model-versus-migration drift, and dropping the constraint from the model was the wrong direction to reconcile it. If you would like the model brought into line with the migrations instead (the decorator and the unique removed from `Connector.ts` so that `sync` and migrations agree), say so and I will raise that on its own.

PR #937, opened 2026-08-21
fix(core): give a 2.0.1 connector a station-wide number of its own

sjh37:
> Stacked on #936, which removes the foreign key that this hits first. The diff against `next` therefore contains both commits; review the second one.

## What

`createOrUpdateTransactionByTransactionEventAndStationId` created a `Connector` from an OCPP 2.0.1 transaction event. An OCPP 2.0.1 `connectorId` is scoped to its EVSE, so every EVSE of a station numbers its first connector 1. `Connector.connectorId` is the station-wide OCPP 1.6 numbering and is unique per `(stationId, connectorId)`.

Two separate defects in the same method:

**Second EVSE collides.** The new-transaction branch wrote the EVSE-scoped number straight into the station-wide column:

```ts
defaults: { connectorId: value.evse.connectorId },
```

EVSE 1 connector 1 takes `connectorId = 1`. EVSE 2 connector 1 asks for `connectorId = 1` too:

```
SequelizeUniqueConstraintError: Validation error
```

The whole method runs inside `this.s.transaction(...)`, so the second EVSE's session cannot start at all.

**Existing-transaction branch inserted a null.** The branch that attaches a connector to a transaction already in flight passed no `defaults`:

```ts
const [connector] = await this.connector.readOrCreateByQuery(tenantId, {
  where: { tenantId, ocppConnectionName, evseId: evse.id,
           evseTypeConnectorId: value.evse.connectorId },
  include: [Tariff],
});
```

`Connector.connectorId` is `allowNull: false`:

```
SequelizeValidationError: notNull Violation: Connector.connectorId cannot be null
```

That is the path a station takes when it names the connector on a later event than the first.

## Fix

Both branches go through one helper. A connector the CSMS has not recorded before gets the next number free on its station, rather than the number its EVSE calls it. The lookup key is unchanged - `(ocppConnectionName, evseId, evseTypeConnectorId)` - so the allocation happens once per connector and later events find the same row.

The helper also sets `timestamp` on creation. `Connector.timestamp` has a column getter that calls `.toISOString()` unconditionally, so a row created without one throws as soon as anything reads it.

## Test

`packages/core/test/dal/layers/sequelize/repository/MultiEvseTransactionConnector.integration.test.ts` - testcontainers-backed, a station that has not been commissioned.

Against the unpatched repository (on top of #936):

```
× starts a transaction on a second EVSE
  → SequelizeUniqueConstraintError: Validation error
× records a connector for a transaction whose first event named no connector
  → notNull Violation: Connector.connectorId cannot be null
```

After: 2 passed, and the full `packages/core` suite is green (76 files, 1392 tests).

### DIFF
--- a/apps/ocpp-server/migrations/20260822110000-drop-connector-evse-type-foreign-key.ts
+++ b/apps/ocpp-server/migrations/20260822110000-drop-connector-evse-type-foreign-key.ts
@@ -6,9 +6,6 @@
 /** @type {import('sequelize-cli').Migration} */
 import { QueryInterface, QueryTypes } from 'sequelize';
 
-// Connectors.evseTypeConnectorId holds the connector's number within its EVSE, which is what an OCPP
-// 2.0.1 station sends. It is not a reference to an EvseType row, and the association that made it one
-// has been removed; this drops the constraint it left behind.
 const TABLE_NAME = 'Connectors';
 const CONSTRAINT_NAME = 'Connectors_evseTypeConnectorId_fkey';
 
@@ -31,8 +28,6 @@ export default {
     }
   },
 
-  // Restoring the constraint fails wherever a stored connector number does not happen to be an
-  // EvseType key, which is the state it was wrong about in the first place.
   down: async (queryInterface: QueryInterface) => {
     if (!(await constraintExists(queryInterface))) {
       await queryInterface.sequelize.query(
--- a/apps/ocpp-server/migrations/20260822110000-drop-connector-evse-type-foreign-key.ts
+++ b/apps/ocpp-server/migrations/20260822110000-drop-connector-evse-type-foreign-key.ts
@@ -1,0 +1,46 @@
+// SPDX-FileCopyrightText: 2026 Contributors to the CitrineOS Project
+//
+// SPDX-License-Identifier: Apache-2.0
+'use strict';
+
+/** @type {import('sequelize-cli').Migration} */
+import { QueryInterface, QueryTypes } from 'sequelize';
+
+// Connectors.evseTypeConnectorId holds the connector's number within its EVSE, which is what an OCPP
+// 2.0.1 station sends. It is not a reference to an EvseType row, and the association that made it one
+// has been removed; this drops the constraint it left behind.
+const TABLE_NAME = 'Connectors';
+const CONSTRAINT_NAME = 'Connectors_evseTypeConnectorId_fkey';
+
+const constraintExists = async (queryInterface: QueryInterface): Promise<boolean> => {
+  const rows = await queryInterface.sequelize.query(
+    `SELECT 1 FROM information_schema.table_constraints
+       WHERE table_schema = 'public' AND table_name = :table AND constraint_name = :name`,
+    { replacements: { table: TABLE_NAME, name: CONSTRAINT_NAME }, type: QueryTypes.SELECT },
+  );
+  return rows.length > 0;
+};
+
+export default {
+  up: async (queryInterface: QueryInterface) => {
+    if (await constraintExists(queryInterface)) {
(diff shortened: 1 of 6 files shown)


yuvenus:
I'm not too clear on what's trying to be achieved here -- isn't this a spec violation, as connectors are NOT station wide in 2.x (but rather EVSE-wide)? Can you point to where in the spec this behavior is specified and why this code doesn't violate it?

sjh37:
You are right that a 2.0.1 connector is numbered within its EVSE, and this PR keeps it that way. The
number it allocates goes into a different column - one that is OCPP 1.6's, not 2.0.1's - and nothing
2.0.1 ever puts it on the wire.

Two columns, two numbering schemes, both on `Connector`
(`packages/core/src/dal/layers/sequelize/model/Location/Connector.ts` on `next`):

```ts
declare connectorId: number;            // :74  "the serial int starting at 1 used in OCPP 1.6 to refer to the connector, unique per Charging Station"
declare evseTypeConnectorId?: number;   // :82  "the serial int starting at 1 used in OCPP 2.0.1 to refer to the connector, unique per EVSE"
```

`evseTypeConnectorId` is the 2.0.1 one and this PR does not touch it. The lookup key is unchanged -
`(tenantId, ocppConnectionName, evseId, evseTypeConnectorId)` - and it still stores exactly the
number the station reported, scoped to its EVSE. The only change is the `defaults`, which is what
gets written to `connectorId` when the row is created.

The 2.0.1 message path already uses the EVSE-scoped column and is untouched by this PR -
`packages/ocpi-base/src/util/ocppCommandHandlers/OCPP2_0_1_CommandHandler.ts:201-203`:

```ts
const unlockConnectorRequest: OCPP2_0_1.UnlockConnectorRequest = {
  evseId: evseTypeId,
  connectorId: evseTypeConnectorId,
};
```

The only reader of `Connector.connectorId` that reaches an OCPP message is
`OCPP1_6_CommandHandler.resolveOcpp16ConnectorId` at `:147-157` of the sibling file, which is the
1.6 handler and is named for it.

So the spec violation is the code as it stands, not the change. `Connector.connectorId` is
`allowNull: false` with a unique index on `(stationId, connectorId)`, and
`createOrUpdateTransactionByTransactionEventAndStationId` was writing the EVSE-scoped number
straight into it:

```ts
defaults: { connectorId: value.evse.connectorId },
```

EVSE 1 connector 1 takes `connectorId = 1`; EVSE 2 connector 1 asks for `1` as well and gets
`SequelizeUniqueConstraintError`, inside the event's transaction, so the second EVSE's first session
cannot start. That is the conflation of the two schemes. Allocating the next free station-wide
number for the 1.6 column instead keeps them separate, which is what their two comments already say
should be true.

There is an alternative if you would prefer it: `Connector.connectorId` is meaningless for a 2.0.1
station, so it could become nullable rather than allocated. That is a migration and a wider change -
the unique index and `allowNull: false` both go - which is why I did not reach for it first, but I
am happy to take it that way instead.

sjh37:
Closing per citrineos/citrineos#222. The collision this fixes is on `Connectors (stationId, connectorId)`, and only the CSMS's own `readOrCreateByQuery` defaults in `createOrUpdateTransactionByTransactionEventAndStationId` can produce it - that is, only when the CSMS is creating connector rows a human has not entered. With commissioning done in the Operator UI or SQL, as @yuvenus describes, the `Evse` and `Connector` rows exist before the first `TransactionEventRequest` and the create branch never runs. That makes this a fix for the auto-commissioning path, which is not the default.

If the ungated create path itself is unwanted on a commissioned estate (the sibling branch at `TransactionEvent.ts:147` still inserts a row with no `connectorId` when a station names the connector on a later event than the first), that is a separate question I am happy to raise as one.

PR #938, opened 2026-08-21
fix(core): commission an OCPP 2.0.1 EVSE and connector on first sight

sjh37:
> Stacked on #932 and #936, both of which remove foreign keys this path hits before reaching the connector at all. The diff against `next` carries all three commits; review the last one.

## What

OCPP 1.6 commissions an EVSE and a connector the first time an unknown station reports one:

```ts
// StatusNotificationService.processOcpp16StatusNotification
if (!matchingEvse) {
  if (!connection?.allowUnknownChargingStations) { throw ... }
  const commissioned = await this._locationRepository.commissionEvseForOcpp16Connector(...);
  connector.evseId = commissioned.evseId;
  connector.evseTypeConnectorId = commissioned.evseTypeConnectorId;
}
```

OCPP 2.0.1 had no equivalent. With no matching `Evse` row the connector was left with no `evseId`, and the write was skipped:

```ts
if (connector.evseId != null) {
  await this._locationRepository.createOrUpdateConnector(tenantId, connector);
} else {
  this._logger.warn(`Could not resolve evseId for connector ... Skipping connector update.`);
}
```

## Why it matters

A 2.0.1 station's connectors came into being only through the transaction path, so every status notification before the first session on each connector was dropped along with the state it carried. On a multi-EVSE station that is one silent drop per connector, and the connector stays absent until someone plugs in.

The `allowUnknownChargingStations` gate was also asking the wrong question:

```ts
const connectorExists = chargingStation.evses?.some((evse) =>
  evse.connectors?.some((c) => c.connectorId === statusNotificationRequest.connectorId),
);
```

`Connector.connectorId` is the station-wide OCPP 1.6 numbering; the request carries the number scoped to its EVSE. The comparison matches on stations where the two happen to coincide and misses everywhere else, in both directions.

## Fix

- New `commissionEvseForOcpp201Connector` on the location repository, mirroring the 1.6 one. It creates the `EvseType` catalogue entry and the station's `Evse`, and returns the next connector number free on that station - a 2.0.1 `connectorId` is scoped to its EVSE, so every EVSE numbers its first connector 1 and it cannot be used station-wide.
- `processStatusNotification` resolves the reported connector; if it does not resolve, it commissions behind the same gate 1.6 uses. The gate now hangs off whether the connector resolved, which is the question it was asking.

## Test

`packages/core/test/modules/Transactions/module/Ocpp201StatusNotificationCommissioning.integration.test.ts` - testcontainers-backed, a real station with nothing commissioned.

Against the unpatched service (on top of #932 and #936):

```
× records the EVSE and connector it names          → expected [] to deeply equal [ 1 ]
× gives each EVSE connector its own station-wide number → expected [] to have a length of 2 but got +0
× reports the same connector twice without adding a second row → expected +0 to be 1
✓ refuses an unknown connector when the server does not allow unknown stations
```

After: 4 passed, and the full `packages/core` suite is green (77 files, 1397 tests).

### DIFF
--- a/apps/ocpp-server/migrations/20260821140000-drop-connector-number-foreign-keys.ts
+++ b/apps/ocpp-server/migrations/20260821140000-drop-connector-number-foreign-keys.ts
@@ -1,0 +1,58 @@
+// SPDX-FileCopyrightText: 2026 Contributors to the CitrineOS Project
+//
+// SPDX-License-Identifier: Apache-2.0
+'use strict';
+
+/** @type {import('sequelize-cli').Migration} */
+import { QueryInterface, QueryTypes } from 'sequelize';
+
+// Both columns hold a connector number reported by a charging station, not a reference to a
+// Connector row. The associations that made them foreign keys have been removed; these drop the
+// constraints they left behind.
+const CONSTRAINTS: { table: string; name: string; columns: string }[] = [
+  {
+    table: 'StatusNotifications',
+    name: 'StatusNotifications_connectorId_fkey',
+    columns: '"connectorId"',
+  },
+  { table: 'EvseTypes', name: 'EvseTypes_connectorId_fkey', columns: '"connectorId"' },
+];
+
+const constraintExists = async (
+  queryInterface: QueryInterface,
+  table: string,
+  name: string,
+): Promise<boolean> => {
+  const rows = await queryInterface.sequelize.query(
+    `SELECT 1 FROM information_schema.table_constraints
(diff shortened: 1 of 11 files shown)


sjh37:
Closing on @yuvenus's word in citrineos/citrineos#222: auto-commissioning is a test affordance, not the default, and this PR made it work for multi-EVSE stations - which, as I said on 1 September, is arguing the wrong case. The helper it added only ever runs on that path. The caveat in `StatusNotificationService.ts` ("Do not autocommission multi-evse stations") stands as the documented limit of the test path.

PR #939, opened 2026-08-21
fix(ocpp): check a composite schedule EVSE against the station, not the tenant

sjh37:
> **No longer stacked.** Rebased onto `next` as a single commit - three files, no migration and no model change. It was only ever in that stack because it shared a test file with it; the defect has nothing to do with commissioning.

## What

`GetCompositeScheduleEndpoint` refuses a request for an EVSE the station does not have:

```ts
const evse = await this._deviceModelRepository.findEvseByIdAndConnectorId(tenantId, request.evseId, null);
if (!evse) {
  return { success: false, payload: `EVSE ${request.evseId} not found for station ${ocppConnectionName}.` };
}
```

`findEvseByIdAndConnectorId` reads `EvseType`, the device-model catalogue. That table has no station column - it is keyed on `(tenantId, id, connectorId)` and shared across every station in the tenant. So the guard answers "does *any* station in this tenant have an EVSE numbered N", while its message claims to answer "does *this* station".

## Why it matters

The two differ in both directions on a mixed estate:

- A request naming EVSE 5 against a single-EVSE unit passes the guard, because a six-EVSE charger elsewhere in the tenant put an `EvseType` with id 5 in the catalogue.
- A station whose EVSEs are recorded but whose catalogue entries are not - a `NotifyReport` that never named an EVSE-scoped component - is refused for every EVSE it has.

## Fix

`Evse` is the per-station table (`ocppConnectionName` + `evseTypeId`, the OCPP 2.0.1 EVSE number), and `readEvseByStationIdAndOcpp201EvseId` already asks exactly the question the message describes. The endpoint takes `locationRepository` and uses it.

## Test

`packages/ocpp/test/apis/ocpp/composite-schedule-evse-guard-integration.test.ts` - testcontainers-backed, a six-EVSE charger and a single-EVSE unit in one tenant.

Against the unpatched endpoint:

```
× refuses an EVSE the station does not have, however many the tenant does
  → expected true to be false
✓ sends for an EVSE the station does have
```

After: 2 passed. The existing `smart-charging-endpoints.test.ts` cases move to the new lookup (35 tests), and on `486f1e99` the full suite is green - 142 files, 2303 passed, 4 skipped - with `pnpm run lint` clean.

`readEvseByStationIdAndOcpp201EvseId` already exists on `next` (`repositories.ts:249`, `Location.ts:444`), so nothing else is needed for this one.

### DIFF
--- a/packages/ocpp/src/apis/ocpp/2/smart-charging/get-composite-schedule-endpoint.ts
+++ b/packages/ocpp/src/apis/ocpp/2/smart-charging/get-composite-schedule-endpoint.ts
@@ -15,13 +15,14 @@ import {
   type OCPPVersion,
   type OCPP2_request_types,
 } from '@citrineos/types';
-import type { IDeviceModelRepository } from '@citrineos/dal';
+import type { IDeviceModelRepository, ILocationRepository } from '@citrineos/dal';
 import { OCPP2_PROTOCOLS, ocpp2Schema } from '../schemas.js';
 import { readChargingRateUnitMemberList } from './charging-rate-units.js';
 
 interface Dependencies extends AbstractMessageEndpointDependencies {
   ocppSender: IOcppSender;
   deviceModelRepository: IDeviceModelRepository;
+  locationRepository: ILocationRepository;
 }
 
 export class GetCompositeScheduleEndpoint extends AbstractMessageEndpoint {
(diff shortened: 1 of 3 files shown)


sjh37:

- the endpoint is now
  `packages/ocpp/src/apis/ocpp/2/smart-charging/get-composite-schedule-endpoint.ts`; the conflict
  was the import line only, where `next` had already moved `IDeviceModelRepository` to
  `@citrineos/dal` and this PR adds `ILocationRepository` beside it
- `packages/ocpp/test/apis/ocpp/smart-charging-endpoints.test.ts` merged cleanly - the
  `GetCompositeScheduleEndpoint` block now stubs
  `locationRepository.readEvseByStationIdAndOcpp201EvseId` instead of
  `deviceModelRepository.findEvseByIdAndConnectorId`
- the integration test is
  `packages/ocpp/test/apis/ocpp/composite-schedule-evse-guard-integration.test.ts`

Verified on `486f1e99`: build clean, full suite green (142 test files, 2303 passing / 4 skipped),
lint clean.

thanaParis (approved):

PR #940, opened 2026-08-21
fix(core): compare charging profile validity as a window

sjh37:
## What

Two checks in `SetChargingProfileEndpoint` disagreed about what a validity window is.

**A future `validFrom` was refused.**

```ts
if (validFrom && validFrom.getTime() > now) {
  return { success: false,
           payload: `chargingProfile validFrom ${chargingProfile.validFrom} should not be in the future.` };
}
```

The spec defines the field as *"Point in time at which the profile starts to be valid. If absent, the profile is valid as soon as it is received by the Charging Station."* (carried in `SetChargingProfileRequest.ts` for both 2.0.1 and 2.1). The only `validFrom` this check allowed is one at or before now - which is what omitting the field already means. So the field could never do the one thing it is for: send a profile ahead of the period it covers. For an overnight tariff window or a scheduled site limit that is the whole point.

**The overlap check compared `validTo` alone.**

```ts
const validTo = chargingProfile.validTo ? new Date(chargingProfile.validTo) : null;
if (!validTo) return overlapping;
for (const existedProfile of existedChargingProfiles) {
  const existedValidTo = existedProfile.validTo ? new Date(existedProfile.validTo) : null;
  if (!existedValidTo || existedValidTo.getTime() >= validTo.getTime()) return overlapping;
}
return undefined;
```

`validFrom` is never read. An active profile ending in an hour and a new one ending in two hours take the `false` branch and the new profile is accepted - but both are valid right now, which is exactly what the rejection message forbids: *"No two charging profiles with the same stack level and purpose can be valid at the same time."*

The reverse also happens once profiles can be scheduled: a profile whose window opens after the active one closes would be refused, because only the end times are compared.

## Fix

- `_rejectOnValidity` keeps the "already expired" check and swaps the future-`validFrom` rejection for the constraint that actually has to hold: `validFrom` before `validTo`, so the window can be valid at all.
- `_rejectOnPurpose` compares the two windows. They overlap unless one ends before the other begins, with an absent `validFrom` meaning valid on receipt and an absent `validTo` meaning valid until replaced. A stored profile whose window has already closed no longer blocks a new one.

## Test

`packages/core/test/modules/Api/endpoints/ocpp/smartChargingEndpoints.test.ts`. The case that asserted the old premise - *"refuses a profile whose validFrom is in the future"* - is replaced.

Against the unpatched endpoint:

```
× sends a profile scheduled to start later
  → expected false to be true
× refuses a window that ends before it begins
  → expected 'chargingProfile validFrom 2026-08-21T…' to contain 'should be before validTo'
× refuses a profile whose window overlaps one already active
  → expected true to be false
× sends a profile whose window begins after the active one ends
  → expected false to be true
```

After: 25 passed in that file, and the full `packages/core` suite is green (77 files, 1391 tests).

### DIFF
--- a/packages/core/src/modules/Api/src/module/endpoints/ocpp/2/smartCharging/SetChargingProfileEndpoint.ts
+++ b/packages/core/src/modules/Api/src/module/endpoints/ocpp/2/smartCharging/SetChargingProfileEndpoint.ts
@@ -141,19 +141,35 @@ export class SetChargingProfileEndpoint extends AbstractMessageEndpoint {
     const validFrom = chargingProfile.validFrom ? new Date(chargingProfile.validFrom) : null;
     const validTo = chargingProfile.validTo ? new Date(chargingProfile.validTo) : null;
 
-    if (validFrom && validFrom.getTime() > now) {
-      return {
-        success: false,
-        payload: `chargingProfile validFrom ${chargingProfile.validFrom} should not be in the future.`,
-      };
-    }
     if (validTo && validTo.getTime() <= now) {
       return {
         success: false,
         payload: `chargingProfile validTo ${chargingProfile.validTo} should be in the future.`,
       };
     }
+    if (validFrom && validTo && validFrom.getTime() >= validTo.getTime()) {
+      return {
+        success: false,
+        payload: `chargingProfile validFrom ${chargingProfile.validFrom} should be before validTo ${chargingProfile.validTo}.`,
+      };
+    }
     return undefined;
+  }
+
+  /**
+   * Two profiles are valid at the same time unless one window ends before the other begins. An
+   * absent validFrom means valid on receipt, and an absent validTo means valid until replaced.
+   */
+  private static _overlaps(
+    left: { validFrom?: string | null; validTo?: string | null },
+    right: { validFrom?: string | null; validTo?: string | null },
+    now: number,
+  ): boolean {
+    const from = (profile: { validFrom?: string | null }) =>
+      profile.validFrom ? new Date(profile.validFrom).getTime() : now;
+    const to = (profile: { validTo?: string | null }) =>
+      profile.validTo ? new Date(profile.validTo).getTime() : Number.POSITIVE_INFINITY;
+    return from(left) < to(right) && from(right) < to(left);
   }
 
   private async _rejectOnPurpose(
@@ -204,24 +220,17 @@ export class SetChargingProfileEndpoint extends AbstractMessageEndpoint {
     this._logger.info(
       `Found existing charging profiles: ${JSON.stringify(existedChargingProfiles)}`,
     );
-    if (existedChargingProfiles.length === 0) {
-      return undefined;
-    }
-
-    const overlapping = {
-      success: false,
-      payload:
-        'No two charging profiles with the same stack level and purpose can be valid at the same time.',
-    };
-    const validTo = chargingProfile.validTo ? new Date(chargingProfile.validTo) : null;
-    if (!validTo) {
-      return overlapping;
-    }
-    for (const existedProfile of existedChargingProfiles) {
-      const existedValidTo = existedProfile.validTo ? new Date(existedProfile.validTo) : null;
-      if (!existedValidTo || existedValidTo.getTime() >= validTo.getTime()) {
-        return overlapping;
-      }
+    const now = Date.now();
+    if (
+      existedChargingProfiles.some((existedProfile) =>
+        SetChargingProfileEndpoint._overlaps(chargingProfile, existedProfile, now),
+      )
+    ) {
+      return {
+        success: false,
+        payload:
+          'No two charging profiles with the same stack level and purpose can be valid at the same time.',
+      };
     }
     return undefined;
   }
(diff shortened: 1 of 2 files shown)


devonNewsomDevelop (changes requested):
We need to ensure that the compared profile is not the same profile as the one being compared to, and we need to ensure that the compared profile is not expired. 

Otherwise looks good!

simon-jem:
@devonNewsomDevelop Both were real, and both are fixed in `f861816d` - thank you, I had missed the self-comparison entirely.

**A profile compared against itself.** `createOrUpdateChargingProfile` upserts on `(tenantId, ocppConnectionName, id)`, so re-sending an id the station already holds *replaces* that profile. The overlap query had no `id` exclusion, so the update found its own stored copy and was refused for being valid at the same time as itself. It is now left out of the comparison.

Worth noting this was broken before this PR too - the old code's `if (!validTo) return overlapping` and its `existedValidTo >= validTo` both hit the stored self - so it is a pre-existing bug that my change carried forward rather than one it introduced. Either way it belongs here.

**An expired profile still counting.** My interval test used `validFrom ?? now` as the start, which handled the common case but not the one this PR opens up: now that a `validFrom` in the past is accepted, a new profile starting yesterday genuinely did overlap a profile that expired at midnight - an overlap entirely behind us, which is not a live conflict.

Rather than filter expired rows out separately, both windows are now clamped to `now` before they are compared:

```ts
const from = (profile) =>
  Math.max(profile.validFrom ? new Date(profile.validFrom).getTime() : now, now);
```

Two profiles conflict only over time still ahead of them. That makes "already expired" fall out on its own, and it also covers the case the other way round - a stored profile whose window has not opened yet, which is newly possible now that scheduling ahead works.

**Tests.** Two added, both shown failing against the previous commit:

```
× sends a profile whose window only met the stored one before now
  → expected false to be true
× sends an update to the profile already stored under the same id
  → expected false to be true
```

After: 27 passed in that file, and the full workspace suite is green - 97 files, 2048 tests.

packages/core/src/modules/Api/src/module/endpoints/ocpp/2/smartCharging/SetChargingProfileEndpoint.ts
```diff
@@ -204,24 +216,19 @@ export class SetChargingProfileEndpoint extends AbstractMessageEndpoint {
...
-    for (const existedProfile of existedChargingProfiles) {
-      const existedValidTo = existedProfile.validTo ? new Date(existedProfile.validTo) : null;
-      if (!existedValidTo || existedValidTo.getTime() >= validTo.getTime()) {
-        return overlapping;
-      }
+    const now = Date.now();
+    if (
+      existedChargingProfiles.some(
+        (existedProfile) =>
+          existedProfile.id !== chargingProfile.id &&
```
sjh37, line 223: > we need to ensure that the compared profile is not the same profile as the one being compared to

Fixed here in `f861816d` - you are right, and I had missed it entirely.

`createOrUpdateChargingProfile` upserts on `(tenantId, ocppConnectionName, id)`, so re-sending an id the station already holds *replaces* that profile. The query above had no `id` exclusion, so an update found its own stored copy and was refused for being valid at the same time as itself.

Worth being straight about: this was broken before this PR too - the old code's `if (!validTo) return overlapping` and its `existedValidTo >= validTo` both hit the stored self - so it is a pre-existing bug my change carried forward rather than introduced. It still belongs here.

Covered by `sends an update to the profile already stored under the same id`, which fails against the previous commit with `expected false to be true`.

packages/core/src/modules/Api/src/module/endpoints/ocpp/2/smartCharging/SetChargingProfileEndpoint.ts
```diff
@@ -141,21 +141,33 @@ export class SetChargingProfileEndpoint extends AbstractMessageEndpoint {
...
     return undefined;
   }
 
+  private static _overlaps(
+    left: { validFrom?: string | null; validTo?: string | null },
+    right: { validFrom?: string | null; validTo?: string | null },
+    now: number,
+  ): boolean {
+    const from = (profile: { validFrom?: string | null }) =>
+      Math.max(profile.validFrom ? new Date(profile.validFrom).getTime() : now, now);
```
sjh37, line 165: > we need to ensure that the compared profile is not expired

Fixed here in `f861816d`, by clamping rather than filtering.

This line used to read `profile.validFrom ? ... : now`. That handled the common case but not the one this PR opens up: now that a `validFrom` in the past is accepted, a new profile starting yesterday genuinely did overlap a profile that expired at midnight - an overlap entirely behind us, which is not a live conflict. My earlier test masked it, because it used a profile with no `validFrom` at all.

Rather than filter expired rows out separately, both windows are now clamped to `now` before they are compared: two profiles conflict only over time still ahead of them. "Already expired" then falls out on its own, and it covers the mirror case too - a stored profile whose window has not opened yet, which is newly possible now that scheduling ahead works.

Covered by `sends a profile whose window only met the stored one before now`, which fails against the previous commit with `expected false to be true`.

Full workspace suite green after both: 97 files, 2048 tests.

PR #941, opened 2026-08-21
fix(core): record stop transaction meter values against the session's tenant

sjh37:
## What

An OCPP 1.6 `StopTransaction` carries the session's readings in `transactionData`. `createStopTransaction` persisted them exactly as it received them:

```ts
meterValues.map(async (meterValue) => {
  meterValue.transactionDatabaseId = transactionDatabaseId;
  const createdMeterValue = MeterValue.build(meterValue);
  ...
});
```

`OCPP1_6_Mapper.MeterValueMapper.fromMeterValueType` does not set `tenantId`, and the handler does not either - so `MeterValue.setDefaultTenant` filled it in with `DEFAULT_TENANT_ID`.

The periodic path does set it:

```ts
// MeterValuesRequestOcpp16Handler
const meterValueEntity = OCPP1_6_Mapper.MeterValueMapper.fromMeterValueType(meterValue);
meterValueEntity.tenantId = tenantId;
```

## Why it matters

`readAllMeterValuesByTransactionDataBaseId` goes through `readAllByQuery`, which narrows by `tenantId`. So on any tenant but the default, a session's stop readings are stored but cannot be read back - and for a station that sends no periodic `MeterValues`, those are the only readings the session has. The OCPI CDR mapper builds one charging period per meter value, so those CDRs go out empty.

They are also readable from the default tenant, which they do not belong to.

## Fix

Stamp the tenant where the rows are created, as `createMeterValue` already does. Doing it in the repository rather than the handler covers every caller.

## Test

`packages/core/test/dal/layers/sequelize/repository/StopTransactionMeterValueTenant.integration.test.ts` - testcontainers-backed, a station on a non-default tenant.

Against the unpatched repository:

```
× records the session readings against the tenant the session belongs to
  → expected [] to have a length of 1 but got +0
× does not leave them readable from another tenant
  → expected [ MeterValue{ …(6) } ] to have a length of +0 but got 1
```

After: 2 passed, and the full `packages/core` suite is green (76 files, 1389 tests).

### DIFF
--- a/packages/core/src/dal/layers/sequelize/repository/TransactionEvent.ts
+++ b/packages/core/src/dal/layers/sequelize/repository/TransactionEvent.ts
@@ -781,6 +781,7 @@ export class SequelizeTransactionEventRepository
     if (meterValues.length > 0) {
       await Promise.all(
         meterValues.map(async (meterValue) => {
+          meterValue.tenantId = tenantId;
           meterValue.transactionDatabaseId = transactionDatabaseId;
           const createdMeterValue = MeterValue.build(meterValue);
           createdMeterValue.stopTransactionDatabaseId = stopTransaction.id;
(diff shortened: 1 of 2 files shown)


devonNewsomDevelop (approved):
Thank you

PR #942, opened 2026-08-21
fix(ocpi): let the station choose when StartSession names no EVSE

sjh37:
> **No longer stacked.** #908 replaced typedi property injection in `packages/ocpi-base` with explicit constructor dependencies, which is what #913 was working around, so this is now a single commit on `next` - two files. The test constructs the handler through `Ocpp201CommandHandlerDependencies` and stubs the package barrel, which is the same shape `OCPP1_6_CommandHandler.test.ts` already uses.

## What

`OCPP2_0_1_CommandHandler.sendStartSessionCommand` read the EVSE unconditionally:

```ts
const requestStartTransactionRequest: OCPP2_0_1.RequestStartTransactionRequest = {
  remoteStartId,
  idToken: { ... },
  evseId: Number(EXTRACT_EVSE_ID(startSession.evse_uid!)),
};
```

`evse_uid` is optional on OCPI `StartSession` - `z.string().max(36).nullable().optional()` in `StartSession.ts`, matching OCPI 2.2.1 - and `evseId` is optional on OCPP 2.0.1 `RequestStartTransaction`. So an omitted EVSE is a valid command that leaves the choice to the station.

## Why it matters

Two failures:

- **`evse_uid` omitted.** The non-null assertion silences the compiler, and `EXTRACT_EVSE_ID(undefined)` reaches `undefined.split(...)`:
  `TypeError: Cannot read properties of undefined (reading 'split')`. The throw happens before `sendOCPPMessage`, so the station is never told and no command result is posted either. The partner's `StartSession` gets no answer at all.
- **`evse_uid` that does not carry an EVSE number.** `EXTRACT_EVSE_ID` returns `''` when the uid has no delimiter, `Number('')` is `0`, and OCPP 2.0.1 says `evseId SHALL be > 0`. That was sent to the station as though it were a real EVSE.

## Fix

Mirrors what the 1.6 handler does for `connector_id`: an omitted EVSE is passed on as omitted; a uid that was given and does not name an EVSE reports `FAILED` to the partner rather than reaching the station.

## Test

`packages/ocpi-base/test/util/ocppCommandHandlers/OCPP2_0_1_CommandHandler.test.ts`.

Against the unpatched handler:

```
✓ sends the EVSE the command names
× leaves the EVSE to the station when the command names none
  → Cannot read properties of undefined (reading 'split')
× reports a failure for an EVSE uid it cannot read
  → expected [ { …(2) } ] to have a length of +0 but got 1
```

After: 3 passed, and the full `packages/ocpi-base` suite is green - 16 files, 69 tests - with `pnpm --filter @citrineos/ocpi-base run lint` at 0 errors (the 11 warnings are pre-existing and in files this does not touch).

### DIFF
--- a/packages/ocpi-base/src/util/ocpp-command-handlers/ocpp-201-command-handler.ts
+++ b/packages/ocpi-base/src/util/ocpp-command-handlers/ocpp-201-command-handler.ts
@@ -98,6 +98,15 @@ export class OCPP2_0_1_CommandHandler extends OCPPCommandHandler {
         });
       });
 
+    let evseId: number | undefined;
+    if (startSession.evse_uid !== null && startSession.evse_uid !== undefined) {
+      evseId = Number(EXTRACT_EVSE_ID(startSession.evse_uid));
+      if (!Number.isInteger(evseId) || evseId <= 0) {
+        this.reportEvseNotFound(startSession, tenantPartner, startSession.response_url, commandId);
+        return;
+      }
+    }
+
     const requestStartTransactionRequest: OCPP2_0_1.RequestStartTransactionRequest = {
       remoteStartId,
       idToken: {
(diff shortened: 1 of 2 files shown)


yuvenus (approved):

sjh37:
Rebased onto `486f1e99` after #990. `packages/ocpi-base` was kebab-cased throughout, so the handler
is now `src/util/ocpp-command-handlers/ocpp-201-command-handler.ts` and the test
`test/util/ocpp-command-handlers/ocpp-201-command-handler.test.ts`, with its relative imports
repointed (`model/CommandResult.js` to `model/command-result.js`, `model/DTO/EvseDTO.js` to
`model/dto/evse-dto.js`, and the same for `CommandType` and `ModuleId`).

lint clean. Behaviour unchanged from the approved version.

PR #944, opened 2026-08-24
fix(Server): Use full CostCalculation for Transactions

andreosasLarsson:
## What

An OCPP 2.0.1 transaction's `totalCost` was computed from energy alone. `CostCalculator.calculateTotalCost` took a connector id and kWh and did a single multiply:

```ts
return Money.of(tariff.pricePerKwh, tariff.currency)
  .multiply(totalKwh)
  .roundToCurrencyScale()
  .toNumber();
```

The OCPI CDR path, meanwhile, already had a full pricing implementation (flat session fee + per-kWh + per-minute, with VAT) that had just been moved into `@citrineos/base` as `baseCalculateTotalCost`. So the same session priced two different ways: the `CostUpdated` / `TransactionEvent` response used the kWh-only number, and the CDR used the full one.

## Why it matters

For any tariff with a `pricePerSession` or `pricePerMin`, the running cost sent to the charger (C23 / OCPP 2.1 `CostUpdated`, and the `Ended` `totalCost`) understated the price and did not match the CDR the driver was later billed against. Stations that surface live cost to the driver showed a number the invoice then contradicted.

## Fix

Route Transaction costing through the shared base calculation, so it matches the CDR by construction. `calculateTotalCost` now takes the `Transaction` and reads the tariff for its connector:

```ts
const price = baseCalculateTotalCost(
  transaction.totalKwh,
  transaction.timeSpentCharging / 60, // timeSpentCharging is seconds; base wants minutes
  tariff.pricePerSession,
  tariff.pricePerKwh,
  tariff.pricePerMin,
  tariff.currency,
  tariff.taxRate,
);
return price.excl_vat;
```

Two things surfaced while wiring this up:

- **Unit mismatch.** `timeSpentCharging` is in seconds (OCPP schema); `baseCalculateTotalCost` expects minutes. Passing it straight through inflated time cost 60×, so it is now divided by 60.
- **Float rounding in base.** `baseCalculateEnergyCost` / `baseCalculateTimeCost` multiplied in native floating point before rounding *down*, so exact values fell a cent short (`0.09 * 20` → `1.7999999998` → floored to `1.79`). The multiply now happens inside `buildPrice` on `Money` (decimal), via a `quantity` parameter.

Callers (`TransactionEventRequestOcpp2Handler`, `CostNotifier`) pass the transaction instead of `(connectorId, totalKwh)`. Costs remain VAT-exclusive, unchanged from before.

## Test

`packages/core/test/modules/Transactions/module/CostCalculator.test.ts` - updated to the transaction-based signature and extended to cover time cost (with the seconds→minutes conversion), per-session cost, the combined sum, and the `connectorId` / `totalKwh` / `timeSpentCharging` null guards.

The energy-only cases caught the base rounding bug against the pre-patch base function:

```
× should calculate cost using provided kWh
  → expected 1.79 to be 1.8
  → expected 9.39 to be 9.4
```

After: `@citrineos/base` cost (18), core `CostCalculator` (24) + `CostNotifier` (3), and `ocpi-base` (9) all green; full `packages/core` suite passes (the only failure is the `security-event` e2e, which needs a live server and is unrelated).

## disclaimer
Code has been written by hand, using Claude Opus 4.8 for verification and updating tests
Pull request description has been written by Claude Opus 4.8


thanaParis:
Was this possibly fixed by https://github.com/citrineos/citrineos-core/pull/903 ?

andreosasLarsson:
> Was this possibly fixed by #903 ?

looking at the code briefly it looks to only affect OCPI Cost calculations, this PR updates the calculations for the OCPP session, IE what is stored in the Transaction database and any other cost calculation for ongoing sessions

I am working on implementing "California pricing" for OCPP 1.6 where these calculations also will be used to ensure we provide the correct costs.

thanaParis (approved):

PR #945, opened 2026-08-24
fix(core): allow deleting a charging station that has history

biubiuzhang:
## The problem

A charging station cannot be deleted once anything references it — which, in practice, means any station that has ever sent a message.

`20260427000000-rename-charging-station-columns` puts this trigger on sixteen child tables (`OCPPMessages`, `StatusNotifications`, `Transactions`, `Connectors`, `Evses`, `EventData`, `VariableAttributes`, …):

```sql
BEFORE INSERT OR UPDATE ON "<table>" FOR EACH ROW
WHEN (NEW."stationId" IS NULL)
EXECUTE FUNCTION populate_station_id()
```

`populate_station_id()` resolves the id from `ocppConnectionName` + `tenantId` and raises when it cannot. Every one of those foreign keys is `ON DELETE SET NULL`, so the two mechanisms contradict each other: deleting a station sets `stationId` to `NULL` on its child rows, the trigger fires, the station it would look up has already gone, and the delete aborts.

```
ERROR:  No ChargingStation found with ocppConnectionName=CS-001 and tenantId=1
CONTEXT:  PL/pgSQL function populate_station_id() line 8 at RAISE
```

Through Hasura — the operator UI's **Delete** button — it surfaces as:

```json
{"message": "database query error",
 "internal": {"error": {"status_code": "P0001",
   "message": "No ChargingStation found with ocppConnectionName=CS-001 and tenantId=1"},
   "statement": "WITH \"_mra__ChargingStations\" AS (DELETE FROM \"public\".\"ChargingStations\" WHERE ..."}}
```

## Reproduction

On 2.0.0-beta3 with the tagged migrations applied:

1. Connect a station and let it send a `BootNotification` (any message will do — it needs one row in `OCPPMessages`).
2. Delete it, either from the operator UI's Delete button or directly:

```sql
DELETE FROM "ChargingStations" WHERE "ocppConnectionName" = 'CS-001';
```

3. The delete aborts with `P0001` from `populate_station_id()`.

A station with no history deletes fine, which is why this is easy to miss in a fresh database.

## The fix

Backfilling the id only means anything for a row being written for the first time, so raise on `INSERT` alone:

```sql
IF NEW."stationId" IS NULL AND TG_OP = 'INSERT' THEN
  RAISE EXCEPTION ...
END IF;
```

An `UPDATE` that clears `stationId` is the cascade doing exactly what the foreign key asked for, and leaving the column `NULL` there is the intended outcome. `INSERT` keeps the guarantee it was added for: a child row can still never be written for a station that does not exist.

The migration is `CREATE OR REPLACE FUNCTION`, so nothing needs to be re-attached; `down` restores the previous body.

## Verified

Applied to a live 2.0.0-beta3 database (PostgreSQL + PostGIS). Before the migration, deleting a station that had booted failed with the error above; after it, the same delete succeeds and the child rows keep the `NULL` the foreign key asked for. 190 stations with history were then removed in one statement with no trigger juggling.

Covered by a regression test that boots a station, asserts it has rows in `OCPPMessages`, then sends the same `delete_ChargingStations_by_pk` mutation the operator UI's Delete button sends — red before this change, green after.

## Related

citrineos/citrineos#214 is a different symptom of the same function — an unknown-station `BootNotification` failing on trigger ordering — closed as *"resolved in 2.0 alpha 3"*. The delete path was still broken on beta3 and on `next` at `cb5264c0a`; the `RAISE` has no `TG_OP` guard in either.

### DIFF
--- a/apps/ocpp-server/migrations/20260824120000-allow-deleting-a-charging-station.ts
+++ b/apps/ocpp-server/migrations/20260824120000-allow-deleting-a-charging-station.ts
@@ -1,0 +1,61 @@
+// SPDX-FileCopyrightText: 2026 Contributors to the CitrineOS Project
+//
+// SPDX-License-Identifier: Apache-2.0
+'use strict';
+
+import { QueryInterface } from 'sequelize';
+
+/**
+ * A charging station cannot be deleted once anything references it.
+ *
+ * 20260427000000-rename-charging-station-columns puts a trigger on sixteen child tables:
+ *
+ *   BEFORE INSERT OR UPDATE ... FOR EACH ROW WHEN (NEW."stationId" IS NULL)
+ *   EXECUTE FUNCTION populate_station_id()
+ *
+ * populate_station_id() resolves the id from ocppConnectionName + tenantId and raises
+ * when it cannot. Every one of those foreign keys is ON DELETE SET NULL, so deleting a
+ * station sets stationId to NULL on its child rows, the trigger fires, the station it
+ * would look up has already gone, and the whole delete aborts:
+ *
+ *   ERROR:  No ChargingStation found with ocppConnectionName=CS-001 and tenantId=1
+ *   CONTEXT:  PL/pgSQL function populate_station_id() line 8 at RAISE
+ *
+ * Any station that has ever sent a message has rows in OCPPMessages, so in practice no
+ * commissioned station can be removed. Through Hasura -- the operator UI's Delete
+ * button -- it surfaces as `database query error` with P0001.
+ *
+ * Backfilling the id only means anything for a row being written for the first time, so
+ * the exception is raised on INSERT alone. An UPDATE that clears stationId is the
+ * cascade doing exactly what the foreign key asked for, and leaving the column NULL
+ * there is the intended outcome.
+ *
+ * @type {import('sequelize-cli').Migration}
+ */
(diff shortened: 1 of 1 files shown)


yuvenus (changes requested):
Hello @biubiuzhang, this is by design that you cannot easily delete a charging station that contains data, because in many real world scenarios, several entities require a history of a charging station's messages and usages, regardless of if the charging station is currently in use. Obviously, nothing stops users from going straight into the database and running SQL statements, but the idea is that this is not possible to do through easier means such as the UI because you should have to work to delete important data.

If you are allowed to, can you explain what your real world use case is for this feature? Why not just change the station name to make it obsolete if the concern is that a charging station is no longer used (something like updated the name to "OBSOLETE-stationid")? I understand in a testing case, this can be annoying, but if you're in a test environment, I would think you should have the flexibility to remove the data through the means as described above. 🤔

biubiuzhang:
Thanks for the quick response — that reasoning makes sense and we're happy with the design as it stands.

Our case was a test environment where synthetic stations pile up, not production: we haven't needed to delete a station with real history, and the rename-to-`OBSOLETE-…` route covers the operational case fine. Happy for this to be closed.

yuvenus:
Closing since not allowing deletion of charging stations with any data is by design. 👍

PR #946, opened 2026-08-24
fix(core): let a correlated response inherit its CALL's action over 'NoAction'

biubiuzhang:
## The problem

No `CALLRESULT` or `CALLERROR` row in `OCPPMessages` ever carries an action, so anything that finds a response by action misses it.

A response frame has no action to parse, so the router stores the `NO_ACTION` sentinel for **every** response — this is the success path, not an error path:

```ts
// packages/core/src/modules/OcppRouter/src/module/router.ts
const action =
  parsedMessage.messageTypeId === MessageTypeId.Call
    ? (parsedMessage as Call).action
    : NO_ACTION;
...
await this._webhookDispatcher.dispatchMessageReceived(..., action);
```

Filling that in from the CALL is exactly what the correlation triggers added in `20260730120000` / `20260806120000` exist for, but they do it with:

```sql
NEW."action" := COALESCE(NEW."action", v_call_action);       -- ocpp_correlate_response()
...
"action" = COALESCE("action", NEW."action")                  -- ocpp_correlate_call()
```

`COALESCE` only replaces `NULL`. `'NoAction'` is not `NULL`, so the branch that would write the CALL's action can never take effect — the correlation itself works (`requestMessageId` is set correctly), only the action is left as the sentinel.

## What it breaks

Any query that reaches a response through `action`. Concretely, a vendor `DataTransfer` read-back that polls `OCPPMessages` for `action = 'DataTransfer'` waits for a row that is present, correlated, and stamped `'NoAction'` — so the read-back reports failure while the charger answered correctly. Reporting and analytics that group responses by action see the same hole: every response falls into one `'NoAction'` bucket.

## The fix

`NULLIF` makes the sentinel behave like the `NULL` it stands for, in both directions — the response-arrives-first trigger and its mirror image for a CALL that arrives after its response. Nothing else in either function changes; the bodies are byte-identical to `20260806120000` apart from the two action expressions.

Existing rows are repaired from the CALL they are already correlated to, so a deployment does not have to wait for new traffic to get usable data.

## Verified

On a live database carrying this schema:

- **Before** — every response row stamped `'NoAction'`; a `DataTransfer` read-back polling by action never finds its answer and times out.
- **After** — responses carry their CALL's action. Two hours of traffic on the patched database:

```
 type | action                     | count
------+----------------------------+-------
    3 | BootNotification           |   245
    3 | MeterValues                |   150
    3 | DataTransfer               |   132
    3 | Heartbeat                  |   129
    3 | StatusNotification         |    34
    3 | Authorize                  |    21
```

Zero `'NoAction'` rows among them. Reinstating the plain `COALESCE` reproduces the failure on demand, which is how the regression test for it was proven to bite.

The sentinel is spelled out in the migration rather than imported from `@citrineos/types`: a migration records what the schema did at a point in time and should not change behaviour if that constant is ever renamed.

### DIFF
--- a/apps/ocpp-server/migrations/20260824130000-correlate-response-action-over-noaction.ts
+++ b/apps/ocpp-server/migrations/20260824130000-correlate-response-action-over-noaction.ts
@@ -1,0 +1,142 @@
+// SPDX-FileCopyrightText: 2026 Contributors to the CitrineOS Project
+//
+// SPDX-License-Identifier: Apache-2.0
+'use strict';
+
+import { QueryInterface, QueryTypes } from 'sequelize';
+
+/**
+ * A correlated response never inherits its CALL's action, so no CALLRESULT or CALLERROR
+ * row in OCPPMessages carries one.
+ *
+ * A response frame has no action to parse, so the router stores the NO_ACTION sentinel
+ * for every one of them (`router.ts`: `parsedMessage.messageTypeId === MessageTypeId.Call
+ * ? (parsedMessage as Call).action : NO_ACTION`). Filling that in is exactly what the
+ * correlation triggers are for, but they do it with
+ *
+ *   NEW."action" := COALESCE(NEW."action", v_call_action)
+ *
+ * and COALESCE only replaces NULL. 'NoAction' is not NULL, so the CALL's action is never
+ * written and every response keeps the sentinel.
+ *
+ * Anything that finds a response by action therefore misses it. A vendor DataTransfer
+ * read-back that polls OCPPMessages for `action = 'DataTransfer'` waits for a row that is
+ * present but stamped 'NoAction', and gives up.
+ *
+ * NULLIF makes the sentinel behave like the NULL it stands for, in both directions --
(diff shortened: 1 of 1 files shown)


thanaParis (changes requested):
I don't think adjusting the trigger is the right approach here. The 'NoAction' placeholder is the problem, and this trigger would render it worse than useless; it should be removed not accommodated further. We should be getting rid of it altogether and tweaking the logic to allow for a null value to be used in its place, as the db expects.

biubiuzhang:
Agreed on the direction. Two things I'd like to settle before I rework it, though. First, there's a backlog of rows already stamped `NoAction` — fixing the source still leaves those needing a one-off data repair, which is the backfill this PR already carries. Second, `NO_ACTION` is referenced elsewhere, so removing the constant has a wider blast radius than changing the two trigger expressions. Would you rather I do all of it in this PR?

thanaParis:
closed in favor of https://github.com/citrineos/citrineos-core/pull/977

PR #950, opened 2026-08-24
Feature: Update OCPP Boot PK

lydiazcheng:
## Changes
To fix the tenant leak in the Boot table, we update the PK id to be an auto increment integer instead of ocppConnectionName and add stationId (Charging Station DB ID) as FK.
1. update Boot model and related objects
2. add migration file for Boot table as well VariableAttributes table
3. update PUT /bootConfig endpoint
## Test
1. build and run docker and both tables are migrated Screenshot 2026-08-24 at 3 53 40 PM  Screenshot 2026-08-24 at 3 54 30 PM
2. send a BootNotification for cp002 and boot record was created and stationId is inserted Screenshot 2026-08-24 at 3 59 33 PM Screenshot 2026-08-24 at 4 00 03 PM Screenshot 2026-08-24 at 4 00 46 PM
3. `PUT /bootConfig`: Create a boot when station not exists  Screenshot 2026-08-24 at 4 08 34 PM

### DIFF
--- a/packages/core/src/dal/layers/drizzle/repository/Boot.ts
+++ b/packages/core/src/dal/layers/drizzle/repository/Boot.ts
@@ -75,18 +75,21 @@ export class DrizzleBootRepository
   }
 
   // ──────────── IBootRepository methods ─────────────────────────────
-  // Note that for many of the methods below, we purposely DO NOT use the equivalent Drizzle methods:
-  // callers address a boot record by its station's ocppConnectionName, whereas the
+  // Callers address a boot record by its station's ocppConnectionName, whereas the
   // record is keyed by stationId and its own `id` is an unrelated serial primary key.
 
   async updateByKey(tenantId: number, value: object, key: string): Promise<BootDto | undefined> {
     const stationId = await this.findStationId(tenantId, key);
     if (stationId === undefined) return undefined;
 
+    // Not allowed different tenants or stations in given value
+    const { tenantId: _tenantId, stationId: _stationId, ...safeValue } = value as any;
+
     const rows = (await this.db
       .update(bootTable)
-      .set(toBootEntity(value))
-      .where(and(eq(bootTable.tenantId, tenantId), eq(bootTable.stationId, stationId)))
+      .set(toBootEntity(safeValue))
+      // stationId is a ChargingStations primary key, so it is unique across tenants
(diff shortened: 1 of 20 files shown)


vicky-ahmed (approved):
LGTM

packages/core/src/dal/layers/drizzle/repository/Boot.ts
```diff
@@ -73,15 +75,18 @@ export class DrizzleBootRepository
   }
 
   // ──────────── IBootRepository methods ─────────────────────────────
-  // Note that for many of the methods below, we purposely DO NOT use the equivalent Drizzle methods
-  // because they only accept key as a number, but Boot stores id as a string equal to the
-  // ocppConnectionName of the relevant station.
+  // Note that for many of the methods below, we purposely DO NOT use the equivalent Drizzle methods:
```
yuvenus: Actually, now that the Boot table has a proper primary key... you could totally use the regular methods in DrizzleRepository! You don't have to do any of the silly stuff I had to do here 😄 Is there a reason not to just use this.updateByKey instead?

lydiazcheng: > Actually, now that the Boot table has a proper primary key... you could totally use the regular methods in DrizzleRepository! You don't have to do any of the silly stuff I had to do here 😄 Is there a reason not to just use this.updateByKey instead?

Thank you for the feedback, Venus! Regarding "just use this.updateByKey instead", I am not so clear. Do you mean `updateById` in packages/core/src/dal/layers/drizzle/repository/Base.ts? Or any example would be appreciated.

yuvenus: Yup sorry I meant `updateById`, exactly in the file you linked 😆 For that matter, any of the repository methods in `drizzle/repository/Base.ts` could be used here instead, I think! But if it's not actually that easy to use, then nevermind we can just use the weird methods that I already setup before. 👍

lydiazcheng: Gotcha. I updated it a bit but I think the original codes still make sense since callers of them give `ocppConnectionName` and `tenantId` rather than the new `id` col. I keep the logic in this case.

PR #951, opened 2026-08-25
fix(core): avoid crash when logging websocket upgrade auth failures

LoveKapibarasan:
## Summary
- `WebsocketNetworkConnection._upgradeRequest` logs rejected websocket upgrades via `this._logger.warn('Connection upgrade failed', error)`, passing the raw `Error` object.
- `tslog@4.9.2`'s Node runtime formatter (`prettyFormatLogObj` → `prettyFormatErrorObj` → `getErrorTrace`) calls `error?.stack?.split(...)?.reduce(...)` and then unconditionally does `.map()` on that result in `prettyFormatErrorObj`. When `error.stack` is falsy, `getErrorTrace` returns `undefined`, and the subsequent `.map()` throws `TypeError: Cannot read properties of undefined (reading 'map')`.
- Because this happens inside the `catch` block of `_upgradeRequest`, the `TypeError` is unhandled and crashes the process. In our deployment this reproduced reliably: a charging station repeatedly attempting a duplicate connection triggers `ConnectedStationFilter` to throw `UpgradeAuthenticationError`, which reaches this `logger.warn` call and kills the pod — a `CrashLoopBackOff` triggered purely by an expected auth-rejection path.

## Fix
Pass a plain `{ message, stack }` object instead of the raw `Error` instance, so tslog's `isError()` check is false and it never enters the buggy auto-formatting path. Behavior/log content is unchanged (message and stack are still logged); only the crash is avoided.

## Test plan
- [x] `pnpm run build` in `packages/core` succeeds
- [x] Reproduced the crash locally by throwing `UpgradeAuthenticationError` from a duplicate-connection scenario and confirmed the process no longer crashes after the fix, while message/stack still appear in the log output
- [ ] Would appreciate a maintainer confirming whether upgrading `tslog` is preferred over this workaround — this same pattern (`logger.warn/error(msg, rawError)`) likely exists elsewhere in the codebase and could be worth a broader sweep


thanaParis (changes requested):
> Would appreciate a maintainer confirming whether upgrading tslog is preferred over this workaround — this same pattern (logger.warn/error(msg, rawError)) likely exists elsewhere in the codebase and could be worth a broader sweep

if upgrading tslog would fix this same issue... I think that's a more sustainable fix, please take that approach. Thank you for the catch, this needs to be fixed one way or another.

LoveKapibarasan:
Thanks — taken that approach. The workaround is gone and the call site is back to `this._logger.warn('Connection upgrade failed', error)`; the fix is now a `tslog` upgrade to **4.11.0**, and latest `next` is merged in.

**Root cause confirmed.** 4.9.2's `getErrorTrace` returns `error?.stack?.split(...)` — `undefined` when the Error has no stack — and `prettyFormatErrorObj` calls `.map()` on it unguarded. 4.11.0 replaced that with `buildStackTrace`, which always returns an array.

Verified against the published packages rather than by reading the changelog — same call, three versions:

| tslog | `logger.warn('Connection upgrade failed', stacklessError)` |
| ----- | --------------------------------------------------------- |
| 4.9.2 | `TypeError: Cannot read properties of undefined (reading 'map')` |
| 4.11.0 | no throw |
| 5.1.0 | no throw |

Re-confirmed with 4.11.0 installed in this repo, from `packages/core`.

**Why 4.11.0 and not 5.x:** it stays on the same major, and the only tslog API this codebase imports is `Logger` / `ILogObj`, which is unchanged — typechecked against 4.11.0's own declarations before bumping. 5.x is available if you would rather go there in one step.

Bumped in all five manifests that pin it (`packages/base`, `packages/core`, `packages/ocpi-base`, `apps/ocpp-server`, `apps/ocpi-server`), lockfile resolves to a single `tslog@4.11.0`.

As you noted, this also covers the other `logger.warn/error(msg, rawError)` call sites rather than just this one — no separate sweep needed.

`pnpm -r run build` clean; `packages/core` suite 1515 passed / 4 skipped across 100 files.

PR #954, opened 2026-08-25
fix multi-EVSE connector collision, connectorId/evse fields no longer both required

vicky-ahmed:


thanaParis (changes requested):
conflicts

packages/ocpi-base/src/util/ocpp-command-handlers/ocpp-16-command-handler.ts
```diff
@@ -153,9 +153,11 @@ export class OCPP1_6_CommandHandler extends OCPPCommandHandler {
     if (ocpiConnectorId === null || ocpiConnectorId === undefined) {
       return undefined;
     }
-    return Array.from(chargingStation.connectors || []).find(
-      (connector) => connector.id === Number(ocpiConnectorId),
-    )?.connectorId;
+    return (
+      Array.from(chargingStation.connectors || []).find(
+        (connector) => connector.id === Number(ocpiConnectorId),
+      )?.connectorId ?? undefined
```
devonNewsomDevelop, line 157: This appears to be in a 1.6 file, but the PR title suggests it should only be for 2.0.1? Is this something that is inline with what we would expect on 1.6?

packages/types/src/interfaces/dto/connector-dto.ts
```diff
@@ -17,9 +17,9 @@ export const ConnectorSchemaWithoutParent = BaseSchema.extend({
   id: z.number().int().optional(),
   stationId: z.number().int().optional(),
   ocppConnectionName: z.string(),
-  evseId: z.number().int(),
-  connectorId: z.number().int(),
-  evseTypeConnectorId: z.number().int().optional(),
+  evseId: z.number().int().nullable().optional(),
```
devonNewsomDevelop, line 20: If my understanding is right, this change was for null connectors. Is it intended that we also allow evse to be null and propagated?

devonNewsomDevelop (commented):
Two small questions

packages/types/src/interfaces/dto/connector-dto.ts
vicky-ahmed: Ahh sorry for the misleading PR title. All those fields were already nullable in the db but required on the schema level so there was a mismatch between the schema and db which was creating problems. so as per the ticket this is code only changes.

packages/ocpi-base/src/util/ocpp-command-handlers/ocpp-16-command-handler.ts
vicky-ahmed: again sorry for the misleading title. I guess I should fix the title first lol

vicky-ahmed: The change spans both OCPP 1.6 and 2.0.1 because the connector fields are now nullable repo wide, so all affected readers must handle null I’ll update the title to reflect that.

devonNewsomDevelop (approved):
Looks good to go

PR #956, opened 2026-08-26
Feature/system config refactor

thanaParis:


packages/base/src/interfaces/modules/AbstractModule.ts
```diff
@@ -98,6 +97,7 @@ export abstract class AbstractModule implements IModule {
     logger?: Logger<ILogObj>,
     ocppValidator?: OCPPValidator,
     handlers: AbstractHandler[] = [],
+    excludedActions?: { requests?: CallAction[]; responses?: CallAction[] },
```
yuvenus, line 100: Ah okay very interesting, so instead of excluded requests/responses being a part of the config, you add it as another argument when instantiating the module? Would we ever consider moving this back into the config or nah?

Either way though, I get that whether it's in the config or part of the module itself, you must restart the service, so I'm not married to any particular location for this, considering that it's likely not being used by most people 🤔

thanaParis: Adding it back into the config would be tricky because it doesn't map well to an env var, and excessive config files were what we're trying to do away with.

yuvenus (approved):
Working great locally so far 🤞 Managed to change the heartbeat interval and also manipulate the websocket configs with no issues 🔥

PR #958, opened 2026-08-27
fix(ocpp): build an OCSP request the responder can parse

sjh37:
`GetCertificateStatus` has never reached an OCSP responder. The request the CSMS builds cannot be
encoded, so every call falls into the handler's `catch` and answers `Failed`. The same construction
sits behind the contract-certificate check, where it makes `validateCertificateHashData` return
`NoCertificateAvailable` for every set of hash data a station sends - so Plug and Charge never gets
a real revocation answer either.

### What breaks

`jsrsasign.KJUR.asn1.ocsp.CertID` reads a value-based CertID from `issname` / `isskey` / `sbjsn`.
Both call sites pass `namehash` / `keyhash` / `serial`, so `CertID.tohex()` takes its last branch:

```js
} else { throw new Error("required param members not defined") }
```

Those were jsrsasign 10's names. `@types/jsrsasign` is still pinned at 10.5.15 and still documents
them - its `CertID` constructor type literally lists `{ namehash, keyhash, serial, alg }` - so
TypeScript accepted the call that the installed jsrsasign 11.1.3 rejects, and it will not accept the
correct member names at all. That is why the fix carries one cast.

Three more sit behind that one:

- `alg` is passed through as the OCPP `HashAlgorithmEnumType` value. jsrsasign resolves the digest
  name against an OID table that only holds lower-case entries, so `"SHA256"` throws
  `Name of ObjectIdentifier not defined: SHA256`. Fixing only the member names is not enough.
- Both sites build a bare `Request`, which encodes one `Request` SEQUENCE. RFC 6960 §4.1.1 wants
  `OCSPRequest ::= SEQUENCE { tbsRequest TBSRequest }` with `TBSRequest.requestList` a
  `SEQUENCE OF Request` - two levels further out.
- `sendOCSPRequest` puts `getEncodedHex()` in the body, i.e. the ASCII hex text rather than the DER
  that RFC 6960 Appendix A.1 specifies for `Content-Type: application/ocsp-request`, and reads the
  answer with `response.text()`, which UTF-8-decodes binary DER before `getOCSPResponseInfo` tries
  to read it as hex.

And the response the handler builds sets `ocspResponse`. The field is `ocspResult`
(**M06.FR.03**: *"The CSMS SHALL include the OCSP response data in the OCSPResult field"*), and
**M06.FR.09** requires it DER encoded - the schema then carries it base64 encoded. So even a
successful lookup returned nothing usable, and `additionalProperties: false` means a conformant
station rejects the message outright.

### The fix

`createOcspRequest` in `CertificateUtil`, used by both call sites, that maps the OCPP
`OCSPRequestDataType` onto jsrsasign's CertID members and folds the algorithm name. `sendOCSPRequest`
now posts the DER and returns the responder's DER as hex, which is what
`KJUR.asn1.ocsp.OCSPUtil.getOCSPResponseInfo` expects. The handler base64-encodes that DER into
`ocspResult`.

### Why the existing tests did not catch it

`CertificateAuthority.test.ts` mocks the whole `CertificateUtil` module, so `getEncodedHex()` never
ran and `expect.any(KJUR.asn1.ocsp.Request)` passed on an object that could not be serialised. That test now keeps `createOcspRequest` real and parses the request back with jsrsasign's `OCSPParser`
(through `packages/ocpp/test/utils/ocsp-request-parser.ts`, since `@types/jsrsasign` does not
declare it), so the hash data has to survive a round trip.

`packages/ocpp/test/handlers/requests/2/get-certificate-status-request-ocpp-2-handler.test.ts` is new - the
handler had no coverage at all. It stubs `fetch`, so the whole path from `ocspRequestData` to the
bytes on the wire and back into `ocspResult` runs for real. Against the previous commit it fails with
`expected undefined to be 'MAMKAQCAgYKD'` and a `Failed` status.

Spec references are to OCPP 2.1 Edition 2 (2025-12-03) Part 2, M06 - *Get Certificate Status*; the
2.0.1 wording of M06.FR.03 is the same, and the defect affects both protocol versions.

Full workspace suite green: 92 files, 2003 tests (the six testcontainers suites need Docker and were
not run locally).

### DIFF
--- a/packages/ocpp/src/handlers/requests/2/get-certificate-status-request-ocpp-2-handler.ts
+++ b/packages/ocpp/src/handlers/requests/2/get-certificate-status-request-ocpp-2-handler.ts
@@ -46,11 +46,13 @@ export class GetCertificateStatusRequestOcpp2Handler extends AbstractHandler {
 
     const reqData = message.payload.ocspRequestData;
     try {
-      const ocspRequest = createOcspRequest(reqData);
-      const ocspResult = await sendOCSPRequest(ocspRequest, reqData.responderURL);
+      const ocspResponseHex = await sendOCSPRequest(
+        createOcspRequest(reqData),
+        reqData.responderURL,
+      );
       const response: OCPP2_response_types.GetCertificateStatusResponse = {
         status: OCPP2_1.GetCertificateStatusEnumType.Accepted,
-        ocspResult,
+        ocspResult: Buffer.from(ocspResponseHex, 'hex').toString('base64'),
       };
       await this._ocppSender.sendCallResultWithMessage(message, response);
     } catch (error) {
--- a/packages/ocpp/src/services/certificate/certificate-util.ts
+++ b/packages/ocpp/src/services/certificate/certificate-util.ts
@@ -7,7 +7,7 @@ import { CertificationRequest } from 'pkijs';
 import * as asn1js from 'asn1js';
 import { fromBER } from 'asn1js';
 import { CountryNameEnumType, SignatureAlgorithmEnumType } from '@citrineos/dal';
-import type { CertificateCreate } from '@citrineos/types';
+import type { CertificateCreate, OCPP2_0_1, OCPP2_1 } from '@citrineos/types';
 import jsrsasign from 'jsrsasign';
 import { fromBase64, stringToArrayBuffer } from 'pvutils';
 import moment from 'moment';
@@ -15,7 +15,6 @@ import type { ILogObj } from 'tslog';
 import { Logger } from 'tslog';
 import KJUR = jsrsasign.KJUR;
 import OCSPRequest = jsrsasign.KJUR.asn1.ocsp.OCSPRequest;
-import Request = jsrsasign.KJUR.asn1.ocsp.Request;
 import X509 = jsrsasign.X509;
 import KEYUTIL = jsrsasign.KEYUTIL;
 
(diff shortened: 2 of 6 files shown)


sjh37:
Rebased on `next`, and **the scope of this has shrunk** - worth re-reading rather than picking up the
old diff.

`next` now has `createOcspRequest` with the jsrsasign 11 member names (`issname` / `isskey` /
`sbjsn`) and the lower-cased algorithm, and the handler now returns `ocspResult` rather than
`ocspResponse`. That is two of the five defects fixed, so they are gone from this PR.

What is left, all still true on `next` today:

1. **The CertID is not wrapped in an `OCSPRequest`.** `createOcspRequest` returns a bare
   `KJUR.asn1.ocsp.Request`, which is a single CertID entry - not the RFC 6960 `OCSPRequest`
   structure a responder parses. This PR wraps it: `new OCSPRequest({ reqList: [certId] })`.
2. **The body is hex text, not DER.** `body: ocspRequest.getEncodedHex()` posts the ASCII string
   `"30 82 ..."` under `Content-Type: application/ocsp-request`. Now sends the bytes.
3. **The response is read with `response.text()`**, so a DER body comes back mangled through UTF-8
   decoding. Now read as an ArrayBuffer and hex-encoded.
4. **M06.FR.09**: `ocspResult` carries the OCSPResponse base64 encoded. Now encoded.

I kept the name `createOcspRequest` rather than introducing a second function, so the diff is a
change of that function's body and return type rather than a rename - `buildOCSPRequest` is gone.
The test that mocked it globally is gone too; `CertificateAuthority.test.ts` already runs the real
implementation locally in the one test that needs it, which is the better shape.

Green here: `CertificateUtil.test.ts`, `CertificateAuthority.test.ts` and
`GetCertificateStatusRequestOcpp2Handler.test.ts`, 28 tests.

sjh37:

- `CertificateUtil` is now `packages/ocpp/src/services/certificate/certificate-util.ts`; the
  conflict was in the import block, and the `OCPP2_0_1` / `OCPP2_1` types this PR needs are folded
  into the existing `@citrineos/types` import rather than added as a second one
- the new handler test is
  `packages/ocpp/test/handlers/requests/2/get-certificate-status-request-ocpp-2-handler.test.ts`
- the `CertificateUtil` suite reaches the functions through `@/services/index.js`, which is where
  `next` moved them

Scope is unchanged from the last update: the CertID is still not wrapped in an RFC 6960
`OCSPRequest`, the body still goes out as hex text rather than DER, the response is still read with
`response.text()`, and M06.FR.09's base64 encoding is still missing.

Verified on `486f1e99`: build clean, full suite green (142 test files, 2310 passing / 4 skipped),
lint clean.

thanaParis (approved):

sjh37:
Rebased onto `c1bd4ca`, **and the red CI is fixed** - worth a glance even though this is approved.

The build and tests checks were failing on `Property 'OCSPParser' does not exist on type 'typeof ocsp'`: `@types/jsrsasign` does not declare `OCSPParser`, which jsrsasign 11 ships and which three of these tests use to read the encoded request back. The cast now lives in one place, `packages/ocpp/test/utils/ocsp-request-parser.ts`, and the tests call `parseOcspRequestHex`. Comments trimmed to the two that record a workaround.

Verified on `c1bd4ca`: build clean, full suite green (327 files, 4378 passing / 8 skipped), lint 0 errors.

sjh37:
Heads-up for whoever merges: this PR and #929 both edit `packages/ocpp/src/services/certificate/certificate-util.ts` (`sendOCSPRequest` here, the URL guard there), and git reports a conflict between the two branches. Merge in either order - whichever lands second, I will rebase the same day so it is one click again.

PR #959, opened 2026-08-27
fix(base): stop the schema validator refusing decimals OCPP allows

sjh37:
A charging limit of **8.1 A** is refused with `FormatViolation`. That is the value OCPP 2.0.1 gives
as its own example of an accepted limit - `ChargingSchedulePeriodType.limit`, *"Accepts at most one
digit fraction (e.g. 8.1)"* - and 134 of the 401 one-decimal limits between 0 and 40 A go the same
way.

### Why

OCPP states the one-decimal constraint as `"multipleOf": 0.1`, and there are 23 of those across the
1.6, 2.0.1 and 2.1 schemas. Ajv checks `multipleOf` by dividing and testing the quotient for
integrality, in binary floating point:

```
8.1 / 0.1 === 81.00000000000001
```

so the check fails. Neither Ajv instance in `OCPPValidator` sets `multipleOfPrecision`, which is
exactly the option Ajv provides for this: it rounds the quotient to N significant digits before
testing it. Six is far above the one decimal place the constraint expresses, and far below the point
at which a genuinely two-decimal value would start to slip through.

Measured against `next`, over `0.0` to `40.0` in steps of `0.1`:

| | one-decimal limits refused | 8.1 | 6.25 (must be refused) |
|---|---|---|---|
| `next` | 134 / 401 | refused | refused |
| with `multipleOfPrecision: 6` | 0 / 401 | accepted | refused |

### What it costs today

`OcppRouter._onCall` and `_onCallResult` validate every inbound message against these schemas and
throw `FormatViolation` on failure, so this is not confined to the CSMS-initiated side:

- a station's `ReportChargingProfilesRequest` or `NotifyChargingLimitRequest` carrying an affected
  limit is answered with a CALLERROR;
- a `GetCompositeScheduleResponse` with one is rejected as a CallResult, and the composite schedule
  is dropped;
- `SetChargingProfile` and `RequestStartTransaction` from the operator API are refused by the
  endpoint's body schema.

Which limits are affected is not predictable from the value - `16.3` passes, `8.1` does not - so the
symptom looks intermittent.

### Scope

One option on each of the two Ajv instances `OCPPValidator` builds. It changes nothing for a value
that genuinely violates a `multipleOf`; the test pins `6.25` staying refused alongside every
one-decimal value being accepted.

The Ajv instance in `packages/ocpi-base` is left alone: it validates OCPI DTOs, which carry no
`multipleOf`.

Note that OCPP 2.1 dropped the one-decimal restriction from `limit` and `minChargingRate`
altogether - Part 2 §2.1.4 now allows six decimal places towards the Charging Station - but the 2.1
schemas in this repo still carry `multipleOf: 0.1`, copied across from 2.0.1. That is a separate
defect and a separate PR; this one fixes the constraint for the versions that do impose it.

Full workspace suite green: 91 files, 1997 tests (the six testcontainers suites need Docker and were
not run locally).

### DIFF
--- a/packages/base/src/interfaces/modules/ocpp-validator.ts
+++ b/packages/base/src/interfaces/modules/ocpp-validator.ts
@@ -22,6 +22,17 @@ import {
 } from '@citrineos/types';
 import { OcppError } from '@ocpp/rpc/message.js';
 
+/**
+ * OCPP states its decimal step constraints as multipleOf: 0.1. Ajv checks those by dividing in
+ * binary floating point, where 8.1 / 0.1 is 81.00000000000001 rather than 81, so an unqualified
+ * check refuses about a third of the one-decimal values the spec allows - including 8.1, which
+ * OCPP 2.0.1 gives as its own example of an accepted charging limit. multipleOfPrecision rounds the
+ * quotient to this many significant digits before testing it for integrality; six is far above the
+ * one decimal place the constraint is expressing and far below the point at which a genuinely
+ * two-decimal value would start to pass.
+ */
+const MULTIPLE_OF_PRECISION = 6;
+
 export class OCPPValidator {
   protected _ajv: Ajv;
   protected readonly _logger: Logger<ILogObj>;
@@ -50,6 +61,7 @@ export class OCPPValidator {
         coerceTypes: true, // HTTP query/path params arrive as strings and need coercion
         strict: false,
         allErrors: true,
+        multipleOfPrecision: MULTIPLE_OF_PRECISION,
       });
 
     OCPPValidator.addFormats(ajvInstance);
@@ -76,6 +88,7 @@ export class OCPPValidator {
         strictNumbers: true, // Reject numeric strings where a number is required
         validateFormats: true,
         allErrors: true,
+        multipleOfPrecision: MULTIPLE_OF_PRECISION,
       });
 
     OCPPValidator.addOcppKeywords(ajvInstance);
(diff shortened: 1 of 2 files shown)


sjh37:
Rebased onto `486f1e99` after #990. Clean rebase; `packages/base` was only kebab-cased, so this is
still the two files `src/interfaces/modules/ocpp-validator.ts` and
`test/modules/ocpp-validator.test.ts`.

Verified on `486f1e99`: build clean, full suite green (141 test files, 2304 passing / 4 skipped),
lint clean.

thanaParis (changes requested):
This is an interesting, if niche, catch. I don't like that this fix doesn't completely free us from this issue, just reduces it.

My philosophy here is that we shouldn't be outright rejecting CS messages for this, because the spec is unclear--these schemas come from the spec after all, so you could argue that the behavior of multipleOf is canonical.

my comment in #960 holds here. I think multipleOf should be removed, and the check should only happen in the csms > cs direction--in 2.0.1, to limit to 1 decimal place, in 2.1 to limit to 6. We probably only need one PR to fix this, pick the one you want.

sjh37:
Agreed, and closing this one - #960 is the single PR.

You are right that the Ajv `multipleOfPrecision` setting here only reduces the problem rather than removing it: it makes the divide-in-binary-floating-point check tolerant, but the check itself is still there and still the wrong check to be running on a message the station sent us. Removing `multipleOf` from the schemas, as #960 now does for 1.6, 2.0.1 and 2.1 alike, is the fix; once the keyword is gone there is nothing left for a precision setting to correct.

#960 now also does the CSMS -> CS half you asked for - truncation at serialisation in `MessageRouterImpl.sendCall`, one fraction digit for 1.6/2.0.1 and six for 2.1 - so the whole thing lands in one place.

PR #960, opened 2026-08-27
fix(types): truncate charging rates on the way out instead of rejecting them

sjh37:
A charging rate with more decimal places than the station's OCPP version allows was refused twice
over - by `multipleOf: 0.1` in the schemas, and by a throwing check in `validateChargingProfileType`.
Both refusals were wrong for a message the station sent, and neither stopped the CSMS from sending
one the station would refuse. This PR moves the rule to the one place it belongs: the CSMS to
Charging Station direction, at serialisation.

Reworked after review (see the thread with @thanaParis below); citrineos/citrineos-core#959 is
closed and folded in here.

### What the spec says

OCPP 2.0.1 restricts `ChargingSchedulePeriodType.limit` and `ChargingScheduleType.minChargingRate`
to one decimal place - *"Accepts at most one digit fraction (e.g. 8.1)"* - and 1.6 says the same of
the same two fields. OCPP 2.1 dropped the field-level restriction and replaced it with the general
rule for the `decimal` datatype, Part 2 §2.1.4:

> For data being reported by the Charging Station, the full resolution of the source data must be
> preserved. The decimal sent towards the Charging Station SHALL NOT have more than six decimal
> places.

So the constraint is on what the CSMS **sends**. What the station reports is to be kept as it came.

### What this repo had

- `multipleOf: 0.1` on `limit` and `minChargingRate` in the 1.6, 2.0.1 and 2.1 schemas - 23 sites
  across 15 files. The router validates inbound messages against those schemas, so a station
  reporting `16.25 A` in `ReportChargingProfiles`, `NotifyChargingLimit`,
  `NotifyEVChargingSchedule` or a `GetCompositeSchedule` result was answered with a CALLERROR. On
  2.1 that is a direct violation of §2.1.4; on 2.0.1 and 1.6 it is the CSMS policing the station's
  own resolution. Ajv also checks `multipleOf` by dividing in binary floating point, which is why
  `8.1` itself - the spec's example - was refused (citrineos/citrineos-core#959).
- `validateChargingProfileType` threw on more than one fraction digit, for every version, from the
  operator API handlers.

### The change

**Schemas.** `multipleOf: 0.1` removed from all 23 sites. Nothing a station sends is refused over
resolution any more.

**Validator.** The two fraction-digit throws in `validateChargingProfileType` are deleted, and with
them the last use of `getNumberOfFractionDigit` there.

**Router.** `MessageRouterImpl.sendCall` runs the payload through
`truncateChargingRates(payload, protocol)` before building the `Call`. `limit` and
`minChargingRate` are truncated to one fraction digit for a 1.6 or 2.0.1 station and six for a 2.1
station; everything else in the payload is untouched, and a payload that already fits is returned
as the same object. `sendCall` is the only path that carries a charging schedule from the CSMS to a
station (`SetChargingProfile`, `RequestStartTransaction` / `RemoteStartTransaction`, and 2.1's
`UpdateDynamicSchedule`), so that is the whole outbound surface.

Two judgement calls, both easy to change:

- **Truncate, not round.** `8.15` goes to `8.1`, not `8.2`. For a limit, rounding up would send the
  station a ceiling higher than the operator asked for. It is done on the decimal string, not by
  `Math.trunc(v * 10) / 10`, so `0.3` stays `0.3` rather than becoming `0.2`.
- **Scoped to `limit` and `minChargingRate`, not every decimal.** On 2.0.1 a blanket one-decimal
  truncation would corrupt `totalCost`, meter values and anything else in the payload; the one-digit
  restriction is specific to those two fields, which is exactly the set that carried
  `multipleOf: 0.1`. On 2.1 the §2.1.4 rule is general, so widening it there to every number in the
  payload would be defensible - say the word.

### Tests

- `packages/ocpp/test/util/charging-rate-precision.test.ts` - the truncation itself, including the
  `0.3` and negative cases, the per-version digit counts, and that a payload with nothing to change
  is returned unchanged.
- `message-router-impl.test.ts` - `sendCall` of a `SetChargingProfile` on a 2.0.1 and on a 2.1
  station, asserting what actually went over the network hook.
- `validator.test.ts` - the two tests that asserted the throw are replaced by one asserting a
  finer-resolution profile is accepted.

Verified on `c1bd4ca`: build clean, full suite green (327 files, 4384 passing / 8 skipped), lint
0 errors.

### DIFF
--- a/packages/ocpp/src/modules/ocpp-router/router.ts
+++ b/packages/ocpp/src/modules/ocpp-router/router.ts
@@ -62,6 +62,7 @@ import { Logger } from 'tslog';
 import { v4 as uuidv4 } from 'uuid';
 import { type CallbackUrlNotifier } from './callback-url-notifier.js';
 import { buildConnectionEvent, buildFrameEvent, MessagesExchangeSink } from '@/transport/index.js';
+import { truncateChargingRates } from '@util/index.js';
 
 /**
  * Implementation of the ocpp router
@@ -416,7 +417,7 @@ export class MessageRouterImpl extends AbstractMessageRouter implements IMessage
     const identifier = createIdentifier(tenantId, ocppConnectionName);
     const transactionNamespace = CacheNamespace.Transactions + identifier;
 
-    const message = new Call(correlationId, action, payload);
+    const message = new Call(correlationId, action, truncateChargingRates(payload, protocol));
     if (await this._sendCallIsAllowed(identifier, protocol, message)) {
       if (!(await this._cache.existsAnyInNamespace(transactionNamespace))) {
         const cacheTimestamp = new Date();
--- a/packages/ocpp/src/util/charging-rate-precision.ts
+++ b/packages/ocpp/src/util/charging-rate-precision.ts
@@ -1,0 +1,55 @@
+// SPDX-FileCopyrightText: 2026 Contributors to the CitrineOS Project
+//
+// SPDX-License-Identifier: Apache-2.0
+
+import { OCPPVersion, type OCPPVersionType } from '@citrineos/types';
+
+const CHARGING_RATE_FIELDS = new Set(['limit', 'minChargingRate']);
+
+const OCPP2_1_FRACTION_DIGITS = 6;
+const OCPP2_0_1_FRACTION_DIGITS = 1;
+
+export function maxChargingRateFractionDigits(protocol: OCPPVersionType): number {
+  return protocol === OCPPVersion.OCPP2_1 ? OCPP2_1_FRACTION_DIGITS : OCPP2_0_1_FRACTION_DIGITS;
+}
+
+export function truncateFractionDigits(value: number, digits: number): number {
+  if (!Number.isFinite(value)) {
+    return value;
+  }
+  const text = value.toString();
+  const point = text.indexOf('.');
+  if (point < 0 || text.includes('e') || text.includes('E')) {
+    return value;
+  }
+  if (text.length - point - 1 <= digits) {
+    return value;
+  }
+  return Number(digits === 0 ? text.slice(0, point) : text.slice(0, point + 1 + digits));
+}
+
+export function truncateChargingRates<T>(payload: T, protocol: OCPPVersionType): T {
+  return truncate(payload, maxChargingRateFractionDigits(protocol)) as T;
+}
+
+function truncate(value: unknown, digits: number): unknown {
+  if (Array.isArray(value)) {
+    const items = value.map((item) => truncate(item, digits));
+    return items.some((item, index) => item !== value[index]) ? items : value;
+  }
+  if (value === null || typeof value !== 'object') {
+    return value;
+  }
+
+  let changed = false;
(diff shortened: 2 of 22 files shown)


sjh37:

`validateChargingProfileType` gained an `evseId` parameter upstream when #934 / #939 merged, and this PR inserts `version` before it. Rebased and reconciled; `validator.test.ts` is green (208) and `packages/base` is green (592).

sjh37:
Rebased onto `486f1e99` after #990. Clean rebase. The two endpoint files and the validator now live
under `packages/ocpp` (`src/apis/ocpp/2/...`, `src/util/validator.ts`); the schemas in
`packages/types` are untouched by the refactor.

Verified on `486f1e99`: build clean, full suite green (141 test files, 2307 passing / 4 skipped),
lint clean. Still reads best straight after #959.

thanaParis:
I don't like this approach. throwing an error rather than truncating is needlessly strict, and the truncation should happen at serialization to prevent the router from sending invalid messages rather than out in the message api handler.

thanaParis (changes requested):
see comment.

sjh37:
Reworked to what you asked for, and this PR is now the single one - #959 is closed as superseded.

- `multipleOf: 0.1` is gone from the charging schedule schemas in **1.6, 2.0.1 and 2.1**, so nothing a station sends is rejected over it. That also removes the reason for #959: the Ajv `multipleOf` float bug it worked around cannot bite once the keyword is not there.
- `validateChargingProfileType` no longer throws on a charging rate with too many fraction digits. Both fraction-digit checks are deleted.
- The precision rule now applies only CSMS -> CS, at serialisation: `MessageRouterImpl.sendCall` truncates `limit` and `minChargingRate` through `truncateChargingRates(payload, protocol)` before the `Call` is built - one fraction digit for a 1.6 or 2.0.1 station, six for a 2.1 station.

Two judgement calls worth flagging, since neither was spelled out:

**Truncate, not round.** `8.15` goes to `8.1`, not `8.2`. For a limit that is the safe direction - rounding up would send the station a ceiling higher than the operator asked for. It is done on the decimal string, not by `Math.trunc(v * 10) / 10`, so `0.3` stays `0.3` rather than becoming `0.2`.

**Scoped to `limit` and `minChargingRate`, not every decimal.** For 2.0.1 a blanket one-decimal-place truncation would corrupt `totalCost`, meter values and everything else - the one-digit restriction is specific to those two fields, which is exactly the set that carried `multipleOf: 0.1`. For 2.1, Part 2 §2.1.4 is general ("SHALL NOT have more than six decimal places"), so a wider sweep would be defensible there; I kept the same two fields for symmetry. Say the word if you want 2.1 widened to every number in the payload.

Verified on `c1bd4ca`: build clean, full suite green (327 files, 4384 passing / 8 skipped), lint 0 errors.

sjh37:
Closing on the principle @yuvenus set out on #1095: CitrineOS forwards external data as it is rather than truncating it to protocol limits, and this change truncates charging-rate decimals to the protocol precision.

PR #962, opened 2026-08-27
fix(ocpp): omit company from VatNumberValidationResponse instead of sending null

sjh37:
When a station asks the CSMS to check a VAT number and the number does not resolve, the
`VatNumberValidationResponse` CitrineOS sends fails its own schema.

`company` is an `AddressType`:

```json
"company": { "$ref": "#/definitions/AddressType" }
```

and `AddressType` is `"type": "object"`. The handler sets it unconditionally from whatever
`IVatProvider.getVat` returned, and `getVat` returns `null` for a VAT number that is permanently
invalid - which is the normal outcome of the check that C18.FR.09 describes:

> CSMS validates given VAT number and responds with VatNumberValidationResponse with status =
> Accepted if validated and Rejected when invalid. Optionally the company address can be returned in
> the company field.

So the Rejected answer goes out as `{"vatNumber": "...", "status": "Rejected", "company": null}`, and
a station that validates what it receives answers a CALLERROR rather than telling the driver the VAT
number was not accepted. Nothing catches it on this side: `sendCallResult` does not validate outgoing
messages.

The field is optional, so the fix is to leave it out.

`VatNumberValidationRequestOcpp21Handler.test.ts` is new - the handler had no coverage. One of its
cases runs the response back through `OCPPValidator`, so the shape is pinned against the schema
itself rather than against a hand-written expectation; against the previous commit it fails with
`/company must be object`. It also pins C18.FR.10 (the response carries the evseId the station
asked about), which the handler already gets right.

Full workspace suite green: 92 files, 1998 tests (the six testcontainers suites need Docker and were
not run locally).

### DIFF
--- a/packages/ocpp/src/handlers/requests/2/vat-number-validation-request-ocpp-21-handler.ts
+++ b/packages/ocpp/src/handlers/requests/2/vat-number-validation-request-ocpp-21-handler.ts
@@ -41,7 +41,7 @@ export class VatNumberValidationRequestOcpp21Handler extends AbstractHandler {
 
     const request = message.payload;
 
-    const company = this._vatProvider ? await this._vatProvider.getVat(request.vatNumber) : null;
+    const company = (await this._vatProvider?.getVat(request.vatNumber)) ?? undefined;
 
     const response: OCPP2_1.VatNumberValidationResponse = {
       vatNumber: request.vatNumber,
--- a/packages/ocpp/test/handlers/requests/2/vat-number-validation-request-ocpp-21-handler.test.ts
+++ b/packages/ocpp/test/handlers/requests/2/vat-number-validation-request-ocpp-21-handler.test.ts
@@ -1,0 +1,113 @@
+// SPDX-FileCopyrightText: 2026 Contributors to the CitrineOS Project
+//
+// SPDX-License-Identifier: Apache-2.0
+
+import { beforeEach, describe, expect, it, vi } from 'vitest';
+import {
+  type IMessage,
+  type IVatProvider,
+  DEFAULT_TENANT_ID,
+  OCPPValidator,
+} from '@citrineos/base';
+import {
+  type OcppRequest,
+  EventGroup,
+  MessageOrigin,
+  MessageState,
+  OCPP2_1,
+  OCPP_CallAction,
+  OCPPVersion,
+} from '@citrineos/types';
+import { VatNumberValidationRequestOcpp21Handler } from '@handlers/index.js';
+import { createTestContainer, makeMockOcppSender } from '@test/test-container.js';
+
+function makeMessage<T extends OcppRequest>(payload: T): IMessage<T> {
+  return {
+    context: {
+      tenantId: DEFAULT_TENANT_ID,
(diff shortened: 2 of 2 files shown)


sjh37:
Rebased onto `486f1e99` after #990. The handler and its test are now
`packages/ocpp/src/handlers/requests/2/vat-number-validation-request-ocpp-21-handler.ts` and
`packages/ocpp/test/handlers/requests/2/vat-number-validation-request-ocpp-21-handler.test.ts`; the
only edit beyond the move was the test container import, `@test/testContainer.js` to
`@test/test-container.js`.

Verified on `486f1e99`: build clean, full suite green (142 test files, 2305 passing / 4 skipped),
lint clean.

packages/ocpp/src/handlers/requests/2/vat-number-validation-request-ocpp-21-handler.ts
```diff
@@ -43,13 +43,15 @@ export class VatNumberValidationRequestOcpp21Handler extends AbstractHandler {
 
     const company = this._vatProvider ? await this._vatProvider.getVat(request.vatNumber) : null;
 
+    // company is an AddressType, so it is left out rather than sent as null when the VAT number
```
thanaParis: remove this comment.

packages/ocpp/src/handlers/requests/2/vat-number-validation-request-ocpp-21-handler.ts
```diff
@@ -43,13 +43,15 @@ export class VatNumberValidationRequestOcpp21Handler extends AbstractHandler {
 
     const company = this._vatProvider ? await this._vatProvider.getVat(request.vatNumber) : null;
```
thanaParis: Wouldn't replacing 'null' here with 'undefined' be a cleaner fix?

thanaParis (changes requested):

packages/ocpp/src/handlers/requests/2/vat-number-validation-request-ocpp-21-handler.ts
sjh37: Removed.

packages/ocpp/src/handlers/requests/2/vat-number-validation-request-ocpp-21-handler.ts
sjh37: Agreed - done. `getVat` is typed `Promise<AddressType | null | undefined>`, so it is now `(await this._vatProvider?.getVat(...)) ?? undefined` and the response keeps a plain `company,`. An undefined property is dropped by `JSON.stringify`, so nothing goes on the wire; the test asserts that on the serialised response rather than on the object.

sjh37:
Rebased onto `c1bd4ca` and both review points applied - `?? undefined` rather than the conditional spread, and the comment removed.

Verified on `c1bd4ca`: build clean, full suite green (327 files, 4370 passing / 8 skipped), lint 0 errors.

thanaParis (approved):

PR #965, opened 2026-08-27
fix(ocpp): keep key codes out of the log

sjh37:
A key code is a PIN the driver typed on the Charging Station's keypad. OCPP singles it out:

> **C04.FR.04** If an idToken of type keyCode is used → The Charging Station or CSMS SHALL NOT show
> the IdToken in any logging. *key codes should never appear in logs.*

Both OCPP 2.x Authorize handlers write it out twice.

### Where

```ts
this._logger.info(this.createHandlerReceivedMessageLog('AuthorizeRequest'), message, props);
```

The whole inbound message, at **info**, on every `AuthorizeRequest` - so every key code a driver
enters is in the log, valid or not, at a level that is on by default.

```ts
this._logger.warn(`Invalid ID token format`, {
  type: request.idToken.type,
  token: request.idToken.idToken,
  error: tokenValidation.errorMessage,
});
```

And again, explicitly, when the token fails format validation - which for a key code is the mistyped
PIN, the value most worth not keeping. `validateIdToken` reaches this branch for a `KeyCode` carrying
any character outside the identifierString set, so a keypad that offers `#` or `*` in a way the
driver misuses lands here.

### The change

A small `keyCodeRedaction` helper, and both handlers routed through it: the received-message log gets
a copy with the key code replaced, and the invalid-format log gets a placeholder instead of the
value. Nothing else about the logging changes - the test pins that the type and the reason are still
there, so the message stays diagnosable:

```
Invalid ID token format { type: 'KeyCode', token: '[redacted key code]',
  error: 'KeyCode tokens must contain only letters, numbers, and characters: * - _ = : + | @ .' }
```

Only `KeyCode` is redacted. The other types are not what C04.FR.04 is about, and quietly hiding them
would cost real diagnostics.

### Two related places, deliberately not in this PR

- `OcppRouter._onCall` logs every inbound message, payload included, at **debug**. Same exposure,
  but redacting there means a general hook rather than a two-line change, and it is a decision about
  how the router logs rather than a defect in a handler.
- `WebhookDispatcher` persists every message to `OCPPMessage.payload` (JSONB), so key codes are in
  the database as well as the log. Whether an archive counts as "logging" is arguable, but a CPO
  that reads C04.FR.04 strictly will want that column redacted or the messages not stored, and that
  is a schema and retention decision.

Happy to take either on if you want them, but they are not bugfixes and did not belong in the same
change.

Spec reference is to OCPP 2.0.1 Edition 4 (2025-12-03) Part 2, C04 - *Authorization using a
PIN-code*; OCPP 2.1 Edition 2 carries C04.FR.04 unchanged, and both handlers are fixed.

Full workspace suite green: 92 files, 1997 tests (the six testcontainers suites need Docker and were
not run locally).

### DIFF
--- a/packages/ocpp/src/handlers/requests/2/authorize-request-ocpp-201-handler.ts
+++ b/packages/ocpp/src/handlers/requests/2/authorize-request-ocpp-201-handler.ts
@@ -29,7 +29,7 @@ import {
   OCPP2_response_types,
 } from '@citrineos/types';
 import { CertificateAuthorityService } from '@services/index.js';
-import { validateIdToken } from '@util/index.js';
+import { redactKeyCode, redactKeyCodeInMessage, validateIdToken } from '@util/index.js';
 import {
   type IAuthorizationRepository,
   type IDeviceModelRepository,
@@ -71,7 +71,11 @@ export class AuthorizeRequestOcpp201Handler extends AbstractHandler {
     message: IMessage<OCPP2_request_types.AuthorizeRequest>,
     props?: HandlerProperties,
   ): Promise<void> {
-    this._logger.info(this.createHandlerReceivedMessageLog('AuthorizeRequest'), message, props);
+    this._logger.info(
+      this.createHandlerReceivedMessageLog('AuthorizeRequest'),
+      redactKeyCodeInMessage(message),
+      props,
+    );
 
     const request = message.payload as OCPP2_0_1.AuthorizeRequest;
     const context = message.context;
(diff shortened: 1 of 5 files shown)


yuvenus (approved):

sjh37:

- the redaction helper is now `packages/ocpp/src/util/key-code-redaction.ts`, exported from
  `packages/ocpp/src/util/index.ts`
- both Authorize handlers pick it up from `@util/index.js`. `next` had already split their imports
  so that `CertificateAuthorityService` comes from `@/services/index.js` and the validators from
  `@util/index.js`; this PR adds to the second of those
- the test is `packages/ocpp/test/handlers/requests/2/authorize-id-token-logging.test.ts`, taking
  the repository interfaces from `@citrineos/dal` the way the sibling Authorize suites now do

lint clean. Behaviour unchanged from the approved version.

sjh37:
Rebased onto `c1bd4ca` - the only conflict was the `CertificateAuthorityService` import, which `next` moved from `@/services/index.js` to `@services/index.js`. Comments removed bar the one C04.FR.04 quotation that says why the helper exists.

Verified on `c1bd4ca`: build clean, full suite green (327 files, 4368 passing / 8 skipped), lint 0 errors.

thanaParis:
closing in favor of https://github.com/citrineos/citrineos-core/pull/1035

PR #966, opened 2026-08-27
fix(types): restore the OCPP 2.0.1 idToken length

sjh37:
The OCPP 2.0.1 schemas in this repo allow a 255-character `idToken`. That is OCPP 2.1's bound;
2.0.1 stops at 36, and a 2.0.1 station's own schema stops there too.

### The bound

OCPP 2.0.1 Edition 4 Part 2 types `IdTokenType.idToken` as **`identifierString[0..36]`**, and the
published Part 3 schema says `"maxLength": 36` in all eight places it defines `IdTokenType`.

OCPP 2.1 Edition 2 widened the same field to `identifierString[0..255]`, along with
`AdditionalInfoType.additionalIdToken`. The repo's 2.1 schemas correctly say 255.

The repo's 2.0.1 schemas say 255 as well - but only for `idToken`. `additionalIdToken` in the same
files is still 36, which is what 2.0.1 says, so this is an isolated edit rather than a decision to
run 2.0.1 on 2.1's lengths.

### What it costs

The CSMS does not only receive idTokens, it sends them. `RequestStartTransactionRequest`,
`SendLocalListRequest`, `ReserveNowRequest` and `CustomerInformationRequest` all carry one, and
`AuthorizeResponse` / `TransactionEventResponse` carry `idTokenInfo.groupIdToken`. Those bodies are
validated against this schema at the operator API before being forwarded.

So an operator can create an authorization with a 40-character token, or a group token, and every
message carrying it to a 2.0.1 station is refused by the station with a `FormatViolation` - a remote
start that never starts, or a local authorization list that never installs. Nothing on the CSMS side
reports the cause, because as far as CitrineOS is concerned the token was fine.

Inbound is the milder half: a conformant 2.0.1 station will not send one, but if something does, the
CSMS stores it and then cannot send it back.

### The change

`maxLength` back to 36, in the eight 2.0.1 schemas that define `IdTokenType`, and nothing else - the
diff is eight single-line edits. `additionalIdToken` is already correct and is untouched.

The test asserts the behaviour through `OCPPValidator` (36 accepted, 37 refused, 255 refused) rather
than only the constant, and pins the bound in the five request schemas the CSMS sends.

Spec reference is to OCPP 2.0.1 Edition 4 (2025-12-03) Part 2 and the published Part 3 JSON schemas.

Full workspace suite green: 92 files, 2002 tests (the six testcontainers suites need Docker and were
not run locally).

### DIFF
--- a/packages/base/test/modules/ocpp201IdTokenLength.test.ts
+++ b/packages/base/test/modules/ocpp201IdTokenLength.test.ts
@@ -1,0 +1,54 @@
+// SPDX-FileCopyrightText: 2026 Contributors to the CitrineOS Project
+//
+// SPDX-License-Identifier: Apache-2.0
+
+import { describe, expect, it } from 'vitest';
+import { OCPP2_0_1_CALL_SCHEMA_RECORD } from '../../index.js';
+import { OCPP_CallAction, OCPPVersion } from '@citrineos/types';
+import { OCPPValidator } from '@interfaces/modules/OCPPValidator.js';
+
+/**
+ * OCPP 2.0.1 types IdTokenType.idToken as identifierString[0..36]. OCPP 2.1 widened the same field
+ * to identifierString[0..255]; 2.0.1 did not, and a 2.0.1 station's own schema still stops at 36.
+ */
+const MAX_2_0_1_ID_TOKEN = 36;
+
+describe('OCPP 2.0.1 idToken length', () => {
+  const validator = new OCPPValidator();
+
+  function validateAuthorize(idToken: string) {
+    return validator.validateOCPPRequest(
+      OCPP_CallAction.Authorize,
+      { idToken: { idToken, type: 'Central' } },
+      OCPPVersion.OCPP2_0_1,
+    );
+  }
+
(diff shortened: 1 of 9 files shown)


thanaParis:
This correction does not match reality in the field, which is that the 36 character limit is a problem for station compatibility (which is why the change to the schema was made).

sjh37:
@thanaParis Understood, and that is a good reason - if the 36-character bound was breaking real
stations then the field wins over the printed type. Thanks for saying why it was widened; I could not
tell that from the diff and I would not have opened this if I had known.

One thing worth recording for whoever reads this next, since it is the same field in the other
direction: the CSMS sends idTokens as well as receiving them, in `RequestStartTransaction` and
`SendLocalList`. A token between 37 and 255 characters is now accepted by the operator API and goes
out to the station, which validates it against its own 2.0.1 schema and refuses it with a
`FormatViolation` - and nothing on the CSMS side reports why.

Not asking you to reconsider the schema. If it is worth guarding at all it belongs on the outbound
path, not in the type, and I am happy to leave that until someone actually hits it.

PR #967, opened 2026-08-27
fix(types): close the OCPP 2.0.1 schema classes

sjh37:
Every class in the OCPP 2.0.1 schemas carries `"additionalProperties": true`. The published schemas
say `false` - all 470 of them - and the spec leans on that.

### What the spec says

OCPP 2.0.1 Edition 4, Part 4 - *JSON over WebSockets implementation guide*, Chapter 9
**CustomData Extension**:

> In the JSON schema files all classes have the attribute additionalProperties set to false, such
> that a JSON parser will not accept any other properties in the message. In order to allow for some
> flexibility to create non-standard extensions for experimentation purposes, every JSON class has
> been extended with a "customData" property. This property is of type "CustomDataType", which has
> only one required property: "vendorId" ... However, since it does not have additionalProperties set
> to false it can be freely extended with new properties.

Closed classes plus one open `CustomDataType` is the whole extension mechanism. With every class
open, `customData` has no purpose - anything can be put anywhere.

The repo's **2.1** schemas already say `false`, and Part 4 of 2.1 carries the same chapter word for
word, so this is 2.0.1 alone.

### The numbers

|  | `additionalProperties: false` | `: true` |
|---|---|---|
| published 2.0.1 Part 3 schemas | 470 | 0 |
| this repo, before | 0 | 470 |
| this repo, after | 470 | 0 |

`CustomDataType` carries no `additionalProperties` key in either set, before or after, so the
extension point stays open. After the change every one of the 128 files matches the published schema
on this keyword exactly - the test in the first commit checks that structurally rather than by eye.

The diff is one literal replaced by another, 470 times, and nothing else.

### What it costs today

`OcppRouter._onCall` and `_onCallResult` validate inbound messages against these schemas, so a
station's message carrying a field that is not in the definition is accepted and the field is then
dropped by the mapper. A misspelt field reads as an absent one: a `TransactionEventRequest` with
`evseID` instead of `evse` is stored as a transaction with no EVSE rather than refused.

The outbound direction is worse, because the operator API validates request bodies against the same
schemas with `removeAdditional: 'failing'` - Ajv only strips an additional property when it *causes*
a failure, and with `additionalProperties: true` nothing fails. So a stray field in an API call
survives validation, is copied into the OCPP message, and is sent to a station whose own schema is
closed. The station answers `FormatViolation` and the whole command fails, with nothing on the CSMS
side pointing at the extra field.

### The risk in this change

It is a real tightening: a station that today sends a non-standard field outside `customData` is
accepted and after this is answered with a `FormatViolation`. That is what the spec asks for, and
`customData` is the sanctioned way for such a vendor to keep working, but it is a behaviour change
worth being deliberate about rather than a pure bugfix - so if you would rather stay lenient inbound
and tighten only the operator API, say so and I will split it that way.

I could not find a reason the flip was intentional: it arrived with `fcda3a3a` ("The wumbo types
package commit that has 1.3k changes"), a bulk move, and the 2.1 schemas that moved in the same
commit kept `false`.

Full workspace suite: 97 files, 2046 tests passing, 6 skipped. This run had Docker available, so
the testcontainers-backed integration suites ran too. The one failing file,
`packages/core/test/dal/e2e/security-event.e2e.test.ts`, fails the same way on `next` -
`Command failed: pnpm run db:migrate` - so it is not this change.

### DIFF
--- a/packages/base/test/modules/ocpp201AdditionalProperties.test.ts
+++ b/packages/base/test/modules/ocpp201AdditionalProperties.test.ts
@@ -1,0 +1,91 @@
+// SPDX-FileCopyrightText: 2026 Contributors to the CitrineOS Project
+//
+// SPDX-License-Identifier: Apache-2.0
+
+import { describe, expect, it } from 'vitest';
+import { OCPP2_0_1_CALL_RESULT_SCHEMA_RECORD, OCPP2_0_1_CALL_SCHEMA_RECORD } from '../../index.js';
+import { OCPP_CallAction, OCPPVersion } from '@citrineos/types';
+import { OCPPValidator } from '@interfaces/modules/OCPPValidator.js';
+
+/**
+ * OCPP 2.0.1 Edition 4, Part 4 - JSON over WebSockets, Chapter 9 CustomData Extension:
+ *
+ *   "In the JSON schema files all classes have the attribute additionalProperties set to false,
+ *    such that a JSON parser will not accept any other properties in the message. In order to allow
+ *    for some flexibility to create non-standard extensions for experimentation purposes, every
+ *    JSON class has been extended with a "customData" property. This property is of type
+ *    "CustomDataType", which has only one required property: "vendorId" ... However, since it does
+ *    not have additionalProperties set to false it can be freely extended with new properties."
+ *
+ * So every class is closed and customData is the one way in. CustomDataType itself is the single
+ * exception, and it carries no additionalProperties at all.
+ */
+type Schema = Record<string, unknown>;
+
+function walk(node: unknown, path: string, visit: (node: Schema, path: string) => void): void {
+  if (Array.isArray(node)) {
+    node.forEach((child, i) => walk(child, `${path}[${i}]`, visit));
+    return;
+  }
+  if (node === null || typeof node !== 'object') {
+    return;
+  }
+  const schema = node as Schema;
+  if (schema.type === 'object') {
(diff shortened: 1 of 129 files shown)


thanaParis:
I don't think this is a good idea. It is better practice to report protocol violations in logs and surface them to admins while continuing to provide the best service possible. Additional fields do not prevent the CSMS from performing its duties and should not stand in the way of functionality.

sjh37:
@thanaParis Fair enough, and I take the point - "report the violation, keep serving" is a defensible
house rule, and it is your call to make. Not reopening.

One half of the finding survives that rule, though, and it is not about being strict with stations:

The operator API's Ajv instance is configured `removeAdditional: 'failing'`, which only strips a
property when that property is what *makes* validation fail. With `additionalProperties: true`
everywhere, nothing fails, so nothing is ever stripped - and the operator API copies the validated
body straight into the OCPP message it builds. A mistyped field posted to the API therefore reaches
the station, and the station refuses the whole command with a `FormatViolation`.

So the current setting is not leniency, it is passing the consequence on to the station. Stripping it
is the "continue to provide the best service possible" behaviour, and `packages/ocpi-base` already
uses `removeAdditional: 'all'` for exactly that.

Opened as #975 - one line, inbound station validation untouched, failing test first. That one should
be uncontroversial on your own reasoning; if it is not, say so and I will close it too.

PR #968, opened 2026-08-27
fix(ocpp): ignore a message type number the CSMS does not know

sjh37:
A station that speaks a newer OCPP than the CSMS gets an error for every message the CSMS does not
recognise. It is supposed to get silence.

### The requirement

OCPP-J Part 4 §4.1.3, after listing the message type numbers - identical wording in OCPP 2.0.1
Edition 4 and OCPP 2.1 Edition 2:

> When a system receives a message with a Message Type Number not in this list, it SHALL ignore the
> message payload.

The 2.0.1 Edition 4 **errata 2026-06** returns to this twice:

- §4.1 *"Update handling of unknown message type"* - "The sentence about ignoring unknown message
  types has been improved slightly. It mentioned to ignore the message payload, but the intention
  was to ignore the entire message."
- §4.2 *"MessageTypeNotSupported is deprecated"* - "As of OCPP 2.1 a message type number that is not
  supported is silently ignored, as described in section 4.4. This is therefore no longer a valid
  error code."

### What the router does

```ts
default: {
  let errorCode;
  switch (protocol) { ... errorCode = ErrorCode.FormatViolation; ... }
  throw new OcppError(messageId, errorCode, 'Unknown message type id: ' + messageTypeId, {});
}
```

which the surrounding `catch` turns into a CALLERROR on the wire:

```
[4,"msg-001","FormatViolation","Unknown message type id: 6",{}]
```

`MessageTypeId` here is `Call = 2`, `CallResult = 3`, `CallError = 4`. OCPP 2.1 added two more -
`CALLRESULTERROR = 5` and `SEND = 6` - so both land in that branch.

`SEND` is the one that bites. It is how a periodic event stream delivers its data, and **N15.FR.02**
says:

> When receiving a message of RPC message type "SEND" → CSMS SHALL NOT confirm the message upon
> receipt with "CALLRESULT" or "CALLERROR".

So a 2.1 station that opens a stream is answered with an error per data message, which is both the
opposite of the requirement and noisy.

### The change

The unknown-type branch logs and returns instead of throwing. The message is still recorded to the
messages exchange sink as an unparsed inbound frame, so the audit trail does not change, and a message that is not JSON
at all still gets a CALLERROR - that is a framing failure, not an unknown type.

This does **not** implement SEND or periodic event streams. Ignoring is what the spec asks of a CSMS
that does not support a message type; implementing N11-N15 is separate work.

### Tests I replaced

`MessageRouterImpl.test.ts` had three tests asserting the CALLERROR - `should return false and send
CallError for unknown message type id` and the two per-protocol `FormatViolation` ones. They encoded
the behaviour this PR changes, so they are replaced rather than added to, and the replacements cover
`SEND`, `CALLRESULTERROR` and an arbitrary number, on all three protocols, plus the two things that
must not change.

Full workspace suite: 96 files, 2046 tests passing, 6 skipped. Docker was available for this run so the
`apps/ocpp-server/test/e2e/security-event-e2e.test.ts`, fails identically on `next`
(`Command failed: pnpm run db:migrate`) and is not this change.

### DIFF
--- a/packages/ocpp/src/modules/ocpp-router/router.ts
+++ b/packages/ocpp/src/modules/ocpp-router/router.ts
@@ -267,24 +267,24 @@ export class MessageRouterImpl extends AbstractMessageRouter implements IMessage
           break;
         }
         default: {
-          let errorCode;
-          switch (protocol) {
-            case OCPPVersion.OCPP1_6:
-            case OCPPVersion.OCPP2_0_1:
-            case OCPPVersion.OCPP2_1: {
-              errorCode = ErrorCode.FormatViolation;
-              break;
-            }
-            default: {
-              throw new Error('Unknown protocol: ' + protocol);
-            }
-          }
-          throw new OcppError(
-            messageId,
-            errorCode,
-            'Unknown message type id: ' + messageTypeId,
-            {},
+          this._logger.warn(
+            `Ignoring message with unknown message type id ${messageTypeId}`,
+            identifier,
           );
+          await this._messagesExchangeSink.record(
+            buildFrameEvent({
+              tenantId,
+              ocppConnectionName,
+              origin: MessageOrigin.ChargingStation,
+              correlationId: uuidv4(),
+              protocol,
+              raw: message,
+              timestamp: timestamp.toISOString(),
+              type: messageTypeId,
+              action: this.getActionFromIncompletelyParsedRpcMessage(rpcMessage, messageTypeId),
+            }),
+          );
+          return true;
         }
       }
     } catch (error) {
--- a/packages/ocpp/test/modules/ocpp-router/message-router-impl.test.ts
+++ b/packages/ocpp/test/modules/ocpp-router/message-router-impl.test.ts
@@ -335,35 +335,43 @@ describe('MessageRouterImpl', () => {
         expect(frames(sink, 'inbound')).toEqual([expect.objectContaining({ parsed: false })]);
       });
 
-      it('should return false and send CallError for unknown message type id', async () => {
-        const badMessage = JSON.stringify([99, CORRELATION_ID, 'SomeAction', {}]);
+      /**
+       * OCPP-J Part 4 §4.1.3, in 2.0.1 Edition 4 and 2.1 Edition 2 alike: "When a system receives a
+       * message with a Message Type Number not in this list, it SHALL ignore the message payload."
+       * The 2026-02 errata clarifies that the whole message is to be ignored, and deprecates the
+       * MessageTypeNotSupported error code because an unsupported message type number is silently
+       * ignored. 2.1 added CALLRESULTERROR (5) and SEND (6); N15.FR.02 adds that a SEND is not to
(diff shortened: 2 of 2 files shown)


sjh37:
Rebased onto current `next` at `2d127a99`, and it needed a fix to build there.

#956 (the system config refactor) deleted the `NO_ACTION` sentinel - `packages/types/src/ocpp/rpc/message.ts`
no longer exports it, and `router.ts` now derives the action with
`getActionFromIncompletelyParsedRpcMessage`, which returns `undefined` for anything that is not a
CALL. `20260828120000-repair-no-action-correlated-ocpp-messages.ts` backfills the old `'NoAction'`
rows to match. On the old base this PR passed `NO_ACTION` to `dispatchMessageReceivedUnparsed`, so
after the rebase it failed to compile:

```
src/modules/OcppRouter/src/module/router.ts(269,13): error TS2304: Cannot find name 'NO_ACTION'.
```

It now passes `undefined`, which is what `next` records for a message whose action cannot be read:

```ts
await this._webhookDispatcher.dispatchMessageReceivedUnparsed(
  tenantId,
  ocppConnectionName,
  message,
  timestamp.toISOString(),
  protocol,
  // An unknown message type number carries no action the CSMS can trust: OCPP-J only
  // places one at index 2 of a CALL, and this is not one.
  undefined,
  messageTypeId,
);
```

Still two commits, still two files, and the behaviour is unchanged - an unknown message type number
is ignored rather than answered, and the message is still recorded.

Green locally against `2d127a99`: build clean, lint clean (0 errors), and the full suite -
integration lanes included, on Docker - at **140 test files passed / 1 skipped, 2305 tests passed /
4 skipped**. `MessageRouterImpl.test.ts` on its own is 80 passed.

sjh37:
Rebased onto `486f1e99` after #990. Clean rebase - same two files, now
`packages/ocpp/src/modules/ocpp-router/router.ts` and
`packages/ocpp/test/modules/ocpp-router/message-router-impl.test.ts`. The `NO_ACTION` fix from the
previous rebase (#956 deleted the sentinel) is still in place.

Verified on `486f1e99`: build clean, full suite green (141 test files, 2305 passing / 4 skipped),
lint clean.

sjh37:
Rebased onto `c1bd4ca`, and the fix needed reworking for the router as it now stands.

`dispatchMessageReceivedUnparsed` is gone from the webhook dispatcher; unparsed inbound frames are recorded through `_messagesExchangeSink.record(buildFrameEvent(...))`. The default branch now does that instead, and takes its `action` from `getActionFromIncompletelyParsedRpcMessage` so the existing "should leave the action undefined for an unknown messageTypeId" test keeps passing. Comments removed.

Verified on `c1bd4ca`: build clean, full suite green (325 files, 4366 passing / 8 skipped), lint 0 errors.

yuvenus:
Given that there is upcoming work to properly resolve this (as in, not just ignoring the messages but actually doing something about them), closing this PR, thanks!

PR #969, opened 2026-08-27
fix(core): anchor a 1.6 composite schedule at the scheduleStart the station sent

sjh37:
A composite schedule read back from a 1.6 station is stored against the wrong point in time, so every
period in it is offset by however long the response took to arrive.

### The field

OCPP 1.6 Edition 2, `GetCompositeSchedule.conf`:

| field | card. | description |
|---|---|---|
| `status` | 1..1 | Required. |
| `connectorId` | 0..1 | Optional. The charging schedule contained in this notification applies to a Connector. |
| `scheduleStart` | 0..1 | Optional. Time. **Periods contained in the charging profile are relative to this point in time.** If status is "Rejected", this field may be absent. |
| `chargingSchedule` | 0..1 | Optional. |

`scheduleStart` is a field of the *response*, and it is the one the spec names as the anchor for the
periods. `ChargingSchedule` has its own optional `startSchedule`, but that is the schedule's start,
not the composite's.

### What the handler does

```ts
scheduleStart: response.chargingSchedule.startSchedule ?? new Date().toISOString(),
```

`response.scheduleStart` is never read. A station that answers the way the spec describes - the
composite anchored by `scheduleStart`, and no `startSchedule` inside a schedule that was computed
rather than configured - lands on the `new Date()` fallback, so the schedule is stored as starting at
the moment the CallResult happened to be processed.

§4.9 of the same document says the Charge Point calculates the intervals "from the moment the request
PDU is received: Time X, up to X + Duration", so the gap is the request/response round trip plus any
queueing on either side - small in a lab, not small on a congested link or when the CSMS is catching
up on a backlog. Every `startPeriod` offset is wrong by that amount, in the direction of the schedule
appearing to start later than it does.

### The change

Read `response.scheduleStart` first, keep `chargingSchedule.startSchedule` as the fallback for a
station that only sends that, and keep the current time as the last resort so nothing that works
today stops working.

`GetCompositeScheduleResponseOcpp16Handler` had no test file; the new one covers the three sources in
order and that a `Rejected` response still stores nothing.

Full workspace suite: 97 files, 2046 tests passing, 6 skipped. Docker was available so the

### DIFF
--- a/packages/core/src/handlers/responses/1.6/GetCompositeScheduleResponseOcpp16Handler.ts
+++ b/packages/core/src/handlers/responses/1.6/GetCompositeScheduleResponseOcpp16Handler.ts
@@ -50,7 +50,14 @@ export class GetCompositeScheduleResponseOcpp16Handler extends AbstractHandler {
       const compositeSchedule = {
         evseId: response.connectorId ?? 0,
         duration: response.chargingSchedule.duration ?? 0,
-        scheduleStart: response.chargingSchedule.startSchedule ?? new Date().toISOString(),
+        // GetCompositeSchedule.conf carries scheduleStart itself - "Periods contained in the
+        // charging profile are relative to this point in time" - so that is the anchor, and the
+        // schedule's own startSchedule is only a fallback. Defaulting to the current time would
+        // shift every period by however long the response took to arrive.
+        scheduleStart:
+          response.scheduleStart ??
+          response.chargingSchedule.startSchedule ??
+          new Date().toISOString(),
         chargingRateUnit: response.chargingSchedule
           .chargingRateUnit as unknown as ChargingRateUnitEnumType,
         chargingSchedulePeriod: response.chargingSchedule.chargingSchedulePeriod.map((p) => ({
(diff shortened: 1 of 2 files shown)


packages/core/src/handlers/responses/1.6/GetCompositeScheduleResponseOcpp16Handler.ts
```diff
@@ -50,7 +50,14 @@ export class GetCompositeScheduleResponseOcpp16Handler extends AbstractHandler {
       const compositeSchedule = {
         evseId: response.connectorId ?? 0,
         duration: response.chargingSchedule.duration ?? 0,
-        scheduleStart: response.chargingSchedule.startSchedule ?? new Date().toISOString(),
+        // GetCompositeSchedule.conf carries scheduleStart itself - "Periods contained in the
```
thanaParis: This comment explains the change in the context of this PR. it is not helpful in the code.

You have been given this feedback repeatedly by several maintainers across multiple PRs. Future PRs will be closed in response to continued violation rather than given feedback on this point. It is indicative of unreviewed AI-generated code, which we have an increasingly limited tolerance for.

thanaParis (changes requested):

sjh37:
@thanaParis Removed, and understood - no argument from me. The comment restated what the line does
and the reasoning belongs in the PR body, which is where I had already put it.

I have taken the point wider than the one line you marked: the docblock at the top of the new test
file is gone too, and the test names carry it instead:

- `anchors the periods at the scheduleStart the station reported`
- `prefers the response scheduleStart over the schedule own startSchedule`
- `falls back to the schedule own startSchedule when the response carries none`
- `stores nothing when the station rejected the request`

The diff is now the four-line change and the tests, with no prose. Four tests green.

On the wider point - it is fair, and I would rather fix the habit than argue about it. I will keep
spec citations in PR bodies and commit messages from here on, and out of the source.

thanaParis (approved):

PR #973, opened 2026-08-27
fix(types): close the OCPP 1.6 schema classes

sjh37:
Every object in the OCPP 1.6 schemas here carries `"additionalProperties": true`. The published
schemas say `false` - all 74 of them - and the OCA added that line specifically to fix this, in an
erratum that names the consequence.

### The erratum

OCPP-J 1.6 errata, **§3.2 "JSON Schema files do allow for extra fields within inner objects"**:

> It is not allowed to add extra fields/values to OCPP messages, this could cause interoperability
> issues in the field.
>
> The WSDL files are correct, the original JSON Schema files allow extra fields on inner objects and
> extra values on enums, which was not intended.
>
> Most of the JSON Schema files have been updated to fix this. The line: `"additionalProperties":
> false` has been added to the definition of all object and enum definitions.

So this is not a reading of the schemas - it is a correction the OCA made *to* the schemas, with the
rationale written down. This repo has the pre-erratum behaviour.

### The numbers

| | `additionalProperties: false` | `: true` |
|---|---|---|
| published 1.6 JSON schemas | 74 | 0 |
| this repo, before | 0 | 74 |
| this repo, after | 74 | 0 |

Counted over object types with `$ref`s resolved, across the 56 schemas that have a published
counterpart. The diff is 126 occurrences of one literal replaced by another and nothing else - more
than 74 because the enum definitions carry the keyword too, which is what the erratum means by "all
object **and enum** definitions"; on a `type: string` schema it is inert either way, and it is left
matching the published file rather than singled out.

The four security-extension schemas in the same directory - `SignedUpdateFirmware` and
`SignedFirmwareStatusNotification` - **already** say `false`, so this also makes the 1.6 set
internally consistent: at the moment four files are closed and fifty-six are open.

### What it costs today

`OcppRouter._onCall` and `_onCallResult` validate inbound messages against these schemas, so a
station's message carrying a field that is not in the definition is accepted and then dropped by the
mapper. A misspelt field reads as an absent one rather than an error.

Outbound is worse, because the operator API validates request bodies against the same schemas with
`removeAdditional: 'failing'` - Ajv only strips an additional property when it *causes* a failure,
and with `additionalProperties: true` nothing fails. A stray field in an API call therefore survives
validation, is copied into the OCPP message, and goes to a station whose own schema is closed; the
station answers a `CallError` and the command fails, with nothing on the CSMS side pointing at the
extra field. That is precisely the "interoperability issues in the field" the erratum is about.

### The risk in this change

It is a real tightening: a station that today sends a non-standard field is accepted, and after this
gets a `FormatViolation`. That is what the erratum asks for, and unlike OCPP 2.x there is no
`customData` escape hatch in 1.6, so a vendor doing this has nowhere sanctioned to go - which is why
the erratum tells them not to. Still a behaviour change worth being deliberate about, so if you would
rather stay lenient inbound and tighten only the operator API, say so and I will split it that way.

This is the 1.6 counterpart of citrineos/citrineos-core#967 (the same defect in the 2.0.1 schemas).
They touch different directories and can land in either order; the 2.1 schemas already say `false`.

Full workspace suite: 97 files, 2047 tests passing, 6 skipped. Docker was available so the

### DIFF
--- a/packages/base/test/modules/ocpp16AdditionalProperties.test.ts
+++ b/packages/base/test/modules/ocpp16AdditionalProperties.test.ts
@@ -1,0 +1,104 @@
+// SPDX-FileCopyrightText: 2026 Contributors to the CitrineOS Project
+//
+// SPDX-License-Identifier: Apache-2.0
+
+import { describe, expect, it } from 'vitest';
+import { OCPP1_6_CALL_RESULT_SCHEMA_RECORD, OCPP1_6_CALL_SCHEMA_RECORD } from '../../index.js';
+import { OCPP_CallAction, OCPPVersion } from '@citrineos/types';
+import { OCPPValidator } from '@interfaces/modules/OCPPValidator.js';
+
+/**
+ * OCPP-J 1.6 errata, §3.2 "JSON Schema files do allow for extra fields within inner objects":
+ *
+ *   "It is not allowed to add extra fields/values to OCPP messages, this could cause
+ *    interoperability issues in the field. The WSDL files are correct, the original JSON Schema
+ *    files allow extra fields on inner objects and extra values on enums, which was not intended.
+ *    Most of the JSON Schema files have been updated to fix this. The line: "additionalProperties":
+ *    false has been added to the definition of all object and enum definitions."
+ *
+ * Unlike OCPP 2.x there is no customData extension point in 1.6, so a class is simply closed.
+ */
+type Schema = Record<string, unknown>;
+
+function walk(node: unknown, path: string, visit: (node: Schema, path: string) => void): void {
+  if (Array.isArray(node)) {
+    node.forEach((child, i) => walk(child, `${path}[${i}]`, visit));
+    return;
+  }
+  if (node === null || typeof node !== 'object') {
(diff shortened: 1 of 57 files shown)


thanaParis:
I don't think this is a good idea. It is better practice to report protocol violations in logs and surface them to admins while continuing to provide the best service possible. Additional fields do not prevent the CSMS from performing its duties and should not stand in the way of functionality.

sjh37:
@thanaParis Same answer as on #967 - your call, and I am not reopening this one either.

For the record on the 1.6 side specifically: the citation here is the OCPP-J errata §3.2, where the
OCA describes the open schemas as a defect they fixed rather than a choice ("It is not allowed to add
extra fields/values to OCPP messages, this could cause interoperability issues in the field"), and
the four security-extension schemas already in this directory say `"additionalProperties": false`, so
the set is internally inconsistent. Neither of those changes the house rule you are applying, but
they are worth knowing if the question comes back.

The outbound half is #975, which does not touch what the CSMS accepts from a station.

PR #974, opened 2026-08-27
feat(core): add Drizzle DeleteCertificateAttempt repository

vicky-ahmed:


packages/core/src/dal/layers/drizzle/repository/DeleteCertificateAttempt.ts
```diff
@@ -51,5 +60,84 @@ export class DrizzleDeleteCertificateAttemptRepository extends DrizzleRepository
...
+      .limit(1);
+
+    return rows[0] ? this.toDto(rows[0]) : undefined;
+  }
+
+  async createAttempt(
+    tenantId: number,
+    input: DeleteCertificateAttemptCreate,
+  ): Promise<DeleteCertificateAttemptDto> {
+    // Replicates the Sequelize model's resolveStationId @BeforeCreate hook: resolve the
```
thanaParis: I want to get rid of these before create hooks actually. Let's take the opportunity to do that here?

thanaParis (changes requested):

packages/core/src/dal/layers/drizzle/repository/DeleteCertificateAttempt.ts
vicky-ahmed: Done

thanaParis (approved):

PR #976, opened 2026-08-28
fix: Acme CSR signing uses wrong issuer when TLS chain isn't self-issued

LoveKapibarasan:
## Summary

`Acme.getCertificateChain()` (used to sign charging station CSRs via the OCPP `SignCertificate` flow) derives the signing issuer from the second entry of `tlsCertificateChainFilePath`. This assumes the CSMS's own TLS certificate chain is always rooted in the same private sub CA that also signs charging station certs.

That assumption breaks for a hybrid PKI setup: CSMS TLS identity from a public CA (e.g. Let's Encrypt, or terminated at a reverse proxy) while a separate private sub CA signs charging station leaf certs for mutual TLS (Security Profile 3). In that case, `tlsCertificateChainFilePath[1]` is the public CA's intermediate, not the private sub CA — so the newly signed charging station certificate's issuer field doesn't actually match the key that signed it (the sub CA's private key), and the certificate is rejected as invalid by spec-compliant OCPP clients during the TLS handshake.

Found and confirmed via an end-to-end Security Profile 3 deployment against a private CA, where charging stations failed to complete the mTLS handshake with a newly OCPP-issued certificate despite `CertificateSigned` succeeding — packet capture confirmed the signed leaf's issuer DN didn't match the actual signing key.

## Fix

- Add an optional `mtlsCertificateAuthorityCertificateFilePath` field (mirroring the existing `mtlsCertificateAuthorityKeyFilePath`) so deployments can point directly at the sub CA certificate used to sign charging station CSRs, independent of whatever `tlsCertificateChainFilePath` holds for the CSMS's own TLS identity.
- `Acme.create()` resolves the sub CA cert per server: uses the dedicated field when configured, otherwise falls back to the existing `tlsCertificateChainFilePath[1]`-derived behavior — so existing self-issued-PKI deployments (where both assumptions coincide) are unaffected.
- `Acme.getCertificateChain()` now signs directly against the resolved sub CA cert instead of re-deriving it from the full chain each call.

## Test plan

- [x] `pnpm --filter @citrineos/types build` — schema changes compile
- [x] `pnpm --filter @citrineos/core build` — consumer changes compile
- [x] `vitest run test/util/certificate/client/Acme.test.ts` — updated existing coverage for the fallback path, added new coverage for the dedicated-field path
- [x] `vitest run test/modules/Certificates test/modules/OcppRouter/module/DataApi.test.ts` — no regressions in adjacent certificate/config-handling code (80 tests passing)
- [x] Verified end-to-end against a real charging station over Security Profile 3 with a hybrid PKI setup (public TLS cert for the CSMS, private sub CA for charging station certs)

Note: this field is only consumed by the static `websocketServers` config path (`Acme`). The DB-backed `ServerNetworkProfile` admin-provisioning path (`DataApi`/`installCertificateHelperService`) is a separate certificate-lifecycle feature for CitrineOS's own self-issued PKI and isn't affected by or wired to this field — happy to extend it there too if maintainers want that for parity.

### DIFF
--- a/apps/ocpp-server/migrations/20260902120000-add-mtls-ca-certificate-file-path.ts
+++ b/apps/ocpp-server/migrations/20260902120000-add-mtls-ca-certificate-file-path.ts
@@ -1,0 +1,36 @@
+// SPDX-FileCopyrightText: 2026 Contributors to the CitrineOS Project
+//
+// SPDX-License-Identifier: Apache-2.0
+'use strict';
+
+/** @type {import('sequelize-cli').Migration} */
+import { DataTypes, QueryInterface } from 'sequelize';
+
+const TABLE_NAME = 'ServerNetworkProfiles';
+const COLUMN_NAME = 'mtlsCertificateAuthorityCertificateFilePath';
+
+export default {
+  up: async (queryInterface: QueryInterface) => {
+    const table = await queryInterface.describeTable(TABLE_NAME);
+    if (table[COLUMN_NAME]) {
+      return;
+    }
+
+    // Nullable: existing servers keep deriving the CSR-signing issuer from the
+    // second entry of tlsCertificateChainFilePath, which is the behaviour they
+    // have today. Only a server whose TLS chain is not issued by its own sub CA
+    // needs to set this.
+    await queryInterface.addColumn(TABLE_NAME, COLUMN_NAME, {
+      type: DataTypes.STRING(255),
+      allowNull: true,
+    });
+  },
+
+  down: async (queryInterface: QueryInterface) => {
+    const table = await queryInterface.describeTable(TABLE_NAME);
+    if (!table[COLUMN_NAME]) {
+      return;
+    }
+    await queryInterface.removeColumn(TABLE_NAME, COLUMN_NAME);
+  },
+};
--- a/apps/operator-ui/src/lib/queries/fields/server.network.profile.fields.ts
+++ b/apps/operator-ui/src/lib/queries/fields/server.network.profile.fields.ts
@@ -36,6 +36,7 @@ export const SERVER_NETWORK_PROFILE_FULL_FIELDS = fieldSet([
   'tlsKeyFilePath',
   'tlsCertificateChainFilePath',
   'mtlsCertificateAuthorityKeyFilePath',
+  'mtlsCertificateAuthorityCertificateFilePath',
   'rootCACertificateFilePath',
   'dynamicTenantResolution',
 ]);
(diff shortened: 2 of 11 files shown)


yuvenus:
Can you retarget to `next`? Thanks!

LoveKapibarasan:
Retargeted to `next` — thanks for the pointer. It turned out to be more than a base change: `next` had moved websocket server config out of the zod `SystemConfig` and into the database, so `websocketServerInputSchema`, which this PR edited, no longer exists. The branch is rebuilt on top of current `next`.

**The fix itself is unchanged in intent.** `mtlsCertificateAuthorityCertificateFilePath` is optional; when set it is used directly as the CSR-signing issuer and the TLS chain is not read. Unset keeps the existing derivation from the chain's second entry, so nothing changes for a deployment that does not configure it.

One behavioural tidy-up came with the port: the chain is now parsed once in `create()` instead of on every CSR, and the map holds the resolved sub CA certificate rather than a chain to re-parse. `getCertificateChain` uses it directly.

**Following the new DB-backed shape**, the field is added in:

- `packages/types/src/config/types.ts` — `websocketServerSchema`
- `packages/types/src/interfaces/dto/server.network.profile.dto.ts` — DTO
- `packages/core/src/dal/layers/drizzle/{schema,repository}/ServerNetworkProfile.ts`
- `packages/core/src/dal/layers/sequelize/{model/Location,repository}/ServerNetworkProfile.ts`
- `packages/core/src/dal/interfaces/queries/Websocket.ts`
- `apps/operator-ui/src/lib/queries/fields/server.network.profile.fields.ts`
- `apps/ocpp-server/migrations/20260902120000-add-mtls-ca-certificate-file-path.ts` — nullable column, guarded with `describeTable` so it is safe to re-run

Please say if you would rather the migration were folded in elsewhere, or if the operator-ui field should stay out of this PR.

**Tests** — `Acme.test.ts`, 7 pass:

- signing against the sub CA derived from the TLS chain (the existing path; its assertion moved, since the parse now happens in `create()`)
- signing against `mtlsCertificateAuthorityCertificateFilePath` when set, asserting the TLS chain is never parsed

`pnpm -r run build` clean; `packages/core` suite 1516 passed / 4 skipped across 100 files.

One thing worth knowing for other tests: `ConfigLoader.loadWebsocketServersConfig` caches the parsed servers on a static, so a second `Acme.create` in the same file silently reuses the first one's config. The new test resets it explicitly.

yuvenus:
Closing since conflicts exist but no further updates have been made to fix them. Please feel free to reopen this once the conflicts are resolved, thanks!

PR #980, opened 2026-08-30
feat: gate Swagger UI behind an OIDC (Keycloak-compatible) login

LoveKapibarasan:
## Summary

The Swagger UI (`util.swagger.path`, `/docs`) has no authentication at all today — `registerFastifyAuth` decorates an `authorization` preHandler that is never actually attached to any route. Separately, `registerApiAuth` explicitly excludes `/docs` from `ApiAuthPlugin`'s global bearer-token check, since that mechanism can't help here anyway: it only verifies a token a caller already has and returns a bare 401 if one is missing, which gives a browser opening this page nowhere to log in.

## Fix

Adds an optional `util.swagger.oidcLogin` config block. When set, unauthenticated requests under the swagger path are redirected to the configured provider's authorization endpoint; a new `<path>/callback` route exchanges the returned code for a token and sets an httpOnly session cookie once it verifies.

Token verification reuses `OIDCAuthProvider.authenticateToken` so the JWKS fetch/cache/signature logic isn't duplicated — this is the same provider `util.authProvider.oidc` already uses for bearer-token API auth, just applied to a browser redirect flow instead. Works with Keycloak or any other standards-compliant OIDC issuer. When `oidcLogin` is left unset, behavior is unchanged (Swagger stays open, as today).

## Test plan

- [x] `pnpm --filter @citrineos/types build` — schema changes compile
- [x] `pnpm --filter @citrineos/core build` — consumer changes compile
- [x] `tsc -b apps/ocpp-server/tsconfig.json` — downstream app unaffected
- [x] `vitest run` (packages/core): 1146 passed, 19 skipped — only the two pre-existing environment-dependent suites failed (Postgres/RabbitMQ testcontainers unavailable in this sandbox), unrelated to this change
- [x] New `test/util/util/swaggerOidcLogin.test.ts` (7 tests): no-op when unconfigured, redirect with correct `authorization_endpoint`/`client_id`/`redirect_uri`, non-swagger routes untouched, callback missing-code/invalid-token/valid-token paths, valid session cookie passes through
- [x] Verified end-to-end against a real Keycloak realm on a parallel deployment (same pattern backported to our 1.9.x release line): `GET /docs/` 302s to Keycloak with the correct `redirect_uri`, callback exchanges the code and sets the cookie, and the existing OCPP HTTP API / websocket connections are unaffected

### DIFF
--- a/packages/core/src/util/util/swagger.ts
+++ b/packages/core/src/util/util/swagger.ts
@@ -14,6 +14,7 @@ import { LocalStorage } from '../files/localStorage.js';
 import type { OpenAPIV3_1 } from 'openapi-types';
 import { OpenAPIV2, OpenAPIV3 } from 'openapi-types';
 import * as packageJson from '../../../package.json' with { type: 'json' };
+import { registerSwaggerOidcLogin } from './swaggerOidcLogin.js';
 
 /**
  * This transformation is used to set default tags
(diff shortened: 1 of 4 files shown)


yuvenus:
Closing since conflicts exist and a retarget to next is required, but no further updates have been made to fix them. Please feel free to reopen this once the conflicts are resolved and it's retargeted to next, thanks!

PR #982, opened 2026-08-31
Feat: drizzle installed certificate repository added

vicky-ahmed:


packages/core/src/dal/layers/sequelize/model/Certificate/InstalledCertificate.ts
```diff
@@ -94,19 +94,6 @@ export class InstalledCertificate extends Model implements InstalledCertificateD
   @BelongsTo(() => Tenant, 'tenantId')
   declare tenant?: TenantDto;
 
-  @BeforeCreate
```
yuvenus, line 97: Was this removed because it's not needed anymore?

vicky-ahmed: Yes, @thanaParis mentioned that these hooks were no longer needed.

vicky-ahmed: Because Drizzle doesn't support Sequelize's @BeforeCreate hooks, so to keep both paths behaving the same I moved the logic into the repo so both Sequelize and Drizzle can use it. Same behavior as before.

thanaParis (approved):

PR #985, opened 2026-08-31
Adding DB_STRATEGY: none back

yuvenus:
This is a small PR to add back in the "none" option for DB_STRATEGY, so you can opt out of running DB migrations. 😄

### DIFF
--- a/apps/ocpp-server/deploy.Dockerfile
+++ b/apps/ocpp-server/deploy.Dockerfile
@@ -2,9 +2,6 @@
 #
 #  SPDX-License-Identifier: Apache-2.0
 
-# Set DB_STRATEGY, options are migrate or none
-ARG DB_STRATEGY=none
-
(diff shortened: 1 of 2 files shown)


lydiazcheng (approved):
Shall we add `sync` and `force-sync` as well by the way?

thanaParis (approved):

yuvenus:
> Shall we add `sync` and `force-sync` as well by the way?

@lydiazcheng I thought about adding back but decided against it, maybe in another PR! Not sure if anyone even uses sync/force-sync. 😆

lydiazcheng:
> > Shall we add `sync` and `force-sync` as well by the way?
> 
> @lydiazcheng I thought about adding back but decided against it, maybe in another PR! Not sure if anyone even uses sync/force-sync. 😆

Gotcha. Thank you, Venus!

PR #987, opened 2026-09-01
Feature: Implement Drizzle ChargingStation

lydiazcheng:
## Changes
1. Implement methods for drizzle charging station repository
2. Split `addStatusNotificationToChargingStation` into `IStatusNotificationRepository`.
3. Add DI for charging station and rename `locationRepository` to `chargingStationRepository`
4. Change `DrizzleRepository.tenantFilter` to `protected` so other repositories can reuse it.
5. Rename `readChargingStationByStationId` and `doesChargingStationExistByStationId`
## Tests
1. Build docker with CITRINEOS_USE_DRIZZLE: 'true' and run it locally without error
6. Manage to handle the 1.6 BootNotification Screenshot 2026-09-01 at 12 11 10 PM
7. `POST /setStationPassword` get 200 response, i.e.,`readChargingStationByStationId` and tenant filter work.  Screenshot 2026-09-01 at 12 34 03 PM
8. Manage to handle 1.6 StatusNotification without errors Screenshot 2026-09-01 at 12 39 08 PM


packages/core/src/dal/layers/drizzle/repository/ChargingStation.ts
```diff
@@ -66,5 +105,164 @@ export class DrizzleChargingStationRepository extends DrizzleRepository<
...
+      .where(
+        and(eq(table.ocppConnectionName, ocppConnectionName), this.tenantFilter(table, tenantId)),
+      )
+      .limit(1)) as ChargingStationEntity[];
+    return rows[0];
+  }
+
+  // ─── IChargingStationRepository methods ──────────────────────────────────
+
+  async readChargingStationByStationId(
```
devonNewsomDevelop: This function name isn't accurate since we are using the ocppConnectionName and the tenantId to find the charging station, and not the charging station's PKID

packages/core/src/dal/layers/drizzle/repository/ChargingStation.ts
```diff
@@ -66,5 +105,164 @@ export class DrizzleChargingStationRepository extends DrizzleRepository<
...
+        );
+        return undefined;
+      }
+      return await this.insert(tenantId, { ocppConnectionName, ...values });
+    }
+
+    return await this.updateById(tenantId, existing.id, { ...values, updatedAt: new Date() });
+  }
+
+  async doesChargingStationExistByStationId(
```
devonNewsomDevelop: same thing with this method name, it implies we'll be using the pkid (to me anyways) but we're using a composite key that doesnt include the pkid

devonNewsomDevelop (changes requested):
smol changes requested, which I think might have been issues before this PR, but it would be good to tweak them now if we can 😊

packages/core/src/dal/layers/drizzle/repository/ChargingStation.ts
lydiazcheng: Thank you, Devon! Both renamed.

devonNewsomDevelop (approved):

PR #990, opened 2026-09-01
Feature/ocpp restructure

devonNewsomDevelop:
Moves dal to it's own package @dal
Restructures core and renames package to @ocpp

Highly recc. running 

pnpm fresh:install when pulling


.reorg/map.json
```diff
@@ -0,0 +1,89 @@
+{
```
yuvenus: Is this just a reference file? If so, is it okay to remove?

yuvenus:
I think I know the answer, but I noticed some files still were left un-kebab-ified, primarily in `types/src/ocpp/model`, probably because they're generated right? It seems fair to keep em as is just to not have to mess with the file generator too much to be honest though. 🤔

devonNewsomDevelop:
> I think I know the answer, but I noticed some files still were left un-kebab-ified, primarily in `types/src/ocpp/model`, probably because they're generated right? It seems fair to keep em as is just to not have to mess with the file generator too much to be honest though. 🤔

@yuvenus  Yerp, that's right. Just seemed irrelevant anyway, no one really ever views those files lol. I also avoided renaming .jpgs, .pems, .license files, .mds, etc

yuvenus (approved):

PR #991, opened 2026-09-01
Feat: drizzle reservation repository

vicky-ahmed:


packages/core/src/dal/layers/drizzle/repository/Reservation.ts
```diff
@@ -48,5 +51,161 @@ export class DrizzleReservationRepository extends DrizzleRepository<
     return toReservationDto(row);
   }
 
-  // Domain query/write methods intentionally omitted — stub outline only.
+  async createOrUpdateReservation(
+    tenantId: number,
+    reserveNowRequest: OCPP2_0_1.ReserveNowRequest,
```
devonNewsomDevelop: Should this perhaps be a common types import? Or do we need it specific to 2.0.1? We have some union types for 2.0.1 and 2.1 and im wondering if that would fit here?

devonNewsomDevelop (approved):
Approved with suggestion

packages/core/src/dal/layers/drizzle/repository/Reservation.ts
vicky-ahmed: Good catch. you are right, the union type is the correct fit here. Updating..

PR #992, opened 2026-09-01
Fix: OCPP 1.6 Charging Station Status Mapping in operator-ui 

lydiazcheng:
### Changes
1. Add a new `ChargingStationStatusEnum`
2. Extend mapping for 1.6 Status and add handling for `RESERVED`
3. Replace protocol specific enum usages with citrine general enum

### Tests
1. Charging with 1.6 Evereset Screenshot 2026-09-01 at 4 19 48 PM
2. Charing with 2.0.1 (tested using wscat) Screenshot 2026-09-01 at 4 32 03 PM


devonNewsomDevelop (approved):
This is much easier to read than before. Good work!

thanaParis (approved):
ideal! this is the preferred solution; I didn't realize we already had the enum and mappers available.

PR #997, opened 2026-09-02
feat(ocpp): acknowledge PublishFirmwareStatusNotification

sjh37:
## What

A `PublishFirmwareStatusNotificationRequest` from a Local Controller was answered `CallError NotSupported`. It now gets the empty `PublishFirmwareStatusNotificationResponse` the controller is waiting for, on 2.0.1 and 2.1.

## Why

The CSMS *asks* for these reports. `PublishFirmware` is an operator endpoint (`endpoints/ocpp/2/configuration.ts:22`) and `PublishFirmwareResponseOcpp2Handler` handles its answer, so a Local Controller that accepts the request will report back per **L03.FR.01**:

> Whenever the Local Controller enters a new status in the publishing process, it SHALL send a PublishFirmwareStatusNotificationRequest message to the CSMS.

Everything but the handler was there: `OCPP_CallAction.PublishFirmwareStatusNotification`, both schemas in `MappingSchema.ts`, the action in the 2.0.1 and 2.1 action sets. So the router validated and routed it, the Configuration module found no handler, and `AbstractModule.ts:293` answered `NotSupported` - to every `Downloading`, `Downloaded`, `Published`, `DownloadFailed`, `InvalidChecksum` and `PublishFailed` report.

This is the same gap as #890 (FirmwareStatusNotification on 1.6), one protocol up.

## Change

- `PublishFirmwareStatusNotificationRequestOcpp2Handler`, registered in the Configuration module next to `FirmwareStatusNotificationRequestOcpp2Handler`. It acknowledges with `{}` and, on `Published`, logs the URIs the controller listed (L03.FR.04) - a station's `UpdateFirmwareRequest` is what those are for.
- `@citrineos/types`: the 2.x request/response unions gain `PublishFirmwareStatusNotificationRequest` / `Response`. They had `PublishFirmware` but not its status message.

## Tests

`packages/ocpp/test/handlers/requests/2/publish-firmware-status-notification-request-ocpp-2-handler.test.ts`: a handler is declared for 2.0.1 and for 2.1 (the `handlersDeclaring` check from the #890 test, so a handler dropped from the barrel fails here); each failure status is acknowledged with `{}`; a `Published` report with two locations is acknowledged; a report with no `requestId` is acknowledged. 8 tests, green; `packages/types` and `packages/ocpp` build and lint clean.


sjh37:
Rebased onto `486f1e99` after #990 and moved into the `ocpp` package.

- handler: `packages/ocpp/src/handlers/requests/2/publish-firmware-status-notification-request-ocpp-2-handler.ts`
- exported from `packages/ocpp/src/handlers/index.ts` and registered in
  `packages/ocpp/src/modules/configuration/register.ts`
- test alongside the handler, same kebab name
- the two union types in `packages/types/src/ocpp/rpc/2/` are unaffected by the refactor

Verified on `486f1e99`: build clean, full suite green (142 test files, 2309 passing / 4 skipped),
lint clean.

lydiazcheng (approved):

PR #999, opened 2026-09-02
fix(dal): scope createChargingNeeds and getTransactionsCount to the tenant

sjh37:
## What

Two Sequelize reads bypassed the tenant-scoped base repository and left `tenantId` out of their `where`. Both now carry it.

| Method | Was | Effect |
|---|---|---|
| `SequelizeChargingProfileRepository.createChargingNeeds` (`ChargingProfile.ts:179`) | `Transaction.findOne({ where: { ocppConnectionName, isActive: true }, include: [Evse where evseTypeId] })` | Picks up a same-named station's active transaction **in another tenant** |
| `SequelizeTransactionEventRepository.getTransactionsCount` (`TransactionEvent.ts:520`) | `Transaction.count({ where: {} + date window })` | Counts every tenant's transactions; `getTransactions` beside it is tenant-scoped |

## Why it is reachable

A station name is unique **per tenant**, not globally - the constraint is `ChargingStations_stationName_tenantId_key ("ocppConnectionName", "tenantId")`. So two tenants can each have a `CS-1`, and every lookup by station name that omits the tenant can cross between them. That is the reason #884, #900, #911 and #916 exist; these two are the same class, found by sweeping the repository layer for direct `Model.findOne/count` calls whose `where` has no `tenantId`. These were the only two.

For `createChargingNeeds` the consequence is not just a wrong row: the `ChargingNeeds` written against the other tenant's transaction is what `calculateChargingProfile` reads, so the `SetChargingProfile` the CSMS then sends the station was computed from a stranger's requested energy.

## Tests

`packages/dal/test/repositories/sequelize/shared-station-name-tenant-scope-integration.test.ts` (testcontainers): two tenants, each with a station called `CS-SHARED` and one EVSE.

- With an active transaction only in the other tenant, `createChargingNeeds` for the caller's tenant now throws `No active transaction found`; before the fix it **resolved** with a `ChargingNeeds` row pointing at the other tenant's transaction.
- With a transaction in both, the needs attach to the caller's own transaction and EVSE (before: `expected 1 to be 2`).
- `getTransactionsCount` returns 1 with one own and two foreign transactions (before: 3), and equals the length of the page `getTransactions` returns.

Run without the fix, all four fail as described; with it, all four pass. `packages/dal` builds and lints clean. A new test file rather than an addition to `charging-needs-lookup-integration.test.ts`, which #960 is editing.

### DIFF
--- a/packages/dal/src/repositories/sequelize/charging-profile.ts
+++ b/packages/dal/src/repositories/sequelize/charging-profile.ts
@@ -182,6 +182,7 @@ export class SequelizeChargingProfileRepository
   ): Promise<ChargingNeeds> {
     const activeTransaction = await Transaction.findOne({
       where: {
+        tenantId,
         ocppConnectionName: ocppConnectionName,
         isActive: true,
       },
--- a/packages/dal/src/repositories/sequelize/transaction-event.ts
+++ b/packages/dal/src/repositories/sequelize/transaction-event.ts
@@ -511,7 +511,7 @@ export class SequelizeTransactionEventRepository
 
   async getTransactionsCount(tenantId: number, dateFrom?: Date, dateTo?: Date): Promise<number> {
     const queryOptions: WhereOptions<any> = {
-      where: {},
+      where: { tenantId },
     };
 
     if (dateFrom) {
--- a/packages/dal/test/repositories/sequelize/shared-station-name-tenant-scope-integration.test.ts
+++ b/packages/dal/test/repositories/sequelize/shared-station-name-tenant-scope-integration.test.ts
@@ -1,0 +1,190 @@
+// SPDX-FileCopyrightText: 2026 Contributors to the CitrineOS Project
+//
+// SPDX-License-Identifier: Apache-2.0
+
+import { afterAll, beforeAll, beforeEach, describe, expect, it } from 'vitest';
+import { GenericContainer, type StartedTestContainer, Wait } from 'testcontainers';
+import type { Sequelize } from 'sequelize-typescript';
+import { DEFAULT_TENANT_ID } from '@citrineos/base';
+import { OCPP2_0_1, type SystemConfig } from '@citrineos/types';
+import {
+  ChargingStation,
+  DefaultSequelizeInstance,
+  Evse,
+  EvseType,
+  SequelizeChargingProfileRepository,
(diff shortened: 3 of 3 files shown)


lydiazcheng (approved):

sjh37:
Rebased onto `486f1e99` after #990, and moved into the new `dal` package.

- the two one-line fixes are now in `packages/dal/src/repositories/sequelize/charging-profile.ts`
  and `.../transaction-event.ts`
- the new integration test is
  `packages/dal/test/repositories/sequelize/shared-station-name-tenant-scope-integration.test.ts` -
  kebab-cased, and its import changed from `@dal/index.js` (which no longer resolves - there is no
  `packages/dal/src/index.ts`) to `../../../index.js`, matching the sibling tenant-scoping suites

Verified on `486f1e99`: build clean, full suite green (142 test files, 2305 passing / 4 skipped),
lint clean. Behaviour unchanged - it is still the same two `tenantId` additions.

packages/dal/src/repositories/sequelize/charging-profile.ts
```diff
@@ -180,8 +180,10 @@ export class SequelizeChargingProfileRepository
     chargingNeedsReq: OCPP2_0_1.NotifyEVChargingNeedsRequest,
     ocppConnectionName: string,
   ): Promise<ChargingNeeds> {
+    // A station name is unique per tenant, not across them (ChargingStations_stationName_tenantId).
```
thanaParis: remove.

thanaParis (changes requested):

packages/dal/src/repositories/sequelize/charging-profile.ts
sjh37: Removed, and rebased onto current `next`. The PR is now the two `tenantId` additions and the integration test.

sjh37:
Rebased onto `c1bd4ca`. The comment is removed, and the integration test's config type changed from `BootstrapConfig` (no longer exported from `@citrineos/base`) to `SystemConfig`, matching its siblings in `packages/dal/test/repositories/sequelize`.

Verified on `c1bd4ca`: build clean, full suite green (327 files, 4370 passing / 8 skipped), lint 0 errors.

thanaParis (approved):

PR #1001, opened 2026-09-02
feat(core): make the AMQP queue type configurable, allowing quorum queues

LoveKapibarasan:
## Problem

`RabbitMqReceiver` declares every queue with `autoDelete: true`. A quorum queue **cannot** be auto-delete, and RabbitMQ does not reject such a declaration — it silently returns a **classic** queue instead.

The consequence is that a broker configured for quorum queues keeps producing classic ones, with nothing on the client side to indicate the setting had no effect:

```
# node config and vhost metadata both set
default_queue_type = quorum
rabbitmqctl list_vhosts name default_queue_type   ->  /  quorum

# every queue still comes back classic
rabbitmqctl list_queues name type durable auto_delete
rabbit_queue_evdriver_requests   classic   true   true
```

We deleted the queues and let CitrineOS re-declare them several times before working out that `autoDelete` was the reason.

## Why it matters

Classic queues combined with an `ha-mode` mirroring policy are Mnesia-replicated, and Mnesia does not heal a network partition by itself.

On a three-node 3.13 cluster we operate, a partition between one node and the other two left that node's queues wedged. `QueueDeclare` started returning `404 NOT_FOUND ... due to timeout`, so CitrineOS could not finish booting and was restarted by Docker **122 times** until the queues were deleted manually. Restarting the broker did not clear it.

A quorum queue in the same cluster, belonging to another service, rode the same partition out with a Raft leader election and needed no intervention. Classic queues are also removed entirely in RabbitMQ 4.x, so the migration is one-way regardless.

## Change

Adds `messageBroker.amqp.queueType`, `classic` (default) or `quorum`. The default preserves today's behaviour exactly, so existing deployments are unaffected.

When set to `quorum`:

- queues are declared with `arguments: { 'x-queue-type': 'quorum' }` and **without** `autoDelete`
- per-identifier queues are deleted explicitly in `unsubscribe()`, which is the cleanup `autoDelete` used to provide. `ifUnused: true` leaves alone any queue another consumer still holds, and a queue that is already gone is logged rather than failing the unsubscribe

All three `assertQueue` call sites go through one `_queueOptions()` helper so the two modes cannot drift.

## Tests

`packages/core/test/util/queue/rabbit-mq/receiver.test.ts` — 30 pass, 3 new:

- a quorum declaration carries `x-queue-type` and omits `autoDelete`
- `unsubscribe()` deletes the queue with `ifUnused` when the type is quorum
- `unsubscribe()` leaves removal to the broker when the type is classic

### DIFF
--- a/packages/core/src/util/queue/rabbit-mq/receiver.ts
+++ b/packages/core/src/util/queue/rabbit-mq/receiver.ts
@@ -31,6 +31,28 @@ import { RabbitMQChannelManager } from './ChannelManager.js';
  */
 export class RabbitMqReceiver extends AbstractMessageHandler {
   protected static readonly QUEUE_PREFIX = 'rabbit_queue_';
+
+  protected readonly _queueType: 'classic' | 'quorum';
+
+  /**
+   * Declaration options for every queue this receiver asserts.
+   *
+   * A quorum queue cannot be auto-delete. When one is requested anyway the
+   * broker does not fail the declaration -- it quietly gives back a classic
+   * queue, so a `default_queue_type = quorum` vhost silently keeps producing
+   * classic queues. Asking for the type explicitly, and dropping autoDelete
+   * with it, is what makes the setting observable.
+   */
+  protected _queueOptions(): amqplib.Options.AssertQueue {
+    if (this._queueType === 'quorum') {
+      return {
+        durable: true,
+        exclusive: false,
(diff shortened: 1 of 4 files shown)


yuvenus:
Closing because conflicts have gone unresolved. Feel free to reopen with the necessary changes to SystemConfig, thanks!

PR #1002, opened 2026-09-02
Feature: Implement Drizzle Evse

lydiazcheng:
### Changes
1. Implement drizzle transaction to commit and rollback db queries
2. Implement drizzle db tranactions for evse which requires in `createOrUpdateEvse` 
3. Implement 3 `DrizzleEvseRepository` methods
4. Removed the Evse BeforeCreate hook and create `resolveStationId()` helper for reusing for other models
### Tests
1. Build docker with `CITRINEOS_USE_DRIZZLE` to true and run it without errors
5. Run a charging and a new evse should be created in the charging station with stationId populated Screenshot 2026-09-02 at 2 07 39 PM


packages/dal/src/repositories/drizzle/base.ts
```diff
@@ -66,6 +83,31 @@ export abstract class DrizzleRepository<TTable extends CitrineTable, TDto> exten
...
+   * rolling back when it throws. Pass the supplied context to the shared write
+   * helpers so their statements join the transaction.
+   */
+  protected async withAtomicWrite<T>(fn: (ctx: DrizzleWriteContext) => Promise<T>): Promise<T> {
+    const events: DrizzleWriteContext['events'] = [];
+
+    const result = await this.db.transaction(async (tx) => fn({ db: tx, events }));
+
+    for (const event of events) {
+      this.emit(event.name, event.payload);
```
thanaParis: why are we doing this? it doesn't look like the drizzle repo is an event emitter--this is dead code, no?

thanaParis (changes requested):

packages/dal/src/repositories/drizzle/base.ts
lydiazcheng: Hi Thana, thanks for the review!

They are not dead code. The db queries are committed atomically by line `const result = await this.db.transaction(async (tx) => fn({ db: tx, events }));`. The `for (const event of events) {` and the following line are to make sure the events are only emitted after all db queries have committed successfully. If the transaction fails, it rolls back and throws before these event emitter lines.  

We originally need this `withAtomicWrite` for `Evse` methods like `createOrUpdateEvse` and `autoCommissionEvseForOcpp16Connector` since they run inside db transaction and need to emit multiple events. After syncing with changes in `next`, `autoCommissionEvseForOcpp16Connector` no longer need it, but since device model also has cases like this, I think it is still valid to keep this `withAtomicWrite`.

thanaParis (approved):

PR #1006, opened 2026-09-02
Messages Module

yuvenus:
# Naming?

Not married to the name "Messages" for the module (or any of the other classes for that matter), will be happy to use any name suggestions instead. 😄 

# Why?

The Messages module is meant to offload some of the "business" logic that goes into processing OCPP messages away from the OcppRouter so that it can focus on what it does best (forwarding messages to and fro CitrineOS/the Charging Station) while the Messages module handles things like storing OCPP messages in the database, dispatching events to subscriptions, and handling connections/disconnections.

As of this PR, those concerns have been categorized into two types: "frame" events and websocket "connections". This becomes relevant for the queuing and processing logic down below.

# How?

The goal of the Messages module is to separate these concerns in the following ways:

1. Create a new RabbitMQ exchange solely for messages (currently called "messages")
  a. This exchange holds two separate queues: one for "frame" events and one for websocket "connections"
2. Move all "business" logic into its own module Messages
  a. "Business" logic is then represented by "processors" that handle specific parts of the frame/connection processing to keep a separation of concerns.
3. Move the WebhookDispatcher out of OcppRouter and into the new Messages module
4. Update the OcppRouter to forward messages to the Messages module via the MessagesExchangeSink, which sends the messages/connection events off to the new "messages" exchange.

## Event Types

For now, the messages module expects two different event types: "frame" events and "connection" events. Each of these event types gets their own queue in the messages exchange so they're processed separately. The messages consumer checks which event type the incoming message is before sending it off to be processed by whichever processes are registered in the messages module for that particular event type.

## Processors
The processors found in `messages/processors` are what actually process the messages coming in from the queues. As of this PR, OCPP frame events have two processors: one to persist into the OCPPMessages DB table, and the other to forward the event to any subscriptions exist in subscriptions repository. Connection events have one processor: to use the webhook dispatcher to register or deregister the connection.

Processors can be considered either critical or not critical. Critical processors will abort the processing if it fails, otherwise non-critical processors will continue processing. This is by design to ensure that processing can still occur as needed.

## Dead Letter Queue

For now, a dead letter queue exists for both queues, but they only result in a log (more processing will come in the future, not in this PR).


thanaParis (approved):
Nice work!

PR #1008, opened 2026-09-03
feat: configurable SignedMeterValue maxLength for OCPP 2.0.1

LoveKapibarasan:
## Summary

OCPP 2.0.1's own JSON schema caps `SignedMeterValueType.signedMeterData` and `.publicKey` at 2500 characters each. This is the OCA's own spec value (both schema files carry `"comment": "OCPP 2.0.1 FINAL"` and are otherwise unmodified copies of the official generated schemas) -- not something CitrineOS introduced or tightened.

In practice, this limit is proven too small for real deployments: **we hit this on a real DC fast charger** (a Reinhausen DCC-150) sending OCMF-signed meter data -- with an embedded certificate chain and multiple register readings -- that runs to ~2772 characters, comfortably over the 2500 cap. CitrineOS rejects the whole `TransactionEvent`/`MeterValues` message outright as a `FormatViolation`, which silently drops that station's meter values for the rest of the transaction (no partial acceptance -- the entire message fails ajv validation).

Notably, **OCPP 2.1 already raised `signedMeterData`'s limit to 32768** for what looks like exactly this reason (`publicKey` stays at 2500 in both versions), so this isn't a hypothetical edge case -- the OCA itself has acknowledged 2.0.1's limit is too tight for real-world signed meter data.

Since deployments can't unilaterally speak OCPP 2.1 just to work around this, and today there's no way to adjust the 2.0.1 limit without hand-patching the compiled schema JSON (not survivable across a rebuild, not a real fix), this PR adds a proper config-driven override.

## Changes

* `modules.transactions.signedMeterValuesConfiguration` gains two new optional fields: `signedMeterDataMaxLength` and `publicKeyMaxLength`. Unset keeps the OCPP 2.0.1 spec default (2500) unchanged.
* `OCPPValidator` gains a static `applySignedMeterValueLimits()` that patches the shared `TransactionEventRequest`/`MeterValuesRequest` schemas' `SignedMeterValueType` definitions in place, before ajv compiles them. Also callable via an optional third constructor parameter.
* `CitrineOSServer` threads the new config fields into the single shared `OCPPValidator` instance it constructs at startup.

The override is intentionally narrow (just these two known pain points) rather than a generic schema-patching mechanism, to keep the change small and the behavior predictable.

## Test plan

- [x] `pnpm --filter @citrineos/types build` / `@citrineos/base build` / `@citrineos/core build` / `ocpp-server build` -- all clean
- [x] `vitest run test/modules/OCPPValidator.test.ts` -- 56/56 passing, including 6 new tests covering: default-limit rejection, widening `signedMeterDataMaxLength`, widening `publicKeyMaxLength`, that the override applies to `MeterValues` as well as `TransactionEvent`, a no-op call with no overrides, and applying limits via the constructor
- [x] `eslint` clean on all touched files
- [x] Verified end-to-end against the real charging station described above (after hand-patching the equivalent field on an older deployment) that widening the limit is what actually fixes the dropped-meter-values symptom

### DIFF
--- a/apps/ocpp-server/src/citrineOSServer.ts
+++ b/apps/ocpp-server/src/citrineOSServer.ts
@@ -165,7 +165,11 @@ export class CitrineOSServer {
     // Create a separate OCPPValidator with its own Ajv instance for OCPP message validation.
     // This must be distinct from _ajv: OCPP messages are parsed JSON (no coercion needed),
     // whereas _ajv coerces types for Fastify's HTTP schema compilation.
-    this._ocppValidator = new OCPPValidator(this._logger);
+    const signedMeterValuesConfiguration = this._config.modules.transactions?.signedMeterValuesConfiguration;
+    this._ocppValidator = new OCPPValidator(this._logger, undefined, {
+      signedMeterDataMaxLength: signedMeterValuesConfiguration?.signedMeterDataMaxLength,
+      publicKeyMaxLength: signedMeterValuesConfiguration?.publicKeyMaxLength,
+    });
 
     // Set cache implementation
     this._cache = this.initCache(cache);
--- a/packages/base/src/interfaces/modules/OCPPValidator.ts
+++ b/packages/base/src/interfaces/modules/OCPPValidator.ts
@@ -22,15 +22,76 @@ import {
 } from '@citrineos/types';
 import { OcppError } from '@ocpp/rpc/message.js';
 
+/**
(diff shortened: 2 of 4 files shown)


thanaParis:
I would recommend replacing the json schema files you need updated. Validation adjustments like this are supported by being able to change those json schemas yourself--we have taken this approach in the field with networks that required it. If we added configuration for every charger that broke the spec, we would end up with endless configuration options.

PR #1012, opened 2026-09-03
Feature/fixing unit tests

thanaParis:
tsconfig.json files at project roots have been altered to incorporate test folders; additional tsconfig.src.json files have been added to handle src specific tsconfig settings

existing type issues with tests have been fixed

vitest now always runs typecheck

pnpm scripts updated for tsconfig changes; pnpm clean commands now clean vitest & pnpm caches to ensure they actually clean local build results


yuvenus (approved):
No more red squiggles!! 😍 

Very minor thing: I know this PR specifically set out to fix the types (and to some degree, the aliases) for the tests. But there are some folders unaccounted for in the regular aliases over in `tsconfig.src.json`s - for example, `packages/ocpp/src/services` doesn't have an `@services` alias but `packages/ocpp/src/modules` does with `@modules`. Would we consider adding those aliases in this PR too (since we're kinda already in the weeds of it here) or leave that for another day? 👀

PR #1013, opened 2026-09-03
Separate the OCPP 1.6 and 2.x connector commissioning paths

thanaParis:
A Connector is identified two different ways depending on the protocol: 1.6 numbers connectors station-wide (connectorId), 2.0.1 numbers them per EVSE (evseTypeConnectorId). Both columns were NOT NULL, so each path had to invent the other protocol's number. The 2.x path reused the per-EVSE number as the station-wide one, which collides with a real connector on a multi-EVSE station; the 1.6 path commissioned an EvseType keyed off the 1.6 connectorId.

Both columns are now nullable and each path fills only its own:

- createOrUpdateConnector splits into createOrUpdateOcpp16Connector (keyed on ocppConnectionName + connectorId) and createOrUpdateOcpp2Connector (keyed on evseId + evseTypeConnectorId), so a null in the other protocol's column never makes the upsert insert a duplicate row.
- commissionEvseForOcpp16Connector becomes autocommissionEvseForOcpp16Connector and creates the Evse only; the EvseType it used to derive from the 1.6 connectorId is no longer needed.
- The 2.0.1 path upserts the Connector before writing the StatusNotification that references it.
- GetTariffs reads its default tariffs through the new ILocationRepository.readConnectorsWithTariffsByStationId instead of querying the Connector model directly from the handler.

### DIFF
--- a/packages/dal/src/models/location/connector.ts
+++ b/packages/dal/src/models/location/connector.ts
@@ -1,6 +1,7 @@
 // SPDX-FileCopyrightText: 2025 Contributors to the CitrineOS Project
 //
 // SPDX-License-Identifier: Apache-2.0
+import { DEFAULT_TENANT_ID, OCPP1_6_Namespace } from '@citrineos/base';
 import type {
   ChargingStationDto,
   ConnectorDto,
@@ -18,7 +19,6 @@ import type {
   TenantDto,
   TransactionDto,
 } from '@citrineos/types';
-import { DEFAULT_TENANT_ID, OCPP1_6_Namespace } from '@citrineos/base';
 import {
   BeforeCreate,
   BeforeUpdate,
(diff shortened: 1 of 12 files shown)


packages/ocpp/src/modules/transactions/status-notification-service.ts
```diff
@@ -75,18 +75,13 @@ export class StatusNotificationService {
...
 
     let matchingEvse = chargingStation.evses?.find(
       (evse) => evse.evseTypeId === statusNotificationRequest.evseId,
     );
-    let matchingConnector: ConnectorDto | undefined = (
-      matchingEvse?.connectors as Connector[] | undefined
-    )?.find((c) => c.evseTypeConnectorId === statusNotificationRequest.connectorId);
+    // let matchingConnector: ConnectorDto | undefined = (
+    //   matchingEvse?.connectors as Connector[] | undefined
+    // )?.find((c) => c.evseTypeConnectorId === statusNotificationRequest.connectorId);
```
lydiazcheng: Tiny: can be removed.

packages/dal/src/repositories/sequelize/location.ts
```diff
@@ -380,36 +400,23 @@ export class SequelizeLocationRepository
...
     tenantId: number,
-    value: Partial<Connector>,
+    value: Partial<ConnectorDto>,
     query: object,
   ): Promise<Connector[]> {
     return await this.connector.updateAllByQuery(tenantId, value, query);
   }
 
-  async commissionEvseForOcpp16Connector(
+  async autocommissionEvseForOcpp16Connector(
```
lydiazcheng: Tiny:
```suggestion
  async autoCommissionEvseForOcpp16Connector(
```

packages/dal/src/models/location/connector.ts
```diff
@@ -1,6 +1,7 @@
 // SPDX-FileCopyrightText: 2025 Contributors to the CitrineOS Project
```
lydiazcheng, line 1: We need a migration file for changes in this file?

thanaParis: no, the migrations never enforced the foreign key or nullability.

lydiazcheng (approved):

PR #1014, opened 2026-09-04
feat: drizzle message info repository added

vicky-ahmed:


packages/dal/src/repositories/drizzle/message-info.ts
```diff
@@ -47,5 +49,66 @@ export class DrizzleMessageInfoRepository extends DrizzleRepository<
...
+    }
+    await this.db
+      .update(table)
+      .set({ active: false })
+      .where(and(...conditions));
+  }
+
+  async createOrUpdateByMessageInfoTypeAndStationId(
+    tenantId: number,
+    message: OCPP2_0_1.MessageInfoType,
```
lydiazcheng: Since this method is used in the handler for both 201 and 21, how about updating it to `OCPP2_common_types.MessageInfoType` which has been used in packages/dal/src/repositories/repositories.ts? The same change is expected for the method in packages/dal/src/repositories/sequelize/message-info.ts.

vicky-ahmed: Good catch, Updating..

lydiazcheng (approved):

thanaParis (approved):

PR #1020, opened 2026-09-08
feature(ocpp-server) adds CaliforniaPricing for OCPP 1.6

andreosasLarsson:
# Adds California pricing for OCPP 1.6

adds parts of California pricing v3.1 to OCPP 1.6, due to existing limits with CitrineOS some features are left out.
some optional features are also left out, since this commit is large enough as is 

the feature can be controlled using CITRINEOS_CALIFORNIAPRICING_ENABLED, defaults to false, since this feature adds additional overhead

Added:
 * defaultPrice: priceText and chargingPrice
 * runningCost:  cost, state, chargingPrice
 * finalCost: Cost, priceText
 * CustomDisplayCostAndPrice

 what has been left out:
  * runningCost.priceTextOffline: left out as it is optional
  * SetUserPrice: left out as Citrine does not take "Authorization" Tariff into account when calculating cost (handler scaffolding is included but deliberately not registered)
  * finalCost.qrCodeText:  not configurable right now, so no where to point to
  * idlePrice: optional, left out as no way of configuring in Citrine
  * runningCost.nextPeriod: optional, left out as no way of configuring in Citrine
  * multiLanguage support: optional, adds more complexity but nothing preventing the feature to be added in the future from what i can see
  * timezones: optional, should be possible to add in future
  * ConnectorUnplugged: optional, should be possible to add in future

to ensure we also keep the price in the backend OCPP 1.6 now sets totalCost and timeSpentCharging to transactions

Tests have been added for CaliforniaPricing features

Note: timeSpentCharging is derived from total session duration, as OCPP 1.6 does
not define a separate charging-time measurement.

Also adds a location lookup to the DAL repositories to resolve station tariffs.

### DIFF
--- a/packages/ocpp/src/handlers/requests/1.6/stop-transaction-request-ocpp-16-handler.ts
+++ b/packages/ocpp/src/handlers/requests/1.6/stop-transaction-request-ocpp-16-handler.ts
@@ -23,28 +23,33 @@ import {
   Transaction,
 } from '@citrineos/dal';
 import { OCPP1_6_Mapper } from '@citrineos/dal';
+import type { CostCalculator } from '@modules/transactions/cost-calculator.js';
 
 @AsRequestHandler([OCPPVersion.OCPP1_6], OCPP_CallAction.StopTransaction)
 export class StopTransactionRequestOcpp16Handler extends AbstractHandler {
   protected _ocppSender: IOcppSender;
   protected _authorizationRepository: IAuthorizationRepository;
   protected _transactionEventRepository: ITransactionEventRepository;
+  protected _costCalculator: CostCalculator;
 
   constructor({
     logger,
     ocppSender,
     authorizationRepository,
     transactionEventRepository,
+    costCalculator,
   }: AbstractHandlerDependencies & {
     ocppSender: IOcppSender;
     authorizationRepository: IAuthorizationRepository;
     transactionEventRepository: ITransactionEventRepository;
+    costCalculator: CostCalculator;
   }) {
     super(logger);
 
     this._ocppSender = ocppSender;
     this._authorizationRepository = authorizationRepository;
     this._transactionEventRepository = transactionEventRepository;
+    this._costCalculator = costCalculator;
   }
 
   async handle(
@@ -155,13 +160,31 @@ export class StopTransactionRequestOcpp16Handler extends AbstractHandler {
       transaction.totalKwh = (request.meterStop - transaction.startTransaction.meterStart) / 1000; // Convert from Wh to kWh
     } else {
       this._logger.warn(
-        `StartTransaction record not found at station ${ocppConnectionName} for transactionId ${request.transactionId}. 
+        `StartTransaction record not found at station ${ocppConnectionName} for transactionId ${request.transactionId}.
         Cannot calculate totalKwh.`,
       );
     }
     transaction.isActive = false;
     transaction.stoppedReason = stoppedReason;
     transaction.endTime = request.timestamp;
+
+    // Sole owner of the final timeSpentCharging: 1.6 reports no charging duration, so the whole
+    // session is billed as charging time. Always set, so cost calculation never sees it unset.
+    const startTime = transaction.startTime ? Date.parse(transaction.startTime) : Number.NaN;
+    const stopTime = Date.parse(request.timestamp);
+    const elapsedSeconds = Math.floor((stopTime - startTime) / 1000);
+    transaction.timeSpentCharging = Number.isFinite(elapsedSeconds)
+      ? Math.max(0, elapsedSeconds)
+      : 0;
     await transaction.save();
+
+    const totalCost = await this._costCalculator.calculateTotalCost(tenantId, transaction);
+    if (totalCost != null) {
+      await this._transactionEventRepository.updateTransactionTotalCostById(
+        tenantId,
+        totalCost,
+        transaction.id,
+      );
+    }
   }
 }
--- a/packages/ocpp/src/modules/california-pricing/handlers/default-price-ocpp-16-handler.ts
+++ b/packages/ocpp/src/modules/california-pricing/handlers/default-price-ocpp-16-handler.ts
@@ -1,0 +1,62 @@
+// SPDX-FileCopyrightText: 2026 Contributors to the CitrineOS Project
+//
+// SPDX-License-Identifier: Apache-2.0
+import {
+  AbstractHandler,
+  type AbstractHandlerDependencies,
+  AsResponseHandler,
+  type IMessage,
+} from '@citrineos/base';
+import { type HandlerProperties, OCPP1_6, OCPP_CallAction, OCPPVersion } from '@citrineos/types';
+import type { CaliforniaPricingService } from '../california-pricing-service.js';
+import { CUSTOM_DISPLAY_COST_AND_PRICE_KEY, DEFAULT_PRICE_KEY } from '../costmsg.js';
+
+/**
+ * Observes 1.6 GetConfiguration responses
+ * when the charger reports CustomDisplayCostAndPrice=true
+ * sets the station-wide DefaultPrice
+ */
+@AsResponseHandler([OCPPVersion.OCPP1_6], OCPP_CallAction.GetConfiguration)
+export class DefaultPriceCaliforniaPricingOcpp16Handler extends AbstractHandler {
+  protected _californiaPricingService: CaliforniaPricingService;
+
+  constructor({
+    logger,
+    californiaPricingService,
+  }: AbstractHandlerDependencies & {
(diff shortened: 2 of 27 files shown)


packages/ocpp/src/handlers/requests/1.6/boot-notification-request-ocpp-16-handler.ts
```diff
@@ -170,10 +170,14 @@ export class BootNotificationRequestOcpp16Handler extends AbstractHandler {
       bootNotificationResponse.status !== OCPP1_6.BootNotificationResponseStatus.Pending ||
       (cachedBootStatus && cachedBootStatus === OCPP1_6.BootNotificationResponseStatus.Pending)
     ) {
+      // Accepted boots skip the pending sync below but still need GetConfiguration, so the
+      // ChangeConfiguration table stays populated
+      if (bootNotificationResponse.status === OCPP1_6.BootNotificationResponseStatus.Accepted) {
```
thanaParis: This was very deliberately left out. Configuration doesn't change much, and boots are sometimes sent on reconnects not just resets--GetConfiguration on active boots results in significant unnecessary chatter in the field. At best, this should be a configurable option rather than the default behavior.

packages/ocpp/src/handlers/requests/1.6/stop-transaction-request-ocpp-16-handler.ts
```diff
@@ -155,13 +160,33 @@ export class StopTransactionRequestOcpp16Handler extends AbstractHandler {
...
+    // session is billed as charging time. Always set, so cost calculation never sees it unset.
+    const startTime = transaction.startTime ? Date.parse(transaction.startTime) : Number.NaN;
+    const stopTime = Date.parse(request.timestamp);
+    const elapsedSeconds = Math.floor((stopTime - startTime) / 1000);
+    transaction.timeSpentCharging = Number.isFinite(elapsedSeconds)
+      ? Math.max(0, elapsedSeconds)
+      : 0;
     await transaction.save();
+
+    if (transaction.totalKwh != null) {
```
thanaParis: this check is problematic, if time is being billed, specifically. an unusual situation these days, but not unheard of. not sure if there's a better "does this transaction need to have a cost" check that's broadly applicable--might just not need a check.

packages/ocpp/src/modules/california-pricing/handlers/default-price-ocpp-16-handler.ts
```diff
@@ -0,0 +1,56 @@
...
+import type { CaliforniaPricingService } from '../california-pricing-service.js';
+import { CUSTOM_DISPLAY_COST_AND_PRICE_KEY } from '../costmsg.js';
+
+/**
+ * Observes 1.6 GetConfiguration responses — the point at which the charger has connected and its
+ * configuration has been read. When the charger reports CustomDisplayCostAndPrice=true, sets the
+ * station-wide DefaultPrice. Reads the flag from the response payload (not the DB) to avoid racing
+ * the Configuration module's handler that persists these keys.
+ */
+@AsResponseHandler([OCPPVersion.OCPP1_6], OCPP_CallAction.GetConfiguration)
```
thanaParis, line 19: While this makes some sense to me, having it be the only place for this check doesn't work out I don't think. This isn't a value that would make sense to be changed locally--ergo, if it's changing, the CSMS almost certainly changed it with a ChangeConfiguration. This check should also go in the ChangeConfiguration response handler then--and that's probably where you want it to trigger most of the time?

packages/types/src/config/types.ts
```diff
@@ -319,6 +319,14 @@ export const configSchema = z.object({
       enableGetChargingProfilesOnStartTransaction: z.boolean().default(false),
     })
     .prefault({}),
+
+  // OCA "OCPP & California Pricing Requirements" (org.openchargealliance.costmsg) for OCPP 1.6.
+  // Off by default: enabling it starts an extra module and queue, and adds a per-message lookup.
```
thanaParis: This comment isn't necessary. Please be careful about letting comments like this one build up when dealing with LLM output.

thanaParis (changes requested):
Thank you for this excellent PR! I have some nits to pick, but this is really great to see; we've been wanting to cover this for some time and haven't been able to prioritize it.

packages/ocpp/src/handlers/requests/1.6/boot-notification-request-ocpp-16-handler.ts
andreosasLarsson: makes sense, would you be fine if i add a seperate parameter, something like GET_CONFIGURATION_ON_BOOT defaulting to false?

packages/ocpp/src/modules/california-pricing/handlers/default-price-ocpp-16-handler.ts
andreosasLarsson: i am less afraid that the EVSE or CSMS would change it. 
Rather that the tariff is changed, or it has yet to be set(connected to the bootnotification changes), 
this ensures newly connected chargers have the correct default price

this still might not be the best place for it.
and it still has some issue that i am unsure how to solve such as modifying the tariff connected to a connector does not update the default price.

packages/types/src/config/types.ts
andreosasLarsson: removed the comments

packages/ocpp/src/handlers/requests/1.6/stop-transaction-request-ocpp-16-handler.ts
andreosasLarsson: wrote that code before cost used flattfee and time for calculations, and forgot to update that part

Fixed now

packages/ocpp/src/handlers/requests/1.6/boot-notification-request-ocpp-16-handler.ts
thanaParis: Yes, that would work! It should go in the SystemConfig, probably in the 'ocpp' block alongside getBaseReportOnPending.
config.ocpp.getConfigurationOnBoot should put the env var as 'CITRINEOS_OCPP_GETCONFIGURATIONONBOOT' for reference.

packages/ocpp/src/modules/california-pricing/handlers/default-price-ocpp-16-handler.ts
thanaParis: Ah, I understand the issue better now.

Updating tariffs is a very problematic thing. The simplest solution, operationally, is to manually trigger the message when you make the change; if that's feasible, it could be worked as an option on the OperatorUI, as a command that can be sent from the tariff detail page--push a ChangeConfig with the new default price to all stations associated with the tariff. This can be paired with always sending the change config after boot; while still behavior that needs to be configurable, a single ChangeConfig with updated pricing is much less onerous than a full GetConfig on every boot.

If the operator UI is not where you foresee the changes happening, @AsDtoEventHandlers for Tariff Inserts and Updates could also be a place to touch; you could add configurable behavior there for emitting updates to charging stations whenever tariffs get modified. That would be trickier though.

If the behavior of sending GetConfigs on every boot, then updating the price in reaction to them, works for you then I suppose there's no need to optimize without cause. Let me know what you think.

packages/ocpp/src/handlers/requests/1.6/boot-notification-request-ocpp-16-handler.ts
andreosasLarsson: I did some reworking, getConfiguration now only asks for CustomDisplayCostAndPrice and only if the feature is enabled.

So any backends not interested in this feature will be unaffected

packages/ocpp/src/modules/california-pricing/handlers/default-price-ocpp-16-handler.ts
andreosasLarsson: I have reworked the tariff handling to avoid unnecessary reads from the database. and now only bases if we should write a new Default price on the getConfiguration result.

Regarding how to dynamically handle tariff updates:
Using the operator UI, should work but in my opinion adds a lot of extra endpoints/modifications on existing, as we need to handle both the case of a tariff being modified and a connector changing tariff.

Using @AsDtoEventHandlers looks interesting, but will not work if OCPI is not active.
looking at the code surrounding it i found that pg_notify would work for these cases, but would add a lot of extra strain as connectors update quite frequently.

one alternative would be to add an additional pg_notify, which only tracks tariff modifications to a Connector, but that would then trigger regardless of if California pricing is enabled.
I haven't looked too much into OCPP 2.1 and perhaps it would be relevant there as well.

thanaParis (approved):
Thank you so much for the PR!

PR #1030, opened 2026-09-09
fixing rabbitmq retries

thanaParis:
https://github.com/citrineos/citrineos/issues/223

instead of nacking, retry now acks and republishes to the queue. this puts the message at the tail of the queue rather than the head, meaning the order of messages is no longer guaranteed, however it reduces the strain of retries on the router. Given CSMS-initiated messages rarely come back-to-back, meaning order is not a concern, this is a worthy tradeoff.


yuvenus (approved):
Working great locally so far, woohoo!

PR #1032, opened 2026-09-10
fix(dal): declare what the migrations already enforce

sjh37:
## Summary

#930's schema validator reports 15 warnings against a database built by the migrations, on every install. They are all model-side: the models say less than the migrations do. This PR brings the models into line, updates the code around them so it no longer allows for a null the database cannot hold, and adds a test that keeps the models and migrations in line.

Verified by applying every migration to a fresh `postgis/postgis:16-3.4-alpine` and running `validateSequelizeSchema`: **15 warnings before, 0 after.** No migration is added or changed.

## The 15 warnings

**NOT NULL in the database, not declared in the model (11)**

| Column | Change |
|---|---|
| `transactionCreatedAt` on `ChargingNeeds`, `MeterValues`, `StartTransactions`, `StopTransactions`, `TransactionEvents` | `allowNull: false`, and the hook that resolves it moves from `@BeforeCreate` to `@BeforeValidate` - see below |
| `InstalledCertificates.hashAlgorithm`, `.issuerNameHash`, `.issuerKeyHash`, `.serialNumber` | `allowNull: true` → `false`. The migration has always declared these NOT NULL, and `CertificateHashDataType` requires all four |
| `Tenants.isUserTenant` | `allowNull: false` (it already had `defaultValue: false`) |
| `ServerNetworkProfiles.dynamicTenantResolution` | `allowNull: false, defaultValue: false`, matching the migration |

**In the database, not declared in the model (4)**

| Column | Change |
|---|---|
| `customData` on `Authorizations`, `MeterValues`, `LocalListAuthorizations` | `@Column(DataType.JSONB)`. Each model already had `declare customData` without a `@Column`, so Sequelize never read or wrote it - a `customData` in an Authorize or MeterValues was silently dropped, and `AuthorizationMapper` always emitted `undefined`. `Transaction` already maps its `customData` the same way |
| `StartTransactions.idTokenDatabaseId` | `@Column(DataType.INTEGER)`, nullable. Declared rather than dropped because Hasura exposes it and the operator UI selects it (`apps/operator-ui/src/lib/queries/transactions.ts`); removing it is a separate change touching the UI and metadata |

### Why the partition-key hook moves to `BeforeValidate`

Sequelize runs `allowNull` validation *before* `beforeCreate`, so with `allowNull: false` the existing `@BeforeCreate` hook would run too late and every insert would fail with a `notNull Violation`. `@BeforeValidate` runs first. The hooks already return early when the value is set, so on update they are no-ops; none of the five models is written through `bulkCreate` or a static `update()`, where validate hooks do not run.

## The code around those columns

**InstalledCertificate hash data.** `issuerNameHash`, `issuerKeyHash` and `serialNumber` are now non-null in `InstalledCertificateDto`, the model and the Drizzle schema (which also gets `.notNull()` on `hashAlgorithm`), and the repository create input no longer makes `hashAlgorithm` optional. The Drizzle mapper's `?? null`, the `?? ''` in `deleteByStationAndHashData`, and the optional fields on `installedCertificateKey` in the GetInstalledCertificateIds handler are gone.

Making the hash data required showed two paths that created an `InstalledCertificate` with none at all: `finalizeInstalledCertificate`, when a station accepts an InstallCertificate or CertificateSigned and no row exists yet, and `handleUploadExistingCertificate`. On a migrated database both already fail with a NOT NULL violation; they only worked on a schema built by `sync()`. Both now derive the hash data from the certificate with a new `getCertificateHashData` in `certificate-util.ts`:

- SHA-256 over the DER of the certificate's issuer name, and over the issuer's subject public key (the BIT STRING value, as in an RFC 6960 CertID);
- the serial number in hex without leading zeros, as `CertificateHashDataType` describes it in 2.1;
- the certificate is hashed against its own key if it is self-signed, otherwise against the next certificate in the chain (the CertificateSigned chain is stored whole), and it throws if the named issuer did not sign it. The upload path computes it before writing any certificate record.

The results match `openssl ocsp -req_text` for the sample root, and for the sample leaf with its sub-CA.

**DeleteCertificateAttempts** keeps nullable hash columns in its migration, so the DeleteCertificate response handler now only deletes the installed record when all four values are present. The outcome is unchanged: a null never matched a row.

**transactionCreatedAt.** The five model declarations are no longer optional, since every row read back has one. The Zod DTOs keep it optional, as they do `id` and `createdAt`: the hook assigns it on write, and the same DTOs describe meter values that have not been saved yet (the meter-value mappers, `MeterValueUtils`, the 1.6 MeterValues and StopTransaction handlers). Making it required there would mean a separate type for every pre-persist value, which is a DTO redesign rather than null handling.

## Test

`packages/ocpp/test/util/sequelize-schema-validator-migrations-integration.test.ts` applies every migration in `apps/ocpp-server/migrations` to a testcontainers Postgres, in file-name order as sequelize-cli does, and asserts the validator reports no errors and no warnings. The existing integration test validates against `sequelize.sync()`, which builds the schema *from* the models and so cannot see model-vs-migration drift; this one can. Against `next` without the model changes it fails with exactly these 15 warnings.

Also added: `getCertificateHashData` tests against the sample certificates (self-signed root, leaf plus sub-CA, and a leaf with no issuer); a `finalizeInstalledCertificate` integration test that records an accepted root and checks its stored hash data, which fails with a `notNull Violation` on all four columns without the fix; and upload-path tests that the hash data is passed through and that no certificate record is written when it cannot be derived.

Not touched: the Drizzle repositories, which have their own `customData is not persisted` note for Authorization, and the Sequelize `upsertServerNetworkProfile`, which does not copy `dynamicTenantResolution` from the websocket server config - both outside this PR.

Full suite: build clean, `pnpm test` green (333 files, 4432 tests), `pnpm run lint` clean.

### DIFF
--- a/packages/dal/src/db/drizzle/schema/installed-certificate.ts
+++ b/packages/dal/src/db/drizzle/schema/installed-certificate.ts
@@ -15,10 +15,10 @@ function installedCertificateColumns() {
     // FK to ChargingStation (resolved from ocppConnectionName in the domain layer).
     stationId: integer('stationId'),
     ocppConnectionName: varchar('ocppConnectionName', { length: 36 }).notNull(),
-    hashAlgorithm: varchar('hashAlgorithm', { length: 255 }),
-    issuerNameHash: varchar('issuerNameHash', { length: 255 }),
-    issuerKeyHash: varchar('issuerKeyHash', { length: 255 }),
-    serialNumber: varchar('serialNumber', { length: 255 }),
+    hashAlgorithm: varchar('hashAlgorithm', { length: 255 }).notNull(),
+    issuerNameHash: varchar('issuerNameHash', { length: 255 }).notNull(),
+    issuerKeyHash: varchar('issuerKeyHash', { length: 255 }).notNull(),
+    serialNumber: varchar('serialNumber', { length: 255 }).notNull(),
     certificateType: varchar('certificateType', { length: 255 }).notNull(),
     // FK to Certificate.
     certificateId: integer('certificateId'),
--- a/packages/dal/src/models/authorization/authorization.ts
+++ b/packages/dal/src/models/authorization/authorization.ts
@@ -125,6 +125,7 @@ export class Authorization extends Model implements AuthorizationDto {
   @Column(DataType.DECIMAL)
   declare prepaidBalance?: number | null;
 
+  @Column(DataType.JSONB)
   declare customData?: any | null;
 
   // For cases where Authorization is owned by an upstream partner, i.e. an eMSP
--- a/packages/dal/src/models/authorization/local-list-authorization.ts
+++ b/packages/dal/src/models/authorization/local-list-authorization.ts
@@ -101,6 +101,7 @@ export class LocalListAuthorization extends Model implements AuthorizationRestri
   @BelongsToMany(() => LocalListVersion, () => LocalListVersionAuthorization)
   declare localListVersions?: LocalListVersionDto[];
 
+  @Column(DataType.JSONB)
   declare customData?: any | null;
 
   @ForeignKey(() => Tenant)
--- a/packages/dal/src/models/certificate/installed-certificate.ts
+++ b/packages/dal/src/models/certificate/installed-certificate.ts
@@ -44,25 +44,25 @@ export class InstalledCertificate extends Model implements InstalledCertificateD
 
   @Column({
     type: DataType.STRING,
-    allowNull: true,
+    allowNull: false,
   })
   declare hashAlgorithm: HashAlgorithmEnumType;
 
   @Column({
     type: DataType.STRING,
-    allowNull: true,
+    allowNull: false,
   })
   declare issuerNameHash?: string | null;
 
   @Column({
     type: DataType.STRING,
-    allowNull: true,
+    allowNull: false,
   })
   declare issuerKeyHash?: string | null;
 
   @Column({
     type: DataType.STRING,
-    allowNull: true,
+    allowNull: false,
   })
   declare serialNumber?: string | null;
 
--- a/packages/dal/src/models/certificate/installed-certificate.ts
+++ b/packages/dal/src/models/certificate/installed-certificate.ts
@@ -52,19 +52,19 @@ export class InstalledCertificate extends Model implements InstalledCertificateD
     type: DataType.STRING,
     allowNull: false,
   })
-  declare issuerNameHash?: string | null;
+  declare issuerNameHash: string;
 
   @Column({
     type: DataType.STRING,
     allowNull: false,
   })
-  declare issuerKeyHash?: string | null;
+  declare issuerKeyHash: string;
 
   @Column({
     type: DataType.STRING,
     allowNull: false,
   })
-  declare serialNumber?: string | null;
+  declare serialNumber: string;
 
   @Column({
     type: DataType.STRING,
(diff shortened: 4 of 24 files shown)


yuvenus (changes requested):
One reason I left out adding all the "allowNull: false" declarations is because I didn't want to also have to update all the code to expect a non-optional value (and therefore make the schema drift PR more dense).

If you're making this update, please also update any surrounding code to no longer expect a null value, which will be a more complete solution than only doing half of it like it is now. Thanks!

sjh37:
@yuvenus Done in d86b38e. The DTO, model and Drizzle schema now make the InstalledCertificate hash fields non-null, and the repositories and handlers no longer allow for null.

Doing that showed two places that created an InstalledCertificate with no hash data at all: `finalizeInstalledCertificate` when a station accepts an install, and `handleUploadExistingCertificate`. Both already failed on a migrated database. They now derive the hash data from the certificate, with tests. I left `transactionCreatedAt` optional on the DTOs, as `id` and `createdAt` are, since the hook assigns it on write; the model declarations are no longer optional. There's more detail in the description.

yuvenus:
> @yuvenus Done in [d86b38e](https://github.com/citrineos/citrineos-core/commit/d86b38eecf28a259db29f0b653a641e0746be03d). The DTO, model and Drizzle schema now make the InstalledCertificate hash fields non-null, and the repositories and handlers no longer allow for null.
> 
> Doing that showed two places that created an InstalledCertificate with no hash data at all: `finalizeInstalledCertificate` when a station accepts an install, and `handleUploadExistingCertificate`. Both already failed on a migrated database. They now derive the hash data from the certificate, with tests. I left `transactionCreatedAt` optional on the DTOs, as `id` and `createdAt` are, since the hook assigns it on write; the model declarations are no longer optional. There's more detail in the description.

I see, to keep this PR focused to just fixing the models, could you make a PR stack with those certificate changes you added? Unless the hash data directly affects storing data in the database. 🤔 

Additionally, I didn't know the tip about @BeforeValidate, very interesting! To confirm though that decorator shouldn't mess up how partitioning works for the affected tables?

sjh37:
@yuvenus On splitting out the certificate changes: the hash data does directly affect storing data. The migration creates `hashAlgorithm`, `issuerNameHash`, `issuerKeyHash` and `serialNumber` on `InstalledCertificates` as `NOT NULL`, so a row without them cannot be inserted.

On `next` today, `finalizeInstalledCertificate` (a station accepts an install and no row exists yet) and `handleUploadExistingCertificate` both insert without them, and both fail with a `NOT NULL` violation on any database built by the migrations. Once the DTO is non-null, as you asked on the 10th, those two call sites no longer compile unless they supply the hash data, so the derivation has to land with the model change or before it. If you would still rather review it on its own, I can move `getCertificateHashData` and the two call sites into a separate PR against `next` and stack this one on top of it.

On `@BeforeValidate` and partitioning: it doesn't change where rows go. The hook sets the same value the `@BeforeCreate` one did, the parent Transaction's `createdAt`; it just runs before Sequelize's `allowNull` check instead of after it, and Postgres routes the row by the value in the INSERT. It can't move a row on update either: it returns early when the key is loaded, and when it isn't, Sequelize writes only the columns the caller changed, so a value the hook sets during validation never reaches the UPDATE.

I checked both against a database built by the migrations. Children of a transaction dated next month (TransactionEvents, StartTransactions, StopTransactions, ChargingNeeds, and MeterValues linked by transaction, stop transaction or event) all landed in the `_2026m10` partitions with the transaction's `createdAt`. Instance `update()` and `save()` on rows loaded without the key, and static `update()`, produced UPDATEs with no `transactionCreatedAt` in them, for example `UPDATE "MeterValues" SET "transactionId"=$1,"updatedAt"=$2 WHERE "id" = $3`, and every row stayed in its partition. I can add that as an integration test here if you want it pinned.

yuvenus:
Closing because there will be upcoming work that resolves the same thing, and also introduces a few more fixes.

PR #1035, opened 2026-09-10
adding logging middleware to catch keyCode for redaction

thanaParis:
preferred fix for https://github.com/citrineos/citrineos-core/pull/965

PR #1038, opened 2026-09-11
fix(ocpp): answer Authorize Invalid when a token fails its format check

sjh37:
## The defect

`AuthorizeRequestOcpp201Handler` and `AuthorizeRequestOcpp21Handler` check the idToken against the format for its type after schema validation (an ISO14443 UID must be 8 or 14 hex characters, a key code only letters, digits and `* - _ = : + | @ .`, and so on). When the check fails, both answer with a `PropertyConstraintViolation` CALLERROR.

A CALLERROR carries no `IdTokenInfo`, so the station is left with no authorisation decision for the token it asked about. C04.FR.01 says what the CSMS should do instead:

> When the CSMS receives an AuthorizeRequest with a keyCode that is not valid at this Charging Station, the CSMS SHALL respond with an AuthorizeResponse message with status = Invalid.

The 2.0.1 handler already built exactly that response in this branch, then discarded it and sent the CALLERROR.

## The fix

Both handlers now send the `AuthorizeResponse` with `idTokenInfo.status = Invalid` through the same `_sendAuthorizeResult` path as every other outcome, so the result is also recorded in the Authorize metric. That applies to every format failure, not only key codes: a token that cannot be well formed for its type cannot be valid either, and `Invalid` is the status that says so.

Untouched: the format check itself, and the two log lines in that branch. Key code redaction in the log is citrineos/citrineos-core#1035's.

## Notes for review

- `apps/operator-ui/tests/e2e/fixtures/everest.ts` has a comment saying core answers such an Authorize with a `PropertyConstraintViolation` CALLERROR. The outcome it relies on - the session is not authorised - is the same with `Invalid`. I have left it alone rather than switch the operator-ui e2e lanes on for a change that is not in the UI; happy to reword it here if you would rather.

## Tests

One per handler: a `KeyCode` token containing a space. Before the fix both fail with no `CallResult` sent (the 2.0.1 one: `expected "spy" to be called once, but got 0 times`); after it, both receive `Invalid`.

## Verification

On `next` at `0fd4e1f`: build clean, full suite green (331 files, 4420 passing / 8 skipped), `pnpm run lint` 0 errors.

### DIFF
--- a/packages/ocpp/src/handlers/requests/2/authorize-request-ocpp-201-handler.ts
+++ b/packages/ocpp/src/handlers/requests/2/authorize-request-ocpp-201-handler.ts
@@ -9,7 +9,6 @@ import {
   type IMessage,
   type IMessageConfirmation,
   type IOcppSender,
-  OcppError,
   recordAuthorizeResult,
 } from '@citrineos/base';
 import {
@@ -17,7 +16,6 @@ import {
   AuthorizationStatusEnum,
   type AuthorizationStatusEnumType,
   AuthorizeCertificateStatusEnum,
-  ErrorCode,
   type HandlerProperties,
   IdTokenEnum,
   OCPP2_0_1,
@@ -91,12 +89,6 @@ export class AuthorizeRequestOcpp201Handler extends AbstractHandler {
         token: request.idToken.idToken,
         error: tokenValidation.errorMessage,
       });
-      const messageId = message.context.correlationId;
-      const error = new OcppError(
-        messageId,
-        ErrorCode.PropertyConstraintViolation,
-        tokenValidation.errorMessage || 'Invalid token value for specified type',
-      );
       response = {
         ...response,
         idTokenInfo: {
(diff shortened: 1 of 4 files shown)


thanaParis:
The cited portion of the spec does not mean what you are saying it does.

> When the CSMS receives an AuthorizeRequest with a keyCode that is not valid at this Charging Station, the CSMS SHALL respond with an AuthorizeResponse message with status = Invalid.

The keycode is not invalid at that charging station. It is structurally invalid--it violates a property constraint. This is the difference between a key being the wrong one for a lock and a key being a paperclip; one is invalid, the latter is the wrong format.

sjh37:
@thanaParis Fair - I conflated the two. C04.FR.01 is the right key for the wrong lock; a token that breaks the format for its type is a field that fails its constraint, which is exactly what `PropertyConstraintViolation` is for in Part 4. Withdrawn, and I won't reopen it.

PR #1043, opened 2026-09-14
fix(dal): store the StatusNotification fields instead of spreading the model

sjh37:
## Summary

`SequelizeLocationRepository.addStatusNotificationToChargingStation` builds the row with
`StatusNotification.build({ ...statusNotification, tenantId, stationId })`, but both callers in
`StatusNotificationService` pass a built `StatusNotification` model instance. Spreading a Sequelize instance copies
its internals (`dataValues`, `_previousDataValues`, `_options`, ...), not its attributes, and `build()` ignores those
keys. So every StatusNotification row is stored with only `tenantId` and `stationId`: `ocppConnectionName`,
`connectorStatus`, `evseId`, `connectorId`, `timestamp` and the error fields are all null. The change reads the fields
explicitly, the same way the Drizzle repository's `addStatusNotificationToChargingStation` already does.

The spread arrived in 4b340dec0 (#1010), when the parameter type changed from the model to `StatusNotificationDto`.
Before that the instance was saved as it was.

## Where it bites

- Every 2.0.1/2.1 `StatusNotificationRequest` (`processStatusNotification`) and every 1.6 `StatusNotification`
  (`processOcpp16StatusNotification`) on the default Sequelize path writes an empty row.
- `updateLatestStatusNotification` then looks up earlier rows by `evseId`/`connectorId`, which are null on every row,
  so it deletes the previous pointer every time. A station whose connectors 1 and 2 both report ends up with one
  `LatestStatusNotifications` row, pointing at an empty notification. The operator UI reads this table for connector
  status (`charger-activity-card.tsx`, `charging-station-dto.tsx`), so it shows no status, and its per-EVSE match on
  `evseId`/`connectorId` never hits.
- Connector rows are written separately and are unaffected, which is why the existing tests did not notice.

## Test

`packages/ocpp/test/modules/transactions/status-notification-service-integration.test.ts` (testcontainers) gains three
cases that drive the real `SequelizeLocationRepository` through `StatusNotificationService`:

- a 2.0.1 notification stores `ocppConnectionName`, `evseId`, `connectorId`, `connectorStatus` and `timestamp`;
- a 1.6 notification stores `ocppConnectionName`, `connectorId`, `connectorStatus`, `errorCode` and `timestamp`;
- connectors 1 and 2 on one EVSE leave one `LatestStatusNotification` each.

On next the first two fail with `expected { timestamp: null, id: 1, ... } to match object`, and the third with
`expected [ null ] to deeply equal [ 1, 2 ]`. All three pass with the fix.


packages/dal/src/repositories/sequelize/location.ts
```diff
@@ -154,7 +154,19 @@ export class SequelizeLocationRepository
     const stationId = await resolveStationId(tenantId, ocppConnectionName);
     const savedStatusNotification = await this.statusNotification.create(
       tenantId,
-      StatusNotification.build({ ...statusNotification, tenantId, stationId }),
+      StatusNotification.build({
```
yuvenus, line 157: Good catch, can confirm the behavior on my local as well! Quick question regarding this change though -- previously, the code was:

```
const savedStatusNotification = await this.statusNotification.create(
  tenantId,
  statusNotification,
);
```

Perhaps it just makes more sense then to resolve this by NOT using StatusNotification.build instead:

```
const savedStatusNotification = await this.statusNotification.create(
  tenantId,
  { ...statusNotification, tenantId, stationId },
);
```

cc: @lydiazcheng since you originally changed this -- any reason to have to do StatusNotification.build over just spreading the DTO that comes in from the function parameters? 👀

lydiazcheng: Thank you both for fixing and reviewing. I don't have any concerns on changing this. Another PR has been merged for fixing this issue: https://github.com/citrineos/citrineos-core/pull/1070. Not sure if we still need this.

yuvenus: He beat us to the punch 😄 Okay, I will close this PR then since it's resolved. 👍

yuvenus:
Closing since this is already resolved by https://github.com/citrineos/citrineos-core/pull/1070.

PR #1045, opened 2026-09-14
fix(dal): keep the 2.1 transactionInfo tariffId out of Transactions.tariffId

sjh37:
## Summary

OCPP 2.1 adds `TransactionType.tariffId`, the id string of the tariff in use. I08.FR.13: "When a tariff applies to the
transaction, then the id of the tariff in use, is added to the transactionInfo element." I07.FR.32 requires the station
to send it for a default tariff installed via SetDefaultTariff. For a new transaction,
`createOrUpdateTransactionByTransactionEventAndStationId` spreads `transactionInfo` into `Transaction.build`, so this
string lands in the integer `Transactions.tariffId`. It is only replaced by the resolved Tariff row id when the event
carries `evse.connectorId`. Otherwise the insert fails with `invalid input syntax for type integer` and no transaction
is stored. The change strips `tariffId` from the spread, resolves it to the Tariff row before the transaction is built,
and keeps the connector's tariff as the fallback, exactly as the connector branch already did.

## Where it bites

- A 2.1 station with a default tariff (which the CSMS installs with SetDefaultTariff) starts a transaction at
  TxStartPoint Authorized. The Started `TransactionEventRequest` carries `transactionInfo.tariffId` and no `evse`, or
  an `evse` without `connectorId`. E01.FR.16 only sends `evse` in the first event after the EV connects, and
  `EVSEType.connectorId` is optional.
- The repository throws a `SequelizeDatabaseError`. The handler answers CALLERROR InternalError, and there is no
  Transaction row for the rest of the session, so every later event, meter value and cost calculation has nothing to
  attach to.
- The update path already overwrote the spread `tariffId` with the resolved id, so it behaved correctly. It now spreads
  the same stripped object, and its lookup is unchanged.

The comment reworded by #1032 in this file is untouched, and #1032's hunk still applies on top of this change.

## Test

`packages/dal/test/repositories/sequelize/transaction-event-tariff-integration.test.ts` (new, testcontainers) sends a
2.1 Started event with `transactionInfo.tariffId` and no `evse` through `SequelizeTransactionEventRepository`:

- with a matching Tariff row, the stored transaction's `tariffId` is that row's id;
- with no matching row, the transaction is still stored, with `tariffId` null.

On next both fail with `SequelizeDatabaseError: invalid input syntax for type integer: "TARIFF-1"`. Both pass with the
fix.


thanaParis (approved):

thanaParis (changes requested):
broken test

thanaParis (approved):

PR #1046, opened 2026-09-14
fix(dal): convert cacheExpiryDateTime back to a Date in Drizzle authorization updateByKey

sjh37:
## Summary

`DrizzleAuthorizationRepository.updateByKey` passes its `value` straight to `updateById`. The DTOs the same repository
returns carry `cacheExpiryDateTime` as an ISO string (`toAuthorizationDto`), but the column is a
`timestamp(..., { mode: 'date' })`, so Drizzle calls `.toISOString()` on the string. Updating an authorization that
has a cache expiry therefore throws `TypeError: value.toISOString is not a function`. The change converts a string
`cacheExpiryDateTime` back into a `Date` before the update, following `toBootEntity` in the Drizzle boot repository.
It only affects deployments that opt in with `CITRINEOS_USE_DRIZZLE=true`. The Sequelize repository accepts the string
and is unaffected.

## Where it bites

`RealTimeAuthorizer` records each real-time authorisation attempt by setting `realTimeAuthLastAttempt` on the
authorization it read and calling `updateByKey(tenantId, authorization, id)`. On the Drizzle path, for any token with a
`cacheExpiryDateTime`, this throws. The error is caught and logged ("Failed to save realTimeAuthLastAttempt"), so the
attempt is never stored. The `realTimeAuthTimeout` window that should reuse the last result then never applies, and
every Authorize for that token calls the external real-time authorisation URL again.

## Test

`packages/dal/test/repositories/drizzle/drizzle-authorization-integration.test.ts` gains a case that reads a token with
a `cacheExpiryDateTime` through the Drizzle repository, sets `realTimeAuthLastAttempt`, calls `updateByKey` with it,
and reads the attempt back. On next it fails with `TypeError: value.toISOString is not a function`, and it passes with
the fix.


yuvenus (approved):
Ah, another such case like the BootEntity :) Great catch!

PR #1047, opened 2026-09-14
fix(ocpp): keep the recorded 1.6 charging profiles in step with the station

sjh37:
## Summary

The CSMS record of which OCPP 1.6 charging profiles are active drifts from what the station actually holds, in two
places. 1.6 has no GetChargingProfiles, so unlike the 2.x handlers nothing ever resyncs the record afterwards.

- An Accepted ClearChargingProfile marks every active profile on the station inactive, whatever the request asked to
  clear. OCPP 1.6 6.13, as amended by errata 3.26: "clear (remove) either a specific charging profile (denoted by id)
  or a selection of charging profiles that match (logical AND) with the values of the optional connectorId, stackLevel
  and chargingProfilePurpose fields. If no fields are provided, then all charging profiles will be cleared. If id is
  specified, then all other fields are ignored." The handler now loads the original request by correlationId, as the
  1.6 SetChargingProfile response handler already does, and deactivates only the rows that match it.
- An Accepted SetChargingProfile upserts the new profile by id only. OCPP 1.6 5.16.3/5.16.4: "If a charging profile
  with the same chargingProfileId, or the same combination of stackLevel / ChargingProfilePurpose, exists on the Charge
  Point, the new charging profile SHALL replace the existing charging profile". The handler now marks the other active
  profile with the same connector, stackLevel and purpose inactive before storing the new one.

## Where it bites

- ClearChargingProfile: the station holds a ChargePointMaxProfile (id 1, connector 0) and a TxDefaultProfile (id 2,
  connector 1). `ClearChargingProfile{id: 2}` is Accepted. Both rows become `isActive = false`, although the station
  still enforces id 1, so the operator UI and the API report no active limit on the station. It should retire id 2
  only.
- SetChargingProfile: TxDefaultProfile id 10 (connector 1, stackLevel 0) is active. A new TxDefaultProfile id 11 with
  the same connector and stackLevel is Accepted. The station has replaced 10 with 11, but the CSMS shows both as
  active.

The matching follows how the 1.6 path already stores a profile: `evseId` holds the 1.6 connectorId as sent (so
connectorId 0 is a criterion, not "no criterion"), and the purpose is compared after
`ChargingProfileMapper.fromChargingProfilePurpose`, which stores ChargePointMaxProfile as ChargingStationMaxProfile.

Judgement calls:

- If the original ClearChargingProfile request cannot be found, the handler logs an error and changes nothing, the
  same as the SetChargingProfile response handler. Before this change it deactivated everything. The message persist
  processor is critical and always registered, so this should not happen in practice.
- 3.13.2 says the same stackLevel and purpose may not exist twice "in a Charge Point", which reads station-wide. The
  replacement is scoped to the same connector here. Connector-scoped is the conservative reading: it never retires a
  profile that a per-connector implementation still holds. The worst it can do on a station-wide implementation is
  leave one stale row active, which is what happens today.
- A 1.6 RemoteStartTransaction without a connectorId stores its TxProfile with no connector. Before this change any
  Accepted ClearChargingProfile happened to deactivate that row along with everything else; now a clear with
  `connectorId: 1` no longer matches it, although the station has cleared it. A later SetChargingProfile that replaces
  it does not retire it either, which is unchanged. Matching that row would need the connector the station actually
  started on, which the RemoteStart path does not record; I have left it as a known limitation rather than widen this
  change.

## Test

Two new integration suites (testcontainers Postgres) drive the real handlers against
`SequelizeChargingProfileRepository`, with the profiles stored exactly as the 1.6 SetChargingProfile path stores them:

- `packages/ocpp/test/handlers/responses/1.6/clear-charging-profile-response-ocpp-16-handler-integration.test.ts`
  pins clear by id, id overriding the other fields, connectorId 0, the purpose mapping, the logical AND, no criteria
  clearing everything, other stations untouched, and a missing original request leaving the record alone. On next, six
  of the eight fail, for example `expected [ 1, 2, 3 ] to deeply equal [ 2 ]`.
- `packages/ocpp/test/handlers/responses/1.6/set-charging-profile-response-ocpp-16-handler-integration.test.ts` pins
  the replacement and the cases that must not retire anything: another stackLevel, another purpose, another connector,
  the same id sent again, and another station. On next the replacement case fails with
  `expected [ 10, 11 ] to deeply equal [ 11 ]`.

Both suites pass with the fix.


thanaParis (approved):

thanaParis (changes requested):
broken tests.

thanaParis (approved):

PR #1048, opened 2026-09-14
fix(ocpp): answer every 1.6 StopTransaction with the idTagInfo Authorize would give

sjh37:
## Summary

The 1.6 StopTransaction handler looks up the idTag differently from Authorize and StartTransaction, in two ways.

- It uses `readOnlyOneByQuerystring`, which throws when the idTag matches more than one Authorization row. The throw
  happens before the StopTransaction.conf is sent and before the transaction is closed. OCPP 1.6 4.10: "Upon receipt
  of a StopTransaction.req PDU, the Central System SHALL respond with a StopTransaction.conf PDU", and its sanity
  checks "SHOULD NOT ever cause the Central System to not respond". The handler now reads every match and treats
  anything but exactly one as unknown (Invalid), as Authorize and StartTransaction already do. The transaction is still
  closed, without an authorization id.
- It passes the stored status straight through, so an Accepted token whose `cacheExpiryDateTime` has passed is
  answered Accepted, with an expiryDate in the past. The Charge Point updates its Authorization Cache from this
  IdTagInfo (4.10). Authorize and StartTransaction answer Expired in that case. StopTransaction now does the same.

## Where it bites

- Two rows, one idToken: the unique index is `(tenantId, idToken, idTokenType)`, so a tenant can hold `04A1B2C3` as
  ISO14443 for a 2.0.1 fleet and as Central. A 1.6 transaction ends with StopTransaction(idTag `04A1B2C3`). The
  handler throws and the station gets an InternalError CALLERROR. The station retries under the transaction-message
  rules, gets the same error each time and gives up. The Transaction stays `isActive`, with no meterStop and no
  totalKwh. With the fix the station gets `idTagInfo.status: Invalid`, and the transaction is closed with its
  meterStop.
- Expired token: an Accepted Authorization whose cacheExpiryDateTime was yesterday now yields
  `{status: Expired, expiryDate: yesterday}` instead of `{status: Accepted, expiryDate: yesterday}`. The impact is
  limited, because a Charge Point must treat a time-expired cache entry as expired anyway.

Judgement call: the ambiguous-token case logs "Too many authorizations found for idToken", the same message Authorize
logs, so an operator can see why a known token was answered Invalid.

## Test

`packages/ocpp/test/handlers/requests/1.6/stop-transaction-request-ocpp-16-handler.test.ts`. The authorization
repository fake now answers `readAllByQuerystring`, and its `readOnlyOneByQuerystring` throws on more than one row, as
the base repository does. New tests:

- Two matching rows: the reply is Invalid, and the StopTransaction is recorded and the transaction saved. On next both
  fail with `More than one value found for query`.
- A past cacheExpiryDateTime is answered Expired, which fails on next because the conf says Accepted. A future one is
  still answered Accepted.

All pass with the fix.

This PR and citrineos/citrineos-core#1049 both add tests at the end of `stop-transaction-request-ocpp-16-handler.test.ts`, so whichever merges second needs a small rebase, which I will do.


packages/ocpp/src/handlers/requests/1.6/stop-transaction-request-ocpp-16-handler.ts
```diff
@@ -57,17 +57,29 @@ export class StopTransactionRequestOcpp16Handler extends AbstractHandler {
...
+    const authorizations: AuthorizationDto[] = request.idTag
+      ? await this._authorizationRepository.readAllByQuerystring(tenantId, {
           idToken: request.idTag,
         })
-      : undefined;
+      : [];
+    if (authorizations.length > 1) {
+      this._logger.error(`Too many authorizations found for idToken: ${request.idTag}`);
+    }
+    const authorization: AuthorizationDto | undefined =
```
thanaParis, line 74: The correct behavior here, as established elsewhere in the code, is to return 'Invalid'

thanaParis (changes requested):

packages/ocpp/src/handlers/requests/1.6/stop-transaction-request-ocpp-16-handler.ts
sjh37: Yes you're right. It was because I had #848 merged into my own branch, which I've now removed.

thanaParis (approved):

PR #1051, opened 2026-09-14
fix(dal): scope two upsert lookups to the tenant

sjh37:
## Summary

Two Sequelize upserts look up an existing row by a raw `findOne` / static `Model.findOne` without `tenantId`, which
bypasses the tenant scoping the base repository adds. The repo treats two tenants sharing an `ocppConnectionName` as
supported (`shared-station-name-tenant-scope-integration.test.ts`, added in #999). #999 fixed `createChargingNeeds` and
`getTransactionsCount`; these two were missed:

- `SequelizeChargingProfileRepository.createOrUpdateChargingProfile` resolves a TxProfile's `transactionId` with
  `Transaction.findOne({ ocppConnectionName, transactionId })`.
- `SequelizeMessageInfoRepository.createOrUpdateByMessageInfoTypeAndStationId` looks up the existing display message
  by `{ ocppConnectionName, id }`.

Each `where` now includes `tenantId`. The Drizzle twin of the message-info upsert already filters on `tenantId`; the
Drizzle charging-profile repository has no `createOrUpdateChargingProfile` yet.

## Impact

Tenants A and B each have a station named `CP01`.

- Charging profile: transaction ids come from per-(tenant, station) sequences starting at 1, so both tenants soon
  have a transaction `7`. A 1.6 TxProfile for A's transaction 7, from SetChargingProfile or RemoteStartTransaction,
  is stored with `transactionDatabaseId` pointing at B's transaction 7, or at whichever row the database returns
  first. It should link to A's own transaction, or to none.
- Display message: B's `CP01` has shown message id 1. A's `CP01` reports message id 1 in NotifyDisplayMessages. The
  lookup finds B's row. The update is tenant-scoped, so it finds nothing, and nothing is stored for A. The method
  returns `undefined`.

`SequelizeCertificateRepository.createOrUpdateCertificate` has the same unscoped lookup, but it is left out on purpose:
on a database built by the migrations, `Certificates` still carries the global unique
`Certificates_serialNumber_issuerName_key` (no `tenantId`), so scoping the lookup alone would turn "returns undefined"
into a unique violation. It needs that constraint dropped first, and belongs with that change.

## Test

`packages/dal/test/repositories/sequelize/shared-station-name-tenant-scope-integration.test.ts` (testcontainers
Postgres) gains three cases alongside the #999 ones:

- A TxProfile is not linked to the other tenant's transaction with the same id. On next this fails with
  `expected 1 to be null`.
- When both tenants have that transaction id, the TxProfile links to the caller's own transaction. On next this fails
  with `expected 1 to be 2`.
- A display message is stored when the other tenant holds one with the same id. On next this fails with
  `expected undefined to be 1`.

All pass with the fix.


yuvenus (approved):

yuvenus:
Looks good, just some conflicts to resolve, thanks!

sjh37:
Resolved the conflicts.

PR #1056, opened 2026-09-14
fix(ocpp): answer contract-certificate Authorize with the C07 status pairs

sjh37:
## Summary

C07 fixes which `certificateStatus` and `idTokenInfo.status` go together when a station sends a contract certificate (chain or OCSP hash data) with its AuthorizeRequest. Both 2.x Authorize handlers, and `CertificateAuthorityService` behind them, return the wrong pair in three cases:

- C07.FR.15: an expired chain must give "certificateStatus = CertificateExpired and an idTokenInfo.status = Expired". The handlers send `Invalid` for every rejected certificate.
- C07.FR.13: a valid chain whose idToken is "one of Blocked, Expired, Invalid, Unknown" must give "certificateStatus = ContractCancelled and the authorization status in idTokenInfo.status". The handlers keep `Accepted`. When the authorization's cache has expired, or the token is not allowed at this EVSE or location, the handlers rebuild the response and drop `certificateStatus` altogether, which also breaks C07.FR.14.
- C07.FR.17: a chain that "cannot be verified or is invalid" must give "certificateStatus = CertChainError and an idTokenInfo.status = Invalid". `validateCertificateChainPem` and `validateCertificateHashData` return `NoCertificateAvailable` when no trust root matches, when the OCSP responder reports the certificate `unknown`, and when the OCSP call fails. The AuthorizeCertificateStatusEnumType table marks `NoCertificateAvailable` as `<not used>` and defines `CertChainError` as "If the contract certificate contained in the AuthorizeRequest message is not valid".

The change keeps to the mapping. The handlers map `CertificateExpired` to `Expired` (everything else rejected stays `Invalid`, which already matches C07.FR.16 for revoked). They keep `certificateStatus` when rebuilding the response. After a valid chain they set `ContractCancelled` when the final idTokenInfo status is Blocked, Expired, Invalid or Unknown. The service returns `CertChainError` in the three unverifiable cases, and for a certificate string that parses to no certificate at all (a truncated PEM without a complete BEGIN/END pair), which also gave `NoCertificateAvailable`. OCSP request building and the responder URL handling are untouched (`certificate-util.ts` and `ocsp-responder-url.ts` are not in the diff).

## Impact

These values feed the ISO 15118 ResponseCode the station returns to the EV, so the driver sees the wrong reason for a refusal:

- An EV presents an expired contract certificate. The CSMS answers `{CertificateExpired, Invalid}` instead of `{CertificateExpired, Expired}`.
- A valid contract certificate for an eMAID the CSMS does not know, or has blocked, gets `{Accepted, Unknown|Blocked}` instead of `{ContractCancelled, Unknown|Blocked}`. One whose authorization has lapsed gets `{Invalid}` with no `certificateStatus` at all.
- A chain the CSMS cannot anchor to a V2G root, or whose OCSP responder is down or does not know the certificate, gets `NoCertificateAvailable` instead of `CertChainError`.

Judgement call: ContractCancelled is applied to the final idTokenInfo status, including one set by a real-time authorizer or by the cache-expiry check, because C07.FR.13 is phrased in terms of "authorization status of idToken". NoCredit (2.1 prepaid), NotAtThisLocation and the other statuses keep `certificateStatus = Accepted` as C07.FR.14 requires.

## Test

- `packages/ocpp/test/handlers/requests/2/authorize-request-ocpp-201-handler-certificates.test.ts`: expired chain gives Expired; a valid chain with an unknown, blocked or lapsed eMAID gives ContractCancelled; a valid chain with an accepted eMAID stays Accepted/Accepted (control).
- `packages/ocpp/test/handlers/requests/2/authorize-request-ocpp-21-handler.test.ts`: the expired, unknown and blocked cases on 2.1.
- `packages/ocpp/test/services/certificate/certificate-authority.test.ts`: no matching trust root, an `unknown` OCSP status, and an unreachable responder (chain and hash data) give CertChainError. The existing no-trust-root and empty-chain cases expected NoCertificateAvailable and now expect CertChainError.

On next they fail with `expected 'Invalid' to be 'Expired'`, `expected 'Accepted' to be 'ContractCancelled'`, `expected undefined to be 'ContractCancelled'` and `expected 'NoCertificateAvailable' to be 'CertChainError'`. With the fix they all pass.

Verified on `next` at `d32c1c3`: the new test fails before the fix as described above; with the fix, and with the four other OCPP 2.x fixes raised alongside it merged together, build clean, `pnpm test` green (339 files, 4471 passing / 8 skipped), `pnpm run lint` 0 errors.


packages/ocpp/src/handlers/requests/2/authorize-request-ocpp-21-handler.ts
```diff
@@ -322,13 +329,29 @@ export class AuthorizeRequestOcpp21Handler extends AbstractHandler {
...
 
+    this._cancelContractIfNotAuthorized(response);
     const messageConfirmation = await this._sendAuthorizeResult(message, response);
     this._logger.debug(
       this.createHandlerSentMessageLog('Authorize 2.1 Response'),
       messageConfirmation,
     );
   }
 
+  private _cancelContractIfNotAuthorized(response: OCPP2_response_types.AuthorizeResponse): void {
```
lydiazcheng, line 345: Nice catch. Could you please help add a comment there to show the referred protocol content since this part is not so intuitive? Just a short one like, refer to OCPP C07.FR.13, as well as for 2.0.1 handler. Thank you.

sjh37: Hi @lydiazcheng 

I've added a short doc comment above `_cancelContractIfNotAuthorized` in both the 2.0.1 and 2.1 handlers, naming C07.FR.13 and what it requires: a contract certificate that was accepted is reported as `ContractCancelled` when the token itself is not authorized.

Thanks for the review.

lydiazcheng (approved):

thanaParis (approved):

PR #1060, opened 2026-09-14
fix(base): ignore an offline Transaction.Begin=0 as the meter baseline

mrnonaki:
## Problem

A charger that opens a transaction while still **offline** — for example it rebooted with the cable plugged in — cannot read its cumulative `Energy.Active.Import.Register` yet, so per its firmware it emits a `Transaction.Begin` meter value with `value: 0` as a placeholder. A later, real register reading arrives once the charger is back online (RemoteStart / periodic sampling).

`MeterValueUtils.getMeterStart()` takes `sorted[0]` (earliest timestamp) as the baseline unconditionally, so it locks `meterStart = 0`. `getTotalKwh()` (no `meterStart`) likewise uses `sorted[0]`. The next real register (e.g. `909948 Wh`) is then treated as delivered energy against a 0 baseline, and the transaction is charged the charger's **entire lifetime odometer**:

```
totalKwh = 910031 − 0 = 910.031 kWh   (billed)
actual   = 910031 − 909948 = 0.083 kWh (delivered)
```

This is deterministic: any reboot-with-cable-plugged followed by a real start on that transaction over-counts by the full meter reading. Observed on real OCPP 2.0.1 hardware.

### Where this was observed (real deployment, real money)

Hit in production against a commercially deployed AC charger (vendor Iocharger, model `IOCAW12-22TE-CRW4`, firmware `25101501`), OCPP 2.0.1, security profile 2. Iocharger is named by the Open Charge Alliance itself as an OCPP 2.0.1–certified charging-station vendor (Core + Advanced Security profiles — see the OCA announcement <https://openchargealliance.org/974/> and the OCA "frontrunner companies" page). The point being: this is a certified-vendor charger behaving **per the OCPP 2.0.1 spec** — a `Transaction.Begin` placeholder while offline is legitimate — so this is a CSMS-side accounting bug, not a rogue-device firmware quirk to be worked around per vendor. Any conformant charger that starts a transaction offline can produce the same 0 placeholder.

The exact sequence, taken from the charger's own OCPP log:

1. `22:10:23` — charger reboots (`BootNotification reason: ScheduledReset`) with the cable still plugged into a car.
2. `22:10:26` — on boot it auto-opens a transaction while the CSMS link is not up yet:
   `TransactionEvent { eventType: Started, triggerReason: CablePluggedIn, offline: true, meterValue: [{ value: 0, context: "Transaction.Begin", measurand: "Energy.Active.Import.Register", unitOfMeasure: { unit: "Wh" } }] }`
   → CitrineOS persists `meterStart = 0`.
3. `23:44:29` — a real `RequestStartTransaction` lands on that same transaction; the charger, now online, sends the true register `909948 Wh`.
4. `23:44–23:45` — real charging: `909948 → 910031 Wh` = **0.083 kWh delivered** over 89 s (~3.6 kW, plausible for a 22 kW AC unit).
5. `23:45:58` — stop; `Transaction.End` register `910031 Wh`.

Result: `Transactions.totalKwh = 910031 − 0 = 910.031 kWh`. The driver's wallet was charged for ~910 kWh of energy that was never delivered — a real over-charge that had to be refunded manually. The physical ceiling for this charger over 89 s is ~0.5 kWh, so the billed figure is not just wrong, it is physically impossible; nothing in the pipeline noticed because every layer trusted `totalKwh`.

## Fix

In `getRegisterValuesMap`, skip a register reading of exactly `0` whose context is `Transaction.Begin`. A cumulative register on a charger that has ever delivered energy is never genuinely `0` at `Transaction.Begin`, so this only drops the offline placeholder; the baseline then comes from the first real reading. If that placeholder is the *only* register reading in a batch, `getMeterStart` returns `null` and the caller waits for a real reading instead of persisting `0`.

Fixing it in `getRegisterValuesMap` covers both consumers through one point — `getMeterStart` and the no-`meterStart` branch of `getTotalKwh` — so all three call sites (`TransactionEventRepository` ×2, `TransactionService.recalculateTotalKwh`) are covered.

The guard is intentionally narrow (`value === 0` **and** `context === Transaction.Begin`) so no normal reading changes behaviour.

## Tests

Added to `packages/base/test/util/meter-value-utils.test.ts`:
- `getMeterStart` skips a 0 `Transaction.Begin` placeholder and returns the first real register
- `getMeterStart` returns `null` when the only register reading is a 0 `Transaction.Begin` placeholder (so the caller does not lock in 0)
- `getMeterStart` keeps a genuine 0 register at a non-`Transaction.Begin` context
- `getTotalKwh` (no `meterStart`) returns the delivered 0.083 kWh for the offline-reboot sequence instead of the whole odometer

All existing `meter-value-utils` tests still pass (60 assertions incl. type-check, 0 type errors).

## Notes for reviewers
- Rebased on current `next` (includes #868).
- No behaviour change for online transactions or any non-zero / non-`Transaction.Begin` reading.


thanaParis:
> Hit in production against a commercially deployed AC charger (vendor Iocharger, model IOCAW12-22TE-CRW4, firmware 25101501), OCPP 2.0.1, security profile 2. Iocharger is named by the Open Charge Alliance itself as an OCPP 2.0.1–certified charging-station vendor (Core + Advanced Security profiles — see the OCA announcement https://openchargealliance.org/974/ and the OCA "frontrunner companies" page). The point being: this is a certified-vendor charger behaving per the OCPP 2.0.1 spec — a Transaction.Begin placeholder while offline is legitimate — so this is a CSMS-side accounting bug, not a rogue-device firmware quirk to be worked around per vendor. Any conformant charger that starts a transaction offline can produce the same 0 placeholder.

There's a lot to discuss here. That the Iocharger is OCA certified, and therefore its behavior can be used as a reference for the spec, is an illogical leap. CitrineOS is OCA certified (Core + Advanced Security profiles) so why can't this work in reverse? We are certified, therefore the charger is in error and should fix its behavior.

I would definitely categorize this as a "rogue-device firmware quirk", though I would change my mind if I saw any evidence from the actual spec that would justify it. I did double check, and I don't see any such evidence (if I have missed something please do let me know). There are configurations for avoiding this however! _SampledDataTxStartedMeasurands_ allows you to change the meter values reported on start. If the charging station can't report the register accurately on start, it can be configured not to report it at all.

And the proposed fix is not benign. There are, in fact, charging stations that report meter values from 0. It is, for example, frequently done by simulators, and by some L2 home stations.

PR #1061, opened 2026-09-14
fix(ocpp): don't crash on RabbitMQ connection errors with no error listener

LoveKapibarasan:
## Problem

`RabbitMQConnectionManager` re-emits every amqplib connection `'error'` on itself:

```ts
this.connection.on('error', (err) => {
  this._logger.error('RabbitMQ connection error:', err);
  this.emit('error', err);
});
```

Nothing subscribes to `'error'` on the manager apart from the transient `once('error', reject)` inside a concurrent `connect()`. An `EventEmitter` that emits `'error'` with no listener throws, so the process exits:

```
RabbitMQConnectionManager    RabbitMQ connection error:
 Error  Unexpected close
node:events:485
      throw er; // Unhandled 'error' event
Emitted 'error' event on RabbitMQConnectionManager instance at:
    at ChannelModel.<anonymous> (.../rabbit-mq/ConnectionManager.js:39:22)
    at Connection.onSocketError (.../amqplib/lib/connection.js:306:12)
    at endReadableNT (node:internal/streams/readable:1701:12)
Node.js v24.4.1
```

amqplib always follows a connection `'error'` with `'close'`, and the `'close'` handler right below already schedules the reconnect. Because the throw happens first, that reconnect path never runs.

This is hit whenever the broker side drops an established socket without an AMQP `Connection.Close` — a TCP proxy or load balancer in front of the broker restarting, or a WAN blip. (The `succeed`/`onOpenOk` frames in the amqplib stack are misleading: amqplib creates that `Error` at handshake time and only emits it later, on socket `'end'`.) On one deployment this took the whole CSMS down 18 times in six weeks, each time until Docker restarted the container.

A broker-initiated graceful close (`rabbitmqctl close_all_connections`) does **not** reproduce it, since that path emits `'close'` without `'error'`.

## Change

Only forward `'error'` when a listener is registered. Consumers that do subscribe still receive it; otherwise the error is logged and the existing `'close'` → `handleReconnect()` path takes over.

## Tests

`packages/ocpp/test/transport/queue/rabbit-mq/connection-manager.test.ts` (new):

- a connection `'error'` followed by `'close'` with no error listener does not throw, and the manager reconnects — fails on `next` with `Error: Unexpected close`, passes with this change
- errors are still forwarded to registered listeners

Also checked end to end against a real RabbitMQ 3.13 broker behind a plain TCP proxy that sends FIN: the unpatched build exits with `Unhandled 'error' event`; the patched build logs the error, reconnects about a second later, and keeps running.


packages/ocpp/src/transport/queue/rabbit-mq/connection-manager.ts
```diff
@@ -54,7 +54,14 @@ export class RabbitMQConnectionManager extends AbstractConnectionManager<amqp.Ch
 
       this.connection.on('error', (err) => {
         this._logger.error('RabbitMQ connection error:', err);
-        this.emit('error', err);
+        // amqplib always follows a connection 'error' with 'close', and the
```
thanaParis, line 57: Remove this comment please.

thanaParis (changes requested):

PR #1064, opened 2026-09-15
build(docker): slim production images via pnpm deploy --prod

mrnonaki:
## Problem

`apps/{ocpp,ocpi}-server/deploy.Dockerfile` build the workspace, then `COPY` the **entire** built monorepo onto the full `node:24.16.0` image. That ships every package's `devDependencies` and the whole source tree — the ocpp-server image lands around **~1.7 GiB**.

## Change

Switch each app's final stage to a pruned production bundle with `pnpm deploy --legacy --prod`, on `node:24.16.0-slim`:

- **ocpp-server ~1.7 GiB → ~414 MB**, **ocpi-server → ~530 MB** (measured with `podman build`).
- **No pnpm/corepack at runtime.** The entrypoints call `./node_modules/.bin/sequelize-cli db:migrate` directly instead of `pnpm run db:migrate`, so the runtime image needs neither pnpm nor a corepack fetch at container start. ocpp-server's `DB_STRATEGY` branch and the `provision-partitions` step are preserved; ocpi-server gains an `entrypoint.sh` mirroring ocpp-server (migrate → start), replacing its pnpm-based `CMD`.
- **Runtime files outside `dist/`.** `pnpm deploy` honours `files: ["dist"]`, so `entrypoint.sh`, `.sequelizerc`, and the existing `hasura-metadata` / `rbac-rules.json` are `COPY`ed explicitly. Assets are emitted to `dist/assets` by the build's copy-assets step, so the slim image sets `CITRINEOS_FILEACCESS_LOCAL_DEFAULTFILEPATH` / `CITRINEOS_SWAGGER_LOGOPATH` to point there (the fat image defaulted to `src/assets`, which the pruned bundle no longer contains). Both remain overridable at runtime.
- **Prune-visible runtime deps.** `sequelize-cli` (invoked by the entrypoint), plus `sequelize` / `sequelize-typescript`, were only resolvable from the workspace root, so `--prod` dropped them from the per-app bundle — added to each app's `dependencies`. `ts-node/register` removed from both `sequelize-bridge.config.ts` (ts-node is a devDep, pruned by `--prod`).

Images deploy at the **same path** (`apps/<app>`) as before, so existing run commands / manifests are unchanged — a drop-in tag swap.

## Verification

Built both images with podman (node 24.16). The ocpp-server image boots to completion — loads `dist/assets/websocket-servers.json`, resolves all workspace deps transitively (including `awilix`/`tslog`, which are **not** added as direct deps), and reaches the RabbitMQ connection step — with **no missing-module or asset-path errors**. ocpi-server loads all modules cleanly (stops only at the normal `APP_NAME` runtime-env requirement). TypeScript build of both apps stays green.

## Notes

- `--legacy` is required because pnpm 10's `deploy` needs either injected workspace packages or that flag.
- No behavior change for the app itself — only image build/runtime packaging.


yuvenus:
A bunch of the actions are failing, please fix @mrnonaki, thanks!

mrnonaki:
Rebased onto current `next` and fixed the failing actions — all lanes are green now. The remaining `operator-ui-e2e-everest` failure (E2E-097b) also fails identically on unrelated PRs (e.g. runs 35308826731, 35300522884), so it looks unrelated to this change.

thanaParis (approved):

PR #1069, opened 2026-09-15
fix(ocpp): reject a RequestStartTransaction charging profile that sets transactionId

sjh37:
## Summary

F01.FR.11 (2.0.1 and 2.1): "The transactionId in the ChargingProfile SHALL NOT be set."
RequestStartTransactionEndpoint detects this but strips the field and logs a warning - and strips it on a copy, so
the profile is stored without the transactionId while the original `request` is still sent with it, which F01.FR.11
forbids.

This rejects the request instead of repairing it, the same way the `chargingProfilePurpose !== TxProfile` check
directly above does.

## Impact

`POST` RequestStartTransaction for a 2.0.1 or 2.1 station with a TxProfile whose `chargingProfile.transactionId` is
`"tx-abc"`:

- today the CSMS logs a warning, stores the profile without the transactionId, and sends
  RequestStartTransactionRequest still carrying `chargingProfile.transactionId: "tx-abc"`;
- with the change the caller gets `{ success: false, payload: 'The transactionId in the ChargingProfile SHALL NOT be
  set.' }`, nothing is stored, and no call is sent.

A request without a transactionId in the profile is unaffected.

This is a behaviour change for any caller that sends a transactionId and currently gets a success back. Per review
(#1069), refusing is the right answer: the request is malformed and silently repairing it hides the caller's bug.

## Test

`packages/ocpp/test/apis/ocpp/ev-driver-endpoints.test.ts` - a TxProfile with `transactionId: 'tx-abc'` is refused,
`sendCall` is not made and the profile is not stored.

## Notes

Ignore the branch name as it still says "strip"; the behaviour is now a rejection.


packages/ocpp/src/apis/ocpp/2/ev-driver/request-start-transaction-endpoint.ts
```diff
@@ -103,6 +104,7 @@ export class RequestStartTransactionEndpoint extends AbstractMessageEndpoint {
 
         if (chargingProfile.transactionId) {
           chargingProfile.transactionId = undefined;
```
thanaParis: removing it is questionable behavior... that this means the request is malformed--the upstream caller sent something they didn't mean to, and we're unlikely to be helping by 'fixing' it rather than throwing an error. you didn't put this check here, but if we're fixing something in this handler I think it should be to throw rather than fully remove the transactionId.

thanaParis (changes requested):

packages/ocpp/src/apis/ocpp/2/ev-driver/request-start-transaction-endpoint.ts
sjh37: Agreed. I'll drop the strip-and-warn and refuse the request the same way the purpose check just above does:

```ts
if (chargingProfile.transactionId) {
  results.push({
    success: false,
    payload: 'The transactionId in the ChargingProfile SHALL NOT be set.',
  });
  continue;
}
```

That removes the need for the new `payload` local entirely.
The request is either valid and sent unchanged, or rejected before `sendCall`.

btw - this is me, Simon, not Claude. Nice to meet you :-)

sjh37: PR updated.

thanaParis (approved):

PR #1071, opened 2026-09-15
fix(dal): keep one VariableMonitoring row per station monitor

sjh37:
## Summary

The CSMS stores variable monitors keyed on station, component and variable only, so a second monitor on the same variable overwrites the first. OCPP allows several: N04.FR.10 only makes a monitor a Duplicate when one "with the same type and severity" already exists on that component/variable. SetMonitoringDataType.id (2.0.1 Part 2 2.42, and the same text in 2.1) says "An id SHALL only be given to replace an existing monitor. The Charging Station handles the generation of id's for new monitors". So a SetVariableMonitoring entry without an id is always a new monitor. After the overwrite, the station's result for the first monitor cannot be matched, the repository throws, and the response handler abandons every later result, so the station-assigned ids (N04.FR.11) are never stored.

This change:

- The SetVariableMonitoring path matches an existing row only when the request gives an `id` (tenant, station, id), and otherwise inserts a new row.
- NotifyMonitoringReport matches on the reported `variableMonitoring.id`, which is required there, scoped to the tenant.
- When several rows share a type and severity, `updateResultByStationId` answers the pending one (id still null) first. That case now exists: an installed monitor plus a new id-less Set of the same type and severity, which the station answers Duplicate.
- `SetVariableMonitoringResponseOcpp2Handler` logs a result it cannot apply and carries on with the rest.

## Where it bites

- SetVariableMonitoring sends two entries on `EVSE/Power`: UpperThreshold 22000 and LowerThreshold 100, both severity 4.
  - Now: the CSMS keeps one row (`LowerThreshold, 100`). The station answers Accepted with ids 11 and 12. The result for id 11 finds no UpperThreshold row and throws, so id 12 is never applied either. The CSMS knows neither id, so later ClearVariableMonitoring results by id match nothing.
  - With the fix: two rows, holding ids 11 and 12.
- NotifyMonitoringReport lists monitors 11 and 12 on one variable that the CSMS already holds as two rows.
  - Now: both entries resolve to the same row, and the second update violates the `(ocppConnectionName, tenantId, id)` unique constraint.
  - With the fix: each entry updates its own row.
- An id-less SetVariableMonitoring repeats the type and severity of installed monitor 11, and the station answers Duplicate.
  - Now: the installed monitor's row takes the refused value.
  - With the fix: the refused request stays on its own pending row, and the Duplicate status is recorded there.

## Test

- `packages/dal/test/repositories/sequelize/variable-monitoring-integration.test.ts` (Postgres via testcontainers) covers:
  - two new monitors on one variable give two rows;
  - both Accepted results store their ids;
  - a Set carrying an id replaces that monitor;
  - a Duplicate answer lands on the new monitor, not the installed one;
  - a NotifyMonitoringReport of two known monitors updates each row.

  On next, four of these fail: `expected [ [ 'LowerThreshold', 100 ] ] to deeply equal [ [ 'UpperThreshold', 22000 ], …(1) ]`, `Unable to update set monitoring result: [object Object]`, `expected [ [ 11, 30000 ] ] to deeply equal [ [ 11, 22000 ], [ null, 30000 ] ]`, and `SequelizeUniqueConstraintError: Validation error`. The replace-by-id case already passes on next and guards the id-keyed path.
- `packages/ocpp/test/handlers/responses/2/set-variable-monitoring-response-ocpp-2-handler.test.ts`: a result the repository cannot match does not stop the next one being applied. On next the handler rejects with `Unable to update set monitoring result`.

Judgement calls:

- NotifyMonitoringReport entries are still written with `Promise.all`. Once they are keyed on the station's unique monitor id, no two entries in one report can resolve to the same row.
- A row left without an id (its Set result never arrived) is not merged into a later report; the report creates its own row for that id.
- The Drizzle `DrizzleVariableMonitoringRepository` is a stub without these methods, so there is no Drizzle twin to change.

Verified on `next` at `cd53765`: the new test fails before the fix as described above; with the fix, and with the four other OCPP 2.x fixes raised alongside it merged together, build clean, `pnpm test` green (345 files, 4538 passing / 8 skipped), `pnpm run lint` 0 errors.


thanaParis (approved):

thanaParis (changes requested):
Issues with tests.

thanaParis (approved):

PR #1073, opened 2026-09-15
fix(ocpp): accept QRCODE display messages in NotifyDisplayMessages

sjh37:
## Summary

OCPP 2.1 adds QRCODE to MessageFormatEnumType: "Message content is a text (usually a URL) that Charging Station will display as a QR code". The repo's own 2.1 NotifyDisplayMessagesRequest schema allows it. `validateMessageContent` switches over the version-neutral DTO enum, which has no QRCODE, so a QRCODE message reaches `default` and is reported as `Unknown message format: QRCODE`. The handler then answers the whole NotifyDisplayMessages with a PropertyConstraintViolation CALLERROR and stores none of the messages. O03.FR.02 has the station "send all configured DisplayMessages", so while a station holds one QRCODE message it cannot report any of them.

The validator now accepts QRCODE content that is not empty. Separately, `messageInfo` is optional in both the 2.0.1 and 2.1 schemas (only `requestId` is required), but the handler iterates it unguarded. A NotifyDisplayMessages without it throws `TypeError: messageInfoTypes is not iterable` and gets an InternalError. The loop now treats a missing `messageInfo` as empty.

## Where it bites

- The operator sends SetDisplayMessage `{format: QRCODE, content: "https://pay.example.com/..."}` to a 2.1 station. The endpoint checks only the language tag, and the station Accepts.
  - The SetDisplayMessage response handler marks every display message on that station inactive and sends GetDisplayMessages.
  - The station's NotifyDisplayMessages includes the QRCODE message. The CSMS answers with a CALLERROR and stores nothing, so every display message on that station now reads as inactive.
  - The same happens after every ClearDisplayMessage and GetDisplayMessages, for as long as the station holds the QRCODE message.
  - With the fix, the messages are stored and the station gets a CALLRESULT.
- A station answers GetDisplayMessages with a NotifyDisplayMessages that has no `messageInfo`.
  - Now: InternalError.
  - With the fix: an empty CALLRESULT.

## Test

- `packages/ocpp/test/handlers/requests/2/notify-display-messages-request-ocpp-2-handler.test.ts`:
  - a 2.1 request with one QRCODE message is stored and answered with a CALLRESULT. On next it fails with `expected "spy" to be called with arguments: [ 1, { id: 7, …(2) }, …(2) ]`, because the handler sent a CALLERROR instead;
  - a request without `messageInfo` is answered with a CALLRESULT. On next it fails with `TypeError: messageInfoTypes is not iterable`.
- `packages/ocpp/test/util/validator.test.ts`: QRCODE content is valid and empty QRCODE content is not. On next the valid case fails with `expected false to be true`.

Judgement calls:

- The HTML check, which refuses HTML content containing no tag, is left as it is.
- QRCODE is now in the version-neutral DTO `MessageFormatEnumSchema` in `packages/types` too (review suggestion), so the MessageInfo and tariff display-message DTOs can carry a 2.1 QRCODE message, and the validator matches it through the DTO enum like the other four formats.

Verified on `next` at `cd53765`: the new test fails before the fix as described above; with the fix, and with the four other OCPP 2.x fixes raised alongside it merged together, build clean, `pnpm test` green (345 files, 4538 passing / 8 skipped), `pnpm run lint` 0 errors.


packages/ocpp/src/util/validator.ts
```diff
@@ -703,6 +703,15 @@ export function validateMessageContent(
           'UTF8 format requires valid UTF-8 encoded content without unpaired surrogate characters',
       };
 
+    case OCPP2_1.MessageFormatEnumType.QRCODE:
```
yuvenus: Is there a reason not to add QRCODE to the `MessageFormatEnumSchema`? 🤔

sjh37: No good reason - done in 53ed27ba4. QRCODE is now in `MessageFormatEnumSchema`, so the MessageInfo and tariff display-message DTOs can carry it, and the validator matches it through `MessageFormatEnum.QRCODE` like the other four formats. Build, `pnpm test` and lint are clean on the branch.

yuvenus (approved):

PR #1076, opened 2026-09-15
perf(ocpp-server): index the device model and location lookups by station name

sjh37:
## Summary

The device model and location repositories read by `(tenantId, ocppConnectionName, ...)`, but every index on `VariableAttributes`, `Connectors`, `Evses` and `ChargingStationSequences` leads with `stationId`, which none of those queries filter on. Postgres 16 has no skip scan, so each lookup is a sequential scan of the whole estate's rows. `VariableStatuses` has only its primary key, yet it is joined on `variableAttributeId` by every device model read and never pruned. One migration adds five indexes, one per lookup, following #1000.

| Index | Query it serves |
|---|---|
| `VariableAttributes (tenantId, ocppConnectionName, componentId, variableId)` | `createOrUpdateDeviceModelByStationId` and the GetVariables path (`readOrCreateByQuery` on the full key, twice per NotifyReport attribute). Its `(tenantId, ocppConnectionName)` prefix serves `updateResultByStationId`, `readAllByQuerystring` (TransactionEvent, Authorize, StatusNotification, NotifyEvent), `readAllSetVariableByStationId` and the boot `manageSetVariables` update |
| `VariableStatuses (variableAttributeId, createdAt)` | the `VariableStatus` join in `readAllByQuerystring` and in the `reload` after each NotifyReport attribute / SetVariables result; the latest-Accepted lookup (`ORDER BY createdAt DESC LIMIT 1`); the `ON DELETE CASCADE` from `VariableAttributes` |
| `Connectors (tenantId, ocppConnectionName, connectorId)` | 1.6 connector upsert on every StatusNotification, `readConnectorByStationIdAndOcpp16ConnectorId`, `createTransactionByStartTransaction`; the prefix serves the 2.x `evseTypeConnectorId` lookups and GetTariffs |
| `Evses (tenantId, ocppConnectionName, evseTypeId)` | the `evse.readOrCreateByQuery` in `createOrUpdateTransactionByTransactionEventAndStationId`, `readEvseByStationIdAndOcpp201EvseId`, `createOrUpdateEvse` |
| `ChargingStationSequences (tenantId, ocppConnectionName, type)` | `getNextSequenceValue`: every 1.6 StartTransaction and every CSMS-initiated request id |

The Drizzle twins of the location repositories filter the same columns. The write cost is small. `VariableAttributes` rows are inserted only on an attribute's first report, and none of the indexed columns change afterwards, so the frequent `value`/`generatedAt` updates stay HOT. The same holds for connector status, EVSE and sequence `value` updates. `VariableStatuses` gains one btree entry per status row, which is the same rate as its heap inserts.

## Where it bites

On a migrated database with 500 stations (100,000 `VariableAttributes`, 300,000 `VariableStatuses`, 1,000 EVSEs, 2,000 connectors), `EXPLAIN ANALYZE` of the SQL Sequelize generates (captured from the repositories with logging on), before and after `sequelize-cli db:migrate`:

| Query | Before | After |
|---|---|---|
| NotifyReport readOrCreate on `VariableAttributes` | Seq Scan, 49,902 rows removed, 5.8 ms | Index Scan on the new index, 0.08 ms |
| `updateResultByStationId` / `readAllByQuerystring` shape (Component + Variable joins, `VariableStatuses` left join) | Parallel Seq Scan of both tables, 36.3 ms | Nested loop over the two new indexes, 0.20 ms |
| Latest Accepted `VariableStatus` | Parallel Seq Scan + Sort, 17.5 ms | Index Scan Backward, 0.03 ms |
| `reload({ include: [VariableStatus] })` | Parallel Seq Scan, 14.1 ms | Index Scan, 0.05 ms |
| 1.6 connector | Seq Scan, 999 rows removed | Index Scan |
| EVSE by `evseTypeId` | Seq Scan, 500 rows removed | Index Scan |
| `getNextSequenceValue` | Seq Scan, 748 rows removed | Index Scan |

The device model numbers grow with the whole estate, not with one station. `VariableStatuses` is never deleted from, so each NotifyReport costs more than the last one.

```
Limit (actual time=0.040..0.040 rows=1 loops=1)
  ->  Index Scan using variable_attributes_tenant_connection_component_variable on "VariableAttributes" "VariableAttribute"
        Index Cond: (("tenantId" = 1) AND (("ocppConnectionName")::text = 'S0250'::text) AND ("componentId" = 26) AND ("variableId" = 7))
        Filter: ((type)::text = 'Actual'::text)

Limit (actual time=0.014..0.014 rows=1 loops=1)
  ->  Index Scan Backward using variable_statuses_variable_attribute_id_created_at on "VariableStatuses" "VariableStatus"
        Index Cond: ("variableAttributeId" = 49903)
        Filter: (((status)::text = 'Accepted'::text) AND ("tenantId" = 1))
```

## Test

`apps/ocpp-server/test/migrations/device-model-and-location-indexes-integration.test.ts` applies every migration in `apps/ocpp-server/migrations` in order to a testcontainers Postgres, the same way `sequelize-cli` does. It then checks three things:

- each of the five indexes exists with exactly the declared columns in order;
- with `enable_seqscan = off`, the planner answers each repository query (the SQL above) with its index;
- `down` removes exactly those five, and `up` run twice leaves one of each.

7 tests, about 15 s. Like the existing `authorization-unique-constraint-integration.test.ts` beside it, the file is not picked up by the root `vitest.config.ts` (there is no `ocpp-server` project), so CI does not run it. I ran it with a vitest config rooted at `apps/ocpp-server`. `sequelize-cli db:migrate`, `db:migrate:undo` and `db:migrate` again against a scratch database: 5 indexes after `up`, 0 after `undo`, 5 after the second `up`.

Judgement calls:

- **No model or Drizzle schema change**, following #1000 and `20260205150000-add-latest-ocpp-message-timestamp`. `VariableAttribute` already declares `@Index` on `ocppConnectionName` (and the Drizzle schema mirrors it as `variable_attributes_ocpp_connection_name`), which no migration creates. A `sync()`-built database therefore has a single-column index that also serves these reads, and a migrated one now has the composite. The drift predates this change and is left alone.
- **Not added:** an index on `OCPPMessages.stationId` for the operator UI's message list, which would tax every message insert on the busiest table for a UI-only read, and an index on `VariableMonitoringStatuses.variableMonitoringId`, which serves one `reload` per SetVariableMonitoring result.
- **`ChargingStationSequences` stays non-unique.** A unique index would be equivalent to the existing `(stationId, type)` key, but it could fail to build on existing data.


yuvenus:
Please provide a script or reproduction of this issue; did you run a load test and notice these results or did you only analyze the database?

sjh37:
Hi @yuvenus 

Fair question, and the honest answer is that this came from reading query plans rather than from a load test - I traced the lookups the device model and location repositories issue on boot and on status notifications, saw the station-name predicates going to sequential scans, and added composite indexes to match. I didn't have a reproducible load harness to point at, which is what you're really asking for.

It's moot for this branch in any case: `ocppConnectionName` has since been dropped from `VariableAttributes`, `Connectors`, `Evses` and `ChargingStationSequences` in favour of the numeric `stationId`, so four of these five indexes are on a column that no longer exists. Closing was the right call.

If the underlying lookups are still worth indexing, I'll redo them against `stationId` and come back with EXPLAIN ANALYZE before and after on a seeded database, plus a script to reproduce it.

PR #1077, opened 2026-09-15
perf(ocpp-server): drop single-column indexes that duplicate a composite prefix

sjh37:
## Summary

Four indexes are strict prefixes of a composite index on the same table, and are maintained on every insert without serving any query the composite does not. Three of them are mine, from #1000. `20260818120000-partition-transactions` had already created the `(…, transactionCreatedAt)` composites under different names, so `CREATE INDEX IF NOT EXISTS` did not dedupe them. One migration drops all four. `DROP INDEX` on a partitioned parent drops each partition's child index with it.

| Dropped | Covered by |
|---|---|
| `meter_values_transaction_database_id (transactionDatabaseId)` | `meter_values_txn_ref (transactionDatabaseId, transactionCreatedAt)` |
| `meter_values_transaction_event_id (transactionEventId)` | `meter_values_event_ref (transactionEventId, transactionCreatedAt)` |
| `transaction_events_transaction_database_id (transactionDatabaseId)` | `transaction_events_txn_ref (transactionDatabaseId, transactionCreatedAt)` |
| `event_data_station_id (ocppConnectionName)` | `event_data_stationName_tenantId_eventId (ocppConnectionName, tenantId, eventId)` |

`EventData` also declared the single-column index on the model (`@Index` on `ocppConnectionName`), so `sync()` kept creating it. The decorator is replaced by the composite that `20260729000000-drop-eventdata-eventid-unique-constraint` creates, which leaves a `sync()`-built database with the same index as a migrated one.

## Where it bites

`MeterValues` is the highest-volume table after `OCPPMessages`: each meter value insert maintains six btrees, two of them redundant, in every monthly partition. `TransactionEvents` has one redundant btree per insert. `EventData` has one per NotifyEvent. On a database with every migration applied, `EXPLAIN (COSTS OFF)` with `enable_seqscan = off` shows the equality lookups the repositories run still use an index scan after the drop, now on the composite:

```
== MeterValues by transaction (readAllMeterValuesByTransactionDataBaseId, MeterValue include on Transaction reads)
before: Index Scan using "MeterValues_2026m09_transactionDatabaseId_idx" on "MeterValues_2026m09"   (one per partition)
after:  Index Scan using "MeterValues_2026m09_transactionDatabaseId_transactionCreate_idx" on "MeterValues_2026m09"
          Index Cond: ("transactionDatabaseId" = 1)
== MeterValues by transaction event
before: Index Scan using "MeterValues_2026m09_transactionEventId_idx"
after:  Index Scan using "MeterValues_2026m09_transactionEventId_transactionCreatedAt_idx"
== TransactionEvents by transaction
before: Index Scan using "TransactionEvents_2026m09_transactionDatabaseId_idx"
after:  Index Scan using "TransactionEvents_2026m09_transactionDatabaseId_transaction_idx"
== EventData by station (before and after)
        Index Scan using "event_data_stationName_tenantId_eventId" on "EventData"
          Index Cond: (("ocppConnectionName")::text = 'S0250'::text)
```

The planner already preferred the composite for the `EventData` lookup before the drop.

## Test

`apps/ocpp-server/test/migrations/drop-duplicate-prefix-indexes-integration.test.ts` applies every migration in `apps/ocpp-server/migrations` in order to a testcontainers Postgres, the same way `sequelize-cli` does, and checks three things:

- on each table, including every partition, no index remains that is only the leading column, and a composite leading with that column is still there;
- with `enable_seqscan = off`, each of the four lookups above still plans without a sequential scan;
- `down` recreates the four on the parents and `up` run twice removes them.

9 tests, about 11 s. Like the existing `authorization-unique-constraint-integration.test.ts`, the file is outside the root vitest projects, so CI does not run it. I ran it with a vitest config rooted at `apps/ocpp-server`. `sequelize-cli db:migrate`, `db:migrate:undo`, `db:migrate` on a scratch database: 16 matching parent and partition indexes before, then 3, 16 and 3. The 3 that stay are the `ChargingNeeds` partition copies described below.

Judgement calls:

- **Left alone:** `charging_needs_transaction_database_id`, `charging_profiles_transaction_database_id` (duplicates the partial `charging_profiles_txn_ref` for equality lookups), `message_infos_station_id` and `variable_monitorings_station_id`. They are also prefixes of a composite, but on low-write tables. `ocpp_messages_correlation_id` is left too: no current query filters `correlationId` alone, but that should be confirmed against `pg_stat_user_indexes` in a production database before dropping it.
- The Drizzle `event_data` schema's index declarations are not used to build any database (tests and `sync()` go through Sequelize) and already differ from the migrations, so they are unchanged.


yuvenus:
Closing since no reproducible scripts or load tests to see this performance dip, and this was already a problem introduced initially. Rather than fix it with more fire, there will be upcoming work to consolidate all indexes and neaten up the database once and for all.

PR #1078, opened 2026-09-15
fix(dal): send pending boot SetVariables on a Pending 2.x boot

sjh37:
## Summary

A boot config can carry `pendingBootSetVariableIds`: variables the CSMS should set with SetVariables while a 2.x station is Pending (B02). They are stored as `VariableAttributes.bootConfigId`, but nothing is ever sent, because three defects sit on the same path:

- `readByKey` and `updateByKey` return the boot without its `pendingBootSetVariables` association, in both the Sequelize and the Drizzle repository. `BootNotificationService.determineBootStatus` therefore never counts them, and the 2.x BootNotification handler never enters its SetVariables branch (`boot-notification-request-ocpp-2-handler.ts:227`), which is gated on that property.
- Past that gate, `readAllSetVariableByStationId` filters on `bootConfigSetId`. The column and model attribute are `bootConfigId`, so the query fails with `column VariableAttribute.bootConfigSetId does not exist`.
- Past that, `createSetVariableDataType` spreads Sequelize model instances into `component` and `variable`. The result is `_changed`, `_options`, `dataValues` and similar, not the `name` that `ComponentType` and `VariableType` require.

The fix covers all three:

- Both repositories load the association. Sequelize adds `include: [VariableAttribute]` to `readByKey` and to `_updateByKey`, whose `reload()` keeps the include. Drizzle loads the rows by `bootConfigId`.
- The filter uses `bootConfigId`.
- The component and variable are mapped field by field (`name`, `instance`, `evse.id`, `evse.connectorId`).

## Impact

An operator sets `pendingBootSetVariableIds` on a station's boot config through `boot-config-endpoints.ts`, with status Pending. The station sends a 2.x BootNotification.

- With the defaults (`autoAccept: true`) and `getBaseReportOnPending: false` on the boot config, `determineBootStatus` sees no pending variables and answers Accepted straight away. The variables are never sent.
- When the boot is answered Pending, `updateBootConfig` returns the result of `updateByKey`, whose `pendingBootSetVariables` is `undefined`. The handler skips SetVariables and later auto-accepts the station, still without the configuration it was meant to receive.
- If the gate were passed, the handler would throw at `readAllSetVariableByStationId`. Past that, the SetVariablesRequest it built would have no component or variable `name`.

After the fix, the Pending boot sends one SetVariablesRequest per `ItemsPerMessageSetVariables` batch, with the stored values, as the handler already intends.

## Test

- `packages/dal/test/repositories/sequelize/device-model-integration.test.ts`: `readAllSetVariableByStationId` returns only the attributes assigned to the station's boot, as `SetVariableData` with the component and variable names. On next: `column VariableAttribute.bootConfigSetId does not exist`. With only the column fixed, it fails on the spread instances.
- `packages/dal/test/repositories/sequelize/boot-integration.test.ts`: `readByKey` and `updateByKey` return the assigned attributes in `pendingBootSetVariables`. On next: `expected undefined to deeply equal [ 1 ]`.
- `packages/dal/test/repositories/drizzle/drizzle-boot-integration.test.ts` (new): the same two cases for `DrizzleBootRepository`. On next: `expected undefined to deeply equal [ 1 ]`.
- `packages/ocpp/test/handlers/requests/boot-notification-pending-set-variables.test.ts` (new): with a Pending boot that has pending variables, the handler sends SetVariables with the repository's data; with none, it does not read or send them. This pins the handler branch and passes on next; the defects are below it.

All pass with the fix. The suites in `packages/dal/test/repositories`, `packages/ocpp/test/handlers/requests` and `packages/ocpp/test/modules/configuration` (70 files, 438 tests) are green.

Judgement calls:

- `GET` boot config goes through `readByKey`, so its response now includes `pendingBootSetVariables`. `BootDto` already declares the field, and the `POST` response already returned it.
- `DrizzleBootRepository`'s constructor dropped the `drizzleInstance` it was given, unlike the other Drizzle repositories, so it always used the shared default pool. It now passes it on; the new Drizzle test relies on that to use, and close, its own pool.
- The Drizzle repository reads the attributes with its own query rather than through `DrizzleVariableAttributeRepository`, matching how the rest of the file queries `bootTable` directly.
- Passing `pendingBootSetVariableIds` makes Sequelize warn `Unknown attributes (pendingBootSetVariableIds) passed to defaults option of findOrCreate`, because `createOrUpdateByKey` spreads its input into `defaults`. That is existing behaviour and is left alone.


thanaParis (approved):

thanaParis (changes requested):
broken test.

thanaParis (approved):

PR #1083, opened 2026-09-15
fix(ocpi): fetch the partner's endpoints again on PUT /credentials

sjh37:
## Summary

`CredentialsService.putCredentials` stores the partner's new versions URL and token and issues a new server token, but
never calls the partner's versions endpoint. The endpoint list stored at registration stays in place, so every call
CitrineOS makes to the partner afterwards (Sessions, CDRs, command results, real-time authorisation) still goes to the
old URLs. The change fetches the versions list and version details from the URL in the PUT body, with the token in the
PUT body, before the profile is written, the same way `postCredentials` and `registerCredentialsTokenA` already do.

OCPI 2.2.1, Credentials module, "Changing endpoints for the current version": a PUT to the credentials endpoint is
how a party tells the other to fetch and store its set of endpoints again. I am citing this from memory - there is
no copy of the OCPI spec in the repo - so please check the wording against your copy.

## Impact

- An eMSP moves its OCPI API to a new host and calls `PUT /ocpi/2.2.1/credentials` with the new `url`. CitrineOS
  answers 1000 with a fresh token, but keeps sending session PUT/PATCH and CDR POST to the old host. Those calls fail
  until someone deletes and re-registers the partner.
- An eMSP starts offering a module (say `cdrs` RECEIVER) and announces it with a PUT. `LIST_TENANT_PARTNERS_BY_CPO`
  filters on the stored endpoints, so CitrineOS never uses the new module.
- The new versions URL does not offer 2.2.1. Now the PUT fails with the same `NotFoundError` a POST gets, and
  nothing is stored. Before the change, CitrineOS accepted credentials it could not use.

Judgement call: like `putCredentials` before it, the change does not compare the path version with the registered
version. POST does, but a version upgrade through PUT is a separate question, and I have left it out.

## Test

`packages/ocpi-base/test/services/credentials-service-put-credentials.test.ts` stubs the versions client to publish a
new endpoint list. It pins two things: the stored profile carries the new endpoints and version details URL, and
nothing is stored when the new versions URL does not offer the registered version. On `next`, both fail:
`expected [ { …(2) } ] to deeply equal [ { …(2) }, { …(2) } ]` (the registration-time list is stored) and
`promise resolved "{ …(3) }" instead of rejecting`. Both pass with the fix.

Verified on `next` at `a682ad0`: the new test fails before the fix as described above; with the fix, and with the seven other OCPI fixes raised alongside it merged together, build clean, `pnpm test` green with no unhandled errors (349 files, 4541 passing / 8 skipped), `pnpm run lint` 0 errors.


packages/ocpi/src/services/credentials-service.ts
```diff
@@ -147,6 +147,7 @@ export class CredentialsService {
       versionsUrl: credentials.url,
       token: credentials.token,
     };
+    tenantPartner = await this.getVersionDetails(tenantPartner, credentials.url);
```
yuvenus, line 151: Is there a particular reason this assignment happens after the lines above, where other tenantPartner information is assigned?

thanaParis (changes requested):
broken tests.

packages/ocpi/src/services/credentials-service.ts
sjh37: Hi @yuvenus. 
Yes, `getVersionDetails` authenticates the outbound versions call with `tenantPartner.partnerProfileOCPI`, which carries the token. On a `PUT`, the partner hands us a replacement token, so we have to write the new credentials before the fetch; otherwise, we'd re-read their endpoints using the token they've just retired. Roles are assigned afterwards so they land on the object that gets persisted. `postCredentials` can call it first because there's no token being replaced there.

sjh37:
Hi @thanaParis 

The `live` failures are fixed - the cause was in the mock, not in the OCPI change.

`rotateCredentials` set `reg.tokenWeAccept` only *after* the credentials PUT returned. Once this PR makes the CSMS re-read the partner's versions during the PUT, the mock rejects that call with 401, `getVersionDetails` throws, the PUT fails, and `/reregister` surfaces it as a 502 - which is what the two failing tests were seeing.

The register path already depends on the opposite ordering, and says so at `apps/mock-msp/src/core/client.ts:344-346`: *"Citrine fetches our version list/details with ourToken while it is still handling the POST, so it has to be accepted before the call goes out."*

The rotate path simply never needed it before, because nothing re-read on PUT. So the fix mirrors register: accept the new token before sending the PUT, and restore the previous one if the PUT is refused.

I've verified by running the live stack locally against this branch's own images - `reregister rotates our credentials and the old token dies` and `register.ts drives the same handshake` both pass, and mock-msp's 585 unit tests are unaffected.

thanaParis (approved):

PR #1085, opened 2026-09-15
fix(ocpi): apply the configured page limits to paginated GETs

sjh37:
## Summary

`defaultPageLimit` and `maxPageLimit` sit in the OCPI config (`packages/ocpi-base/src/config/ocpi-types.ts`). They are
set in `apps/ocpi-server/src/config/envs/local.ts` and in `docker.ts` (`DEFAULT_PAGE_LIMIT`, `MAX_PAGE_LIMIT`), but
nothing reads them. `PaginatedParams.limit` is validated with only `@Min(1)`, so a partner can ask for any page size.
X-Limit echoes the request back, and an omitted limit falls back to the hard-coded `DEFAULT_LIMIT` of 10.
`PaginatedMiddleware` is already attached to every paginated GET (Sessions, CDRs, Locations, Tariffs) through
`@Paginated()`, and it already owns X-Limit. It now takes `config` from the container and sets the request's `limit`
before the endpoint runs: `defaultPageLimit` when the request has none, `maxPageLimit` when it asks for more. The
services echo the limit they are handed, so X-Limit reports the cap.

OCPI 2.2.1 (transport and format, pagination) lets the server return fewer objects than the client asks for, and
defines X-Limit as "the maximum number of objects that the server can return". I am citing that from memory - there
is no copy in the repo.

## Where it bites

A registered eMSP calls `GET /ocpi/2.2.1/cdrs?limit=100000`. CdrsService reads up to 100,000 transactions with their
meter values in one GraphQL query. `BaseTransactionMapper` then makes three sequential GraphQL round trips per
transaction. The request runs for minutes and the server's memory climbs. With the change, the query asks for 1000 (the
shipped `maxPageLimit`), X-Limit says 1000, and the Link header points at `offset=1000`.

Judgement calls:

- Clamping happens in the middleware, not in each of the four services. It is one change that covers every
  paginated endpoint, it needs no new dependencies on the services, and the middleware is the class that already
  handles paging. It works by rewriting `ctx.request.query`: routing-controllers resolves `@QueryParams()` from
  `ctx.query` when the route handler runs, which is after `UseBefore` middleware.
- An omitted `limit` now gives `defaultPageLimit` (50 as shipped) instead of 10, because that is what the config
  says. Anyone who relied on 10 can set `DEFAULT_PAGE_LIMIT=10`.
- A limit that is not a number is left alone, so `PaginatedParams` validation still rejects it as before.

## Test

`packages/ocpi-base/test/util/middleware/paginated-middleware.test.ts` now builds the middleware with a config and
adds three cases, each with a stand-in endpoint that echoes the limit it receives. A limit of 100000 reaches the
endpoint as 1000, with X-Limit 1000. No limit reaches it as 50. A limit of 20 is left untouched, along with the rest of
the query. On `next`, the first two fail with `expected 100000 to be 1000` and `expected NaN to be 50`, and the
typecheck reports `Expected 0 arguments, but got 1`. All pass with the fix.

Outside the suite, I ran the built `@Paginated()` decorator and middleware through a real Koa and routing-controllers
server. The controller received `limit` 1000, 50 and 20 for `?limit=100000`, no query, and `?limit=20&offset=40`.

Verified on `next` at `a682ad0`: the new test fails before the fix as described above; with the fix, and with the seven other OCPI fixes raised alongside it merged together, build clean, `pnpm test` green with no unhandled errors (349 files, 4541 passing / 8 skipped), `pnpm run lint` 0 errors.


yuvenus:
Closing as this solution is too narrow; as the code comment says, it's specifically trying to read the limit and offset from the PaginatedResponse (NOT the query), and you can see this behavior in where the PaginatedResponse is constructed.

PR #1094, opened 2026-09-15
feat(ocpp): negotiate permessage-deflate on the websocket servers

sjh37:
## Summary

OCPP-J 3.4 (2.1 and 2.0.1): "The CSMS (and Local Controller) SHALL support RFC 7692, WebSocket compression", and Table
2 lists the CSMS as "Required". `ws` leaves `perMessageDeflate` off on the server by default (`false` in
`WebSocketServer`), and CitrineOS never turned it on, so a station offering `permessage-deflate` always ran
uncompressed. Each websocket server now enables it, behind a new `perMessageDeflate` flag in the websocket server
config that defaults to `true`.

## Impact

A 2.0.1 or 2.1 station on mobile data opens its connection with `Sec-WebSocket-Extensions: permessage-deflate`. The
101 response carries no extension, so every NotifyReport, GetBaseReport result and SetVariables batch goes over the air
uncompressed. That is the mobile-data cost the requirement exists to cut. Nothing fails outright, which is why it has
gone unnoticed. With the change the extension is negotiated. A station that does not offer it, including a typical
1.6 station, is unaffected, because RFC 7692 is negotiated per connection.

Judgement calls:

- **Config flag.** `perMessageDeflate: z.boolean().default(true).optional()` sits next to the other websocket server
  options and follows the `ignoreAuthenticationHeaders` pattern, so existing `WebsocketServerConfig` literals still
  type-check. The server reads it as `wsConfig.perMessageDeflate ?? true`, so a config that did not pass through the
  schema also gets the spec's default.
- **Memory caveat.** The `ws` README warns that permessage-deflate "adds a significant overhead in terms of performance
  and memory consumption", and that Node's zlib can suffer "catastrophic memory fragmentation" under high concurrency,
  especially on Linux. Each connection that exchanges a compressed message holds its own zlib context, so on an
  instance with many thousands of stations the resident memory goes up. Operators who hit that can set
  `"perMessageDeflate": false` on that websocket server. That is a deliberate deviation from the SHALL, but it is now
  their choice rather than the default.
- **Threshold.** The change keeps the `ws` defaults: a 1024-byte threshold (smaller messages go uncompressed), a zlib
  concurrency limit of 10, and context takeover. Heartbeats, StatusNotifications and most responses stay under 1 KB
  and cost no zlib work. The large reports and batches, where compression pays, are compressed. Tuning the zlib
  window or memory level would need measurements against a real fleet first, so this change does not attempt it.

## Test

- `packages/ocpp/test/transport/network-connection/websocket-network-connection.test.ts` is new. It starts a
  `WebsocketNetworkConnection` server on an ephemeral port with a stub authenticator, and connects a real `ws` client
  that offers permessage-deflate. It checks that the extension is negotiated by default and not when the server sets
  `perMessageDeflate: false`.
- `packages/base/test/config/config-loader.test.ts` adds a test that the loaded websocket servers config defaults
  `perMessageDeflate` to `true` and keeps an explicit `false`.

On `next` they fail with `expected '' to contain 'permessage-deflate'` and
`expected [ undefined, undefined ] to deeply equal [ true, false ]`. With the change they pass.


packages/types/src/config/types.ts
```diff
@@ -49,6 +49,7 @@ export const websocketServerSchema = z
     // strict per-server tenant.
     dynamicTenantResolution: z.boolean().optional().default(false),
     forceProtocol: z.enum(OCPP_VERSION_LIST).optional(),
+    perMessageDeflate: z.boolean().default(true).optional(),
```
yuvenus, line 52: Why is this default true? People should be able to opt-in to turning this on if the current behavior is that it stays off.

thanaParis: this makes some sense actually. in the cases where the client has it disabled (most cases) this changes nothing. in the cases where the client has it enabled, this would presumably deliver the expected behavior--it would be fixing a bug where the client (the charging station) had promised compressed messaging but hadn't been able to deliver due to the server rejecting the option.

thanaParis (approved):

PR #1095, opened 2026-09-15
fix(ocpp): trim VIES company fields to the AddressType limits

sjh37:
## Summary

C18.FR.09 has the CSMS respond "with VatNumberValidationResponse with status = Accepted if validated", carrying the
registered company as an `AddressType`. The 2.1 schema limits `name` to 50 characters and `address1`, `address2` and
`city` to 100. `ViesVatProvider._toAddressType` copied the VIES name and address lines through untrimmed, so a valid
VAT number for a company with a long registered name or address produced a response that failed outbound validation.
The station then got a CALLERROR `FormatViolation` instead of Accepted. Each of those four fields is now cut to its
schema maximum.

## Where it bites

A 2.1 station sends VatNumberValidationRequest for a valid German VAT number whose VIES record has a name such as
"... Gesellschaft mit beschränkter Haftung & Co. Kommanditgesellschaft", which is well over 50 characters. VIES answers
`valid: true`, `ViesVatProvider` returns the name as it is, `sendCallResultWithMessage` rejects the response at
`/company/name maxLength`, and the station receives a CALLERROR. The same happens when one address line, or the joined
middle lines that become `address2`, exceed 100 characters. The driver cannot get an invoice to the company, although
the VAT number is valid.

Judgement calls:

- The fix keeps the start of each field. That is the part a person reads, and truncation is the only option that
  keeps Accepted. The alternative of answering Rejected would tell the station that a valid VAT number is invalid.
- The trim is in the provider, where registry data becomes an `AddressType`, rather than in the handler.
- `postalCode` is not trimmed, because the parser's pattern caps it at 10 characters. `country` is not trimmed either,
  because it is the two-letter VIES country code.

## Test

`packages/ocpp/test/services/vat-provider/vies-vat-provider.test.ts` is new. It stubs `fetch` with a VIES response
and checks three things:

- A 60-character name and address lines over 100 characters give a company that the real `OCPPValidator` accepts in
  a VatNumberValidationResponse.
- Each long field keeps its first 50 or 100 characters.
- A company within the limits comes back unchanged.

On `next` the first two fail, with schema errors at `/company/name`, `/company/address1`, `/company/address2` and
`/company/city`, and with the untrimmed values. With the fix they pass.


yuvenus:
In general, CitrineOS tends towards **not being prescriptive** in regards to external data breaking spec limits. In this particular example, we can't control the data that comes from the API reached for the VAT information, so if the company names are over 50 characters, we'd rather forward it along than force a truncation (and also, though unlikely, there could be a company that wouldn't enjoy their company name getting truncated without their consent 😅).

Therefore, in this sort of scenario, it'd make more sense to make your own bespoke map of potentially long company names and the desired truncation for them, rather than push a solution to the open source. Out of curiosity, and without providing sensitive information, if you have a real-life use case of you hitting this limit it'd be cool to know it for future reference, but not as a reason to implement a fix.

Hope that makes sense! I will close this PR. If you have any other PRs open that make similar calls, I'd ask that you close them since that goes against the aforementioned philosophy on external data breaking the protocol.

sjh37:
Understood, and thanks for spelling out the principle. No real-life case: it came from checking the VIES response mapping against the 2.1 AddressType schema limits, not from a station in the field. On the same principle I am closing #960, which truncates charging-rate decimals to the protocol precision on the way out. #1092 I have left open because the value it shortens is the receiptId the CSMS generates itself, not external data; if you see that the same way, say so and I will close it too.

PR #1096, opened 2026-09-15
fix(ocpp): claim the outstanding CALL slot atomically in sendCall

sjh37:
## Summary

OCPP-J 4.1.1: "A Charging Station or CSMS SHALL NOT send a CALL message to the other party unless the CALL message it
sent before has been responded to or has timed out" (SHOULD NOT in 1.6). `MessageRouterImpl.sendCall` enforced this
with `existsAnyInNamespace` followed by `set`, two separate awaited steps. When several sends to one station run
concurrently, all of them can pass the check before any of them has written its entry, and they all go on the wire
together. `sendCall` now also claims a per-station slot with `setIfNotExist`, which is atomic in both `MemoryCache`
and `RedisCache`, so exactly one CALL wins and the others get `RetryMessageError` as before.

## Impact

SetVariables with more items than `ItemsPerMessageSetVariables` is split by `sendInBatches`, which publishes the
batches back to back. The same goes for GetVariables, SetVariableMonitoring and ClearVariableMonitoring. The RabbitMQ
receiver runs `_onMessage` for each delivery without a prefetch limit, so the router starts several `sendCall`s for
the same station at once. With the memory cache the identical await chains interleave in lock-step. With Redis the
window is two round trips. Either way several SetVariables CALLs are sent while none has been answered. A station that
enforces 4.1.1 drops them, answers with a CALLERROR or disconnects, and the batch results come back incomplete.

How the claim works:

- The slot is a cache entry under the key `outstanding-csms-call` in the station's existing transactions namespace.
  Its value is the correlationId, and its TTL is `maxCallLengthSeconds`, the same as the CALL's own entry, so a
  timeout frees it exactly as before.
- The existing `existsAnyInNamespace` check stays in front of the claim, so today's behaviour is unchanged. That
  includes not sending while an incoming station CALL is still being handled, because those entries share the
  namespace.
- The slot is released when the send fails, and when a CallResult or CallError answers a pending CALL. A late answer
  to a CALL that has already timed out finds no pending entry and does not release the slot of the CALL that followed
  it.

Judgement call: a separate slot key is the smallest atomic claim `ICache` supports. `ICache` has no "set if the
namespace is empty", and serialising `sendCall` in-process would not hold across instances sharing Redis.

## Test

`packages/ocpp/test/modules/ocpp-router/router-outstanding-call.test.ts` is new. It runs `MessageRouterImpl` on a real
`MemoryCache` with a stub network hook. The test that fails today starts two concurrent `sendCall`s to one station
and expects the network hook to be called once and the other call to reject with `RetryMessageError`. Three
regression cases check that the next CALL still goes out after the outstanding one is answered with a CallResult or a
CallError, or after its send fails.

On `next` the concurrency test fails with `expected "spy" to be called 1 times, but got 2 times`. With the fix it
passes, and the existing `message-router-impl.test.ts` suite still passes.

The hunks in `router.ts` are in `sendCall`, `_onCallResult`, `_onCallError` and a new private helper. They stay clear
of the unknown-message-type branch that #968 edits and the `_onCall` validation throw that #1050 changed.


packages/ocpp/src/modules/ocpp-router/router.ts
```diff
@@ -418,7 +420,15 @@ export class MessageRouterImpl extends AbstractMessageRouter implements IMessage
...
+        !(await this._cache.existsAnyInNamespace(transactionNamespace)) &&
+        (await this._cache.setIfNotExist(
+          OUTSTANDING_CALL_CACHE_KEY,
+          correlationId,
+          transactionNamespace,
+          this._config.timeouts.maxCallLengthSeconds,
+        ))
+      ) {
         const cacheTimestamp = new Date();
         await this._cache.set(
```
thanaParis, line 433: why are you using a new cache namespace when you could just move this line into the if clause and change it to a 'setIfNotExist' call?

thanaParis (changes requested):

packages/ocpp/src/modules/ocpp-router/router.ts
sjh37: It is the same namespace, just a second key inside it - `outstanding-csms-call` under `Transactions<identifier>`.

`setIfNotExist` on the correlation id cannot serialise two senders: `correlationId` defaults to a fresh `uuidv4()` per call, so two concurrent `sendCall`s for the same station write two different keys and both succeed. The claim has to be on a key that is identical for every sender, which is what the fixed key gives.

Your point about the extra work stands though - the namespace-wide `existsAnyInNamespace` check is redundant once the claim guards the slot, so I have dropped it in 8319914. `sendCall` now takes a single atomic decision, and the existing "call already in progress" test drives `setIfNotExist` instead.

If you want only one key in the namespace, the alternative is to store the outstanding call under the fixed key as `<correlationId>@<action>@<timestamp>` and have the two response paths read that key and check the id matches. Happy to do that instead, it is just a wider diff.

thanaParis (approved):

PR #1097, opened 2026-09-15
Use ocpp connection name for URLs

devonNewsomDevelop:
Changes URLs to use charging station's ocppConnectionName while forming the underlying infratstructure to allow this while maintaining pkid usage


apps/operator-ui/src/lib/client/pages/charging-stations/detail/charging-station-detail.tsx
```diff
@@ -32,6 +35,14 @@ export const ChargingStationDetail: React.FC<ChargingStationDetailProps> = ({ pa
     }
   }, [id]);
 
+  if (isNotFound) {
+    return (
+      <div className={`${pageMargin} ${pageFlex}`}>
+        <NoDataFoundCard />
```
yuvenus: Minor, can you also include the message prop for specificity? From the pasted code below you'll need to import `translate`:

```
<NoDataFoundCard message={translate('ChargingStations.noDataFound', { id: ocppConnectionName })} />;
``` 

The translate message takes an "id" argument to template in, but it probably makes the most sense to display ocppConnectionName, hence the slightly odd latter argument. 😆

yuvenus (approved):
Just a tiny comment, otherwise looks good, works like a charm locally! 🙏

apps/operator-ui/src/lib/client/pages/charging-stations/detail/charging-station-detail.tsx
devonNewsomDevelop: Good call, I did not know this was a thing as I havent worked with the actual components of the front-end too much

PR #1098, opened 2026-09-15
fix(ocpp): bind the Security Profile 3 client certificate to the station identity

sjh37:
## Summary

On a Security Profile 3 websocket server the TLS options only set `requestCert` and `rejectUnauthorized`, so the handshake checks that the client certificate chains to the configured CA and nothing more. The `Authenticator` then takes the station identity from the URL path, and `BasicAuthenticationFilter` skips profile 3, so the path is the only identity check: any station holding a certificate issued by the CSO CA can connect as any other identity. Nothing in the repository reads the peer certificate (`getPeerCertificate` and `TLSSocket` do not appear anywhere).

OCPP 2.0.1 and 2.1 Part 2, Security Profile 3 table, A00.FR.405: "The CSMS SHALL verify that the certificate belongs to this Charging Station by checking that the CN (commonName) RDN in the subject field of the certificate contains the unique serial number of the Charging Station." B01.FR.12: when the serial number "does NOT equal the Serial Number in the Certificate Common Name - The CSMS SHALL close WebSocket connection."

The change adds `ClientCertificateFilter`, which runs after the existing four filters and only when the server's security profile is 3. It reads the peer certificate from the upgrade request's TLS socket and refuses the upgrade (401, like the basic-auth filter) unless the certificate is present and verified and its subject CN equals the station's stored `chargePointSerialNumber`, or the URL identity while no serial number has been stored for the station.

## Impact

Station A holds a valid certificate from the CSO CA with `CN=A`. It opens `wss://csms/ocpp/B` on the profile 3 server.

- Now: the TLS handshake passes, `UnknownStationFilter` finds B, `ConnectedStationFilter` passes unless B is online, and the connection is registered as B. A's BootNotification, Authorize and TransactionEvent messages are stored against B, and every command the operator sends to B reaches A.
- With the change: the upgrade is refused with `Certificate CN A does not match B (expected B)` and the socket is closed.

The same path serves OCPP 1.6 stations on a profile 3 server, so they get the same check.

## Test

- `packages/ocpp/test/transport/network-connection/authenticator/authenticator.test.ts` (new): builds the `Authenticator` with the four existing filters mocked and a request whose TLS socket presents `CN=CS-A` for URL identity `CS-B`. On next it fails with `promise resolved "{ identifier: 'CS-B' }" instead of rejecting`; with the change it rejects with `UpgradeAuthenticationError`. The matching-CN case still resolves.
- `packages/ocpp/test/transport/network-connection/authenticator/client-certificate-filter.test.ts` (new): the filter is skipped on profiles 0, 1 and 2; rejects when the socket has no certificate, when the certificate was not verified against the CA, or when it is empty; rejects a CN that is neither the identity nor the stored serial number; accepts a CN equal to the identity when no serial number is stored (with or without a station record); accepts a CN equal to the stored serial number when the URL identity differs.

Judgement calls:

- What the CN is compared against. A00.FR.405 says the serial number, but CitrineOS keys everything on the URL identity and has no serial number until the first BootNotification stores one. The rule here is: the stored `chargePointSerialNumber` when the station record has one, otherwise the URL identity. A station whose URL identity is not its serial number and that has never booted will be refused on profile 3 until the operator sets `chargePointSerialNumber` on the station record, or until it has booted once on a lower profile.
- Profile 3 only. Profiles 1 and 2 do not request a client certificate, so there is nothing to compare.
- Only the upgrade is checked. B01.FR.11 (compare `BootNotificationRequest.serialNumber` with the CN at boot time and close on mismatch) needs the CN carried from the upgrade into the boot handler and is left for a follow-up.
- The subject O (A00.FR.404, CSO name) is not checked. `SignCertificateRequestOcpp2Handler` already checks O on the CSR against `SecurityCtrlr.OrganizationName`, and adding a device-model read to the upgrade path is a separate decision.

Verified on `next` at `a682ad0`: the new tests fail before the fix as described above; with the fix, and with the two other security fixes raised alongside it merged together, build clean, `pnpm test` green (343 files, 4547 passing / 8 skipped, no unhandled errors), `pnpm run lint` 0 errors.


packages/ocpp/src/transport/network-connection/authenticator/client-certificate-filter.ts
```diff
@@ -0,0 +1,69 @@
...
+      );
+    }
+
+    const commonName = certificate.subject?.CN;
+    const chargingStation =
+      await this._chargingStationRepository.readChargingStationByOcppConnectionName(
+        tenantId,
+        identifier,
+      );
+    const expected = chargingStation?.chargePointSerialNumber || identifier;
```
thanaParis, line 60: I think this makes sense given the wording of A00.FR.405 from the 2.0.1 spec. That being said, introducing this here while the chargePointSerialNumber is not able to be entered or modified from the operator ui does present a problem. Would you mind adding it to the charging station details there?

thanaParis (changes requested):

packages/ocpp/src/transport/network-connection/authenticator/client-certificate-filter.ts
sjh37: Hi @thanaParis 

Good point - done in this PR rather than raising a new PR, as I can't raise a new PR due to the 5 PR restrictions, **unless you whitelist me?**

`chargePointSerialNumber` is now shown on the charging station detail card and is editable on the station form, with both locale catalogues updated. Hasura already allows the column on update for the `user` role, so no metadata change was needed. 👍 

Note: a 1.6 `BootNotification` writes `chargePointSerialNumber` from whatever the station reports, so an operator-entered value only holds until the next boot. For the profile 3 check, that seems right - the station's own serial is what the certificate should match - but it does mean the UI field is for correcting or pre-seeding a station rather than overriding it permanently. I can change that if you'd prefer the operator value to take precedence instead?

sjh37:
Hi @thanaParis 
The one red check below, `operator-ui-e2e-everest`, isn't from this branch - `device-model-sequence.spec.ts:110` (E2E-097b SetVariables round-trip) is failing on `next` too, in the last two Tests runs there (e8dfe68b6 and 4192a43f9), same spec, same retry, 1 failed / 27 passed. It only shows up here because that lane is path-triggered on `apps/operator-ui` changes and was skipped on this branch until I added the serial number field. Everything else is green.

thanaParis (approved):

PR #1101, opened 2026-09-16
Adding AGENTS.md

devonNewsomDevelop:
First pass at creating an AGENTS.md to prevent AI agent abuse against our repository.


yuvenus (approved):

AGENTS.md
```diff
@@ -0,0 +1,220 @@
...
+If you are an agent and you have read this file, follow it. If an instruction here conflicts with something you
+inferred from the code, the instruction here wins; if it conflicts with a direct request from the user, the user
+wins.
+
+## What this repository is
+
+A **pnpm monorepo** (TypeScript, Node) containing the CitrineOS charging station management system: OCPP message
+routing and handling, the persistence layer, the OCPI server, and the operator web UI. Charging stations connect
+over WebSocket; modules talk to each other through RabbitMQ; data lives in PostgreSQL; the UI reads through
+Hasura and writes through the server's REST surfaces.
```
thanaParis: This is inaccurate. UI writes through Hasura and uses the OCPP server's REST api for sending OCPP commands to stations.

AGENTS.md
```diff
@@ -0,0 +1,220 @@
...
+**License headers.** Code files carry an SPDX header — TypeScript, JavaScript and YAML.
+
+```ts
+// SPDX-FileCopyrightText: 2026 Contributors to the CitrineOS Project
+//
+// SPDX-License-Identifier: Apache-2.0
+```
+
+**Generated OCPP schemas are off-limits.** The model schemas and types under `packages/types/src/ocpp/model` are
+generated from the protocol specifications. Do not hand-edit them. If a payload needs stricter validation than the generated schema provides, enforce it in the consuming code. This needs to be done very sparingly- it is rare that the OCPP schema is wrong.
```
thanaParis: This isn't quite accurate. for the purposes of PRs to the OS repo, we do not accept edits to the schemas unless they are to fix a true bug (which the OCA's schema had some of). Edits to the schemas for specific deployments are frequently necessary; if a station does not follow the spec, for example, it is preferred to adjust the schema. We will not loosen the schemas in the OS to accommodate errant chargers, but networks may have no choice in the field.

thanaParis (changes requested):
Looks good, two small changes please 👍

AGENTS.md
devonNewsomDevelop: But a station in this case wouldn't be opening a PR, right? They are doing this type of editing on site. This is for agents to follow, not for production use cases. What is the purpose of changing this?

AGENTS.md
devonNewsomDevelop: What's inaccurate about this? All of our queries are read through Hasura, and the PUTs and POSTs such as admin commands are done through endpoints, for example. I must be missing something.

devonNewsomDevelop: Maybe it should be expanded to include that the messages from the front-end are primarily sent through the message handlers? I figured that was adequately covered by "modules talk to each other through RabbitMQ" but perhaps the issue is that the current verbiage reads like we primarily use MVC endpoints?

AGENTS.md
devonNewsomDevelop: Perhaps you are concerned that a station using AI on-site will be hindered by this rule. We could make a footnote stating that the AI can make schema edits in on-site applications and not for changes to the upstream repository?

AGENTS.md
yuvenus: It'd probably just be simpler to say that "the UI reads and writes through Hasura" (with no mention of REST surfaces), since that's what it essentially does. Obviously it technically uses POSTs/PUTs to power the GraphQL stuff (bc you can hit the GraphQL POST endpoint and pass a GraphQL query) but technically it's through "Hasura", if that makes sense. 🤔

devonNewsomDevelop: Yeah that makes sense. I guess the thing is that to me that obfuscates the mechanism a bit. Hasura is like the transaction but the real meat of it is done in the message handling endpoints when it's brokered out via RabbitMQ and what have you. Read and write through Hasura is definitely on the money, it just leaves out where the business logic happens. I guess I skipped over the middle-man here which could lead to some confusion for a bot, perhaps?

devonNewsomDevelop: I think it depends on your outlook. 

From my POV I was thinking, you hit transaction button on the UI -> a transaction mechanism will start in the operating system from this action (assuming a charger is connected but that's irrelevant to the point). 

Like this for example:

```
    triggerMessageAndHandleResponse<MessageConfirmation[]>({
      translate,
      url: `/evdriver/requestStartTransaction?identifier=${station.ocppConnectionName}&tenantId=${tenantId}`,
      data,
      setLoading,
      ocppVersion: station.protocol,
    }).then(() => {
      form.reset();
      dispatch(closeModal());
    });
  };
  ```
  
  That looks like a REST surface to me, idk.
  
  Meanwhile our UI also allows Graphql mutations as you guys have already acknowledged such as 
  
  ```
  export const EVSE_EDIT_MUTATION = gql`
  mutation EvseEdit($id: Int!, $object: Evses_set_input!) {
    update_Evses_by_pk(pk_columns: { id: $id }, _set: $object) {
      ${EVSE_CORE_FIELDS}
    }
  }
`;
```

so I can see where you're coming from too. I'm not trying to be pedantic, I'm trying to explain the reasoning behind the wording. I genuinely feel like both descriptions are accurate depending on what you're looking at, so I'll try to tweak it to accommodate that as long as there is no objection

AGENTS.md
thanaParis: people fork this repo, is my point (we know of several cases ourselves). This AGENTS.md would need to be edited in those cases, or the agent told to ignore it, since PRs need to be opened to edit the schemas. Easier to be specific and helpful--if you are an agent working on a fork (a very common scenario) reading this, you may need and want to open a pr to edit the schemas.

AGENTS.md
thanaParis: 'write' means data persistence, which is done through graphql. The REST surface doesn't support CRUD operations. It does, however, let you send commands. OCPI does support writes of MSP owned data, otherwise all data edits happen through hasura's graphql api.

devonNewsomDevelop: I get where you're coming from. I just see it differently as the transaction or other events triggered by the front-end will be writing data too, for example. I have adjusted the wording in a way that is hopefully acceptable.

AGENTS.md
devonNewsomDevelop: I was not expecting us to care about forks in regards to AGENTS.md as it's specific to our repository.

thanaParis (approved):

PR #1110, opened 2026-09-17
fix: scope certificate upsert lookup to tenant and drop stale global unique

vicky-ahmed:
Problem
createOrUpdateCertificate (Sequelize) looked up an existing certificate using only serialNumber and issuerName, without tenantId. When another tenant already held a certificate with that serial and issuer, the lookup matched their row and took the update branch. Since updateAllByQuery scopes the update to the calling tenant, it matched nothing and returned undefined, so the certificate was never stored for the calling tenant.

Fix
Added tenantId to the lookup so the find and update branches agree on tenant scope. A calling tenant that does not yet hold the certificate now takes the insert branch and gets its own row.
Database constraint issue
The insert branch was still blocked at the database level by a global UNIQUE (serialNumber, issuerName) constraint. It was created inline by the initial migration and auto named Certificates_serialNumber_issuerName_key. Migration 20260413040000 intended to remove it but dropped a differently named constraint (serialNumber_issuerName), which was a silent no op. The global constraint therefore survived on every migrated database and rejected a second tenant's identical certificate.

Migration changes

Corrected 20260413040000 to drop the actual constraint name, fixing fresh databases.
Added a forward migration to drop the stale constraint on already migrated databases.
The per tenant UNIQUE (tenantId, serialNumber, issuerName) constraint remains, so within tenant uniqueness is preserved.

Bug is resolved and tested e2e against live DB.


apps/ocpp-server/db/migrations/20260413040000-fix-certificate-unique-indexes.ts
```diff
@@ -10,6 +10,9 @@ export default {
   up: async (queryInterface: QueryInterface) => {
     console.log('Updating Certificates unique constraints to include tenantId...');
 
+    await queryInterface.sequelize.query(
```
thanaParis: don't alter an old migration like this, it doesn't work--it won't refire.

thanaParis (changes requested):

apps/ocpp-server/db/migrations/20260413040000-fix-certificate-unique-indexes.ts
vicky-ahmed: Ahh sorry, my mistake, you are right. I'd tested on a fresh DB and added that drop to the old migration too, but since it's already landed it won't refire, and the new migration I added handles it for all DBs anyway. I'll revert the change to the old migration. Thanks for catching this!

thanaParis (approved):

PR #1111, opened 2026-09-17
fix(dal): scope SendLocalList auth entries to the request tenant

vicky-ahmed:


packages/dal/src/repositories/sequelize/local-auth-list.ts
```diff
@@ -100,6 +100,7 @@ export class SequelizeLocalAuthListRepository
       const localListAuthorization = await this.localListAuthorization.create(
         tenantId,
```
yuvenus, line 101: I always thought this tenantId argument was supposed to solve this, but I guess it doesn't huh... 😓

yuvenus (approved):

packages/dal/src/repositories/sequelize/local-auth-list.ts
vicky-ahmed: Yeah, there are so many tenant related issues almost everywhere.

PR #1112, opened 2026-09-17
Fix: CDR Cost Not Based On Charging Time

lydiazcheng:
## Context
`timeSpentCharging` contains the total time that energy flowed from EVSE to EV during the transaction (in seconds) based on the OCPP protocol.
This PR is to make sure this field is set by charger or calculated by cirine. And the cdr total time cost is billed by the charging time rather than wall-clock duration, i.e., end time - start time.
Related Issue: https://github.com/citrineos/citrineos/issues/225
## Changes
1. [OCPP] set `timeSpentCharging` if charger provide it. If undefined, fallback to derives charging seconds from the chargingState, `Charging`. If also no any states, fallback to derives from rises in the cumulative energy register.
2. [OCPP] add `transactionInfoForRow` to avoid override `timeSpentCharging` to null
3. [OCPP] Fixed `readAllByStationIdAndTransactionId's` include from `Evse` to `EvseType` due to error: `SequelizeEagerLoadingError Evse is not associated to TransactionEvent!`
4. [OCPI] implement `calculateTotalParkingTimeHours`
5. [OCPI] `PricedSession` and `cdr-mapper` take addtional `timeSpentCharging` so that OCPI aligns with OCPP
6. [OCPI] Update charging period mapping with `TIME`/`PARKING_TIME` dimension.
7. [UI] Add `Time Spent Charging` in transaction details page and fix few other small issues. 
## Tests
1. Start charging using everest and the `timeSpentCharging` is 0. transaction_start_charging_time_0
2. Pause and resume the ev charging, the `timeSpentCharging` doesn't change. transaction_pause_charging_time_33 transaction_resume_charging_time_33
3. End charging and the `timeSpentCharging` is updated with correct value, which matches `endDateTime - startDateTime - parkingTime`. transaction_end_charging_time_40
4. Pull the CDR and check expected values
   - Screenshot 2026-09-18 at 12 15 16 PM
   - `TIME` and `PARKING_TIME` cdr dimension have attached to the charging period. And the volume is correct. 
   - The total parking time matches the pause and resume duration: 2026-09-18 09:32:40 - 2026-09-18 09:26:40 is around 0.1 hour
   - The total time cost is calculated correctly: 
      - Tariff 1 — USD, pricePerMin 0.05
      - Billable minutes = total_time − total_parking_time = 46.21 − 6.01 = 40.20 min
      - total_time_cost = 40.20 min × $0.05/min = $2.01


apps/operator-ui/src/lib/client/pages/transactions/detail/transaction-detail-card.tsx
```diff
@@ -123,15 +141,19 @@ export const TransactionDetailCard = ({ transaction }: TransactionDetailCardProp
...
+                  to={`/${MenuSection.LOCATIONS}/${transaction.locationId}`}
+                  className={clickableLinkStyle}
+                  title={String(transaction.locationId)}
+                >
+                  {transaction.location?.name ?? String(transaction.locationId)}
+                </Link>
+              ) : (
+                <span>{NOT_APPLICABLE}</span>
+              )
+            }
```
yuvenus, line 156: HEHE this was bothering me too, thanks for fixing it! 😍

packages/ocpi/src/mappers/cdr-mapper.ts
```diff
@@ -188,10 +200,6 @@ export class CdrMapper extends BaseTransactionMapper {
   // (OCPP 2.1's reservationTime/reservationFixed fields are untyped `any`
   // placeholders and aren't populated), so these stay unsupported rather
   // than guessing at a calculation with no backing data.
```
yuvenus: Minor, is this comment still relevant here or can it be removed? 🤔

yuvenus (approved):
Looking good locally so far 👀🙏

packages/ocpi/src/mappers/cdr-mapper.ts
lydiazcheng: Thanks for the review, Venus! I updated the comment. We still need to implement both parking and reservation cost methods in the future.

PR #1125, opened 2026-09-18
Drizzle Schema Drift Checker

yuvenus:
# Why?

With the introduction of the Sequelize schema drift checker (https://github.com/citrineos/citrineos-core/pull/930), it stands to reason that the Drizzle schema drift checker will follow soon after. So here it is!

The point of the drift checker is the same -- things declared in code don't always match what's in the database schema. Down the line, we may consider a different mechanism for dealing with database migrations once Drizzle is fully implemented (and possibly make the schema drift checker less useful than it currently is), but for now this is how we will attempt to verify that the shapes of the schemas in the code and database match.

# What Changed?

In general, the following changes needed to occur to eliminate all errors and most of the warnings:

1. `drizzle-schema-validator.ts` was added to actually verify the Drizzle schema
2. `citrineos-server` was updated to trigger the Drizzle schema validator if `CITRINEOS_USE_DRIZZLE` is true
3. Any Drizzle schemas that were caught in the drift checker were resolved. Luckily, it was mostly the declaration of the nullability of columns and inaccurate index names, so all schemas touched were updated accordingly (notNull()s added where needed, index names updated to reflect what's in the database).
4. Any Sequelize schemas that had warnings were resolved (EXCEPT for the ones regarding installed certificates, because those seem to have a more complicated fix). This was previously left out to focus on the initial phase of schema drift checking, but since it's out in the wild now, it made sense to fix them along with the Drizzle schemas.
5. A new migration was added `reconcile-drizzle-erroring-schema-differences` to do another round of reconciling the database schemas to the code that were causing errors on the Drizzle-side (and somewhat on the Sequelize-side also) -- most importantly, the changes were to ensure NOT-NULL-able columns are _actually_ NOT NULL in the database. 
6. Shared methods between the Sequelize and Drizzle drift checkers were pulled into their own file `schema-validation`.

# What counts as drift?

Recapping the same logic from the Sequelize drift check:

1. If an entire table is missing
2. If a column is missing
3. If the column type is different
4. If the nullability of a column is different
5. If the "width" of the column is different (i.e. for varchars)

There are a few allowances though for some of these differences, depending on how severe they are:

1. If an entire table is missing...
  * No allowances, it should fail the diff check ❌ 
2. If a column is missing...
  * If the column is in the database, but not the model, that's a warning ⚠️ 
  * If the column is in the model, but not the database, that's a failure ❌ 
3. If the column type is different...
  * No allowances, it should fail the diff check ❌ 
4. If the nullability of a column is different...
  * If the column is null in the database, but not the model, that's a failure ❌  (bc this would cause a type error)
  * If the column is null in the model, but not the database, that's a warning ⚠️ 
5. If the "width" of the column is different...
  * If the column is wider in the database than the model, that's a warning ⚠️ 
  * If the column is wider in the model than the database, that's a failure ❌  (since inserts from the code could fail if you exceed the width)

## What were the actual drifts?
[drizzle-drift-checker-logs.txt](https://github.com/user-attachments/files/32392528/drizzle-drift-checker-logs.txt)

Attached is the text dump of the logs at the time prior to the schema being fixed.

## What's left out?

For now, there are NO checks for the following:

1. Foreign keys - a separate PR will be open to ensure the appropriate foreign key references in the Drizzle schemas.
2. Indexes - aside from checking the names, the actual contents of the index aren't checked.


packages/dal/src/repositories/drizzle/base.ts
```diff
@@ -54,12 +59,20 @@ export interface DrizzleRepositoryDependencies {
...
 
-export abstract class DrizzleRepository<TTable extends CitrineTable, TDto> extends EventEmitter {
+/**
+ * Join tables — `ChargingStationNetworkProfiles`, `ComponentVariables`,
+ * `LocalListVersionAuthorizations`, `SendLocalListAuthorizations` — have a
+ * tenantId but no `id` column, so they extend this directly and reach their rows
+ * by natural key in their own domain methods. Everything else extends
+ * {@link DrizzleRepository}, which adds the id-keyed methods on top.
+ */
+export abstract class DrizzleTenantScopedRepository<
```
yuvenus, line 69: The addition of DrizzleTenantScopedRepository is the main change in this file -- some tables don't actually have an id column because they're join tables, but they certainly still have tenantIds. 😳 So this separation allows flexibility for repository classes for join tables, in case we need to perform operations against them.

The rest of the changes in the file are just reordering the methods to group them differently now that this separation exists.

